Issue #369, opened 2013-11-27
Assign values to variables in SLIM Query tables like in Decision tables


raboof:
In decision tables you can assign the values of output cells to variables by putting '$variablename=' in the cell.

It would be nice if this were possible for query tables, too.

Issue #91, opened 2012-05-14
Ability to use symbol in a Query table


EricD2:
Currently, the documentation states that symbols are only supported in "a Decision Table, Script Table, or any other kind of table that can call functions", sadly this doesn't seem to include Query table. 

I want to have something like this

<pre>
|Query: list foo by name|
|name |property |id     |
|foo1 |prop1    |$foo1= |
|foo2 |prop2    |$foo2= |
|foo3 |prop3    |$foo3= |

|Query: list bar with foo|
|name |idFoo |property52 |
|bar1 |$foo1 |1          |
|bar2 |$foo1 |2          |
|bar1 |$foo2 |4          |
</pre>

amolenaar:
This is kind-of resolved I'd guess? (if not, please reopen)

DmitryKonev:
It is partly resolved. #525 provides ability to assign symbol values in Query tables. But using symbols for asserts is still unavailable.

    |Query: list bar with foo|
    |name |idFoo |property52 |
    |bar1 |$foo1 |1          |

In this case $foo1 symbol is interpreted as a string.

PR #525, opened 2014-09-19
Slim Query/Table Table: manage variables assignment in VariableStore

antoine-aumjaud:
Variable set ($X=) was not possible on QueryTable(#91 and #369) and TableTable.

I have found a solution for TableTable (pull request #469) but it was not applicable on QueryTable.

This pull request propose a solution to manage both tables:
I have added a variables map in TestResult returned by expectation.
And do a new instruction 'assign' on SlimClient which set variables in the VariableStore.

This pull request requires a protocol modification, other implementations need to be updated (http://www.fitnesse.org/PlugIns#slim).
If the modification is not done in plugins, variables in theses tables can't be used.

### DIFF
--- a/test/fitnesse/testsystems/slim/tables/TableTableTest.java
+++ b/test/fitnesse/testsystems/slim/tables/TableTableTest.java
@@ -305,7 +305,7 @@ public class TableTableTest {
   public void tableWithSetSymbols() throws Exception {
     makeTableTableAndBuildInstructions(tableTableHeader + "|$A=|$B=|$C=|$D=|$E=|$F=|\n");
     Map<String, Object> pseudoResults = SlimCommandRunningClient.resultToMap(
-asList(
+            asList(
                     asList("tableTable_id_0", "OK"),
                     asList("tableTable_id_1", asList(
                         asList("pass:1", "ignore:2", "fail:3", "sole:l", "no change", "pass")


jplambert:
Hello everybody,

I am concerned about using symbols in Slim TableTable. Reading at your comments and the code, I understand the following. Can you confirm?
- The Slim protocol supports symbols or "run-time variables"
- The symbols are handled by the Slim server, not by the FitNesse runner
- This is implemented
- But the symbols are not available to TableTable fixtures because such fixtures delegate all the work to custom code, but the current interface (POJO with a doTable(...) method and constructor arguments matching the Wiki) do not give access to the symbols store

As a proof of concept, I hacked FitNesse so that when we write "$V=", it puts the result provided by the TableTable custom fixture into the corresponding symbol, and that works. However we see that this solution is not optimal:
- As stated in this thread or another one, that would not work from one row to another in the same table
- There is a conflict between what returns the TableTable custom fixture, which is meant to be formatted for FitNesse exposure (HTML, comparison between expected vs. actual) and what we want to store in the symbol (the raw value with no formatting at all)

... so this is not a valid solution.

This makes me thinking that the only way to would work is to expose the symbols store to the TableTable custom fixture. Which means changing the contract on the underlying POJO. Am I right? What do you think?

Thanks in advance,

Jean-Pierre Lambert

antoine-aumjaud:
Hi Jean-Pierre,

I validate your analysis.
In fact, it is possible to set a variable in variable store by implementing StatementExecutorConsumer interface.

> This makes me thinking that the only way to would work is to expose the symbols store to the TableTable custom fixture. Which means changing the contract on the underlying POJO. Am I right? 

I'm agree with you, I think that InstructionExecutor interface (injected when StatementExecutorConsumer is implemented by a fixture) should give access to all variables and paths defined in SlimExecutionContext.

Regards,
Antoine

jplambert:
Thanks Antoine,

Does that mean that when TableTable instantiates the underlying POJO, it checks whether it is an implementor of StatementExecutorConsumer to provide some additional data? Thus meaning that my concern (accessing the symbol store in a TableTable) already has a solution.

Or is it just a suggestion on how we could move forward?

antoine-aumjaud:
The additional object injected in the underlying POJO by TableTable (in fact by all Tables) gives access to the variable store: you can store a variable BUT can't read from it...
What I suggest is to add behavior in the injected object (InstructionExecutor) to permit to read a variable from the store.

One question : why do you need to access to a variable created in the same TableTable; in the doTable method you create the response, so you can access to all values, can't you ?

jplambert:
Actually, I do not need to access to variables ('symbols' in FitNesse-speak) created in the same TableTable. My problem is more about sharing symbols between tables. That implies to be able to replace symbols by their values, in other words to have access to the symbols table.

So far I have been able to get something working by hacking FitNesse so that TableTable replaces the symbols by their value, so that the underlying POJO gets the replaced text. Symbol assignment being already kind of supported in TableTable, putting the result provided by the POJO into the symbols table.

However I still have an issue with this approach, it is that the POJO may want to return values formatted for pretty-print by FitNesse, while what I want to assign is the raw value.

So, I am just talking about the symbols created and used in the same table for the sake of completeness about the subject. That would be an additional reason to "fix it".

To be honest, I don't see how the current solution would work completely. One can think of the following way to do:
- TableTable substitues the symbols he knows, "so that the POJO do not need to read the store"
- The POJO assigns the symbols as needed, and keeps track of these assignments
- The POJO then substitutes the symbols he has assigned, "thus still not needing to read the store"
Ignoring the fact that there would be two different components in charge of symbols substitution, it would not work in some cases:

| Table: Table 1 |
| $V= | assigns 'foo' to V |

| Table! Table 2 |
| $V | TableTable will substitute by 'foo' |
| $V= | assigns 'bar' to V |
| $V | should be 'bar' but has been substituted as 'foo' by TableTable |

So in short, I agree with you that we need to provide a way for the POJO to substitue variables by himself, with the nice side-effect that it would not need to maintain any local symbols sets.

I still need to try what you are saying about StatementExecutorConsumer! I've not seen anything about that in the code :-)

jplambert:
Hi, I have found in FitNesse code the hook on StatementExecutorConsumer, and tested it quickly.

So it looks like the only missing part is adding some 'get' method in InstructionExecutor so that TableTable fixtures can substitute symbols.

Is there any reason not to make this change? Is there any impact outside of the FitNesse code-base?

Thanks in advance,

JP

antoine-aumjaud:
not to my knowledge

jplambert:
Thanks! How should we move forward? Make some "Pull Request" with the proposed change? (I am new to GitHub)

antoine-aumjaud:
yes you can do a pull request

Issue #558, opened 2014-11-22
Enhancement:  Color Test Table Columns & Rows


mwarhaftig:
Our in-house test cases tend to be 300+ line ActionFixtures.  To make tests more readable we use custom JavaScript and CSS to add color-coded row background colors to the html (colored based on the action name):
![image](https://cloud.githubusercontent.com/assets/1711265/5151996/d896e2b0-71b4-11e4-9872-e70ddb21bac4.png)

The color coding greatly improves test readability but the in-house implementation is inflexible as it assumes all test tables have a vertical test flow.  For a better (and community shared) solution, at the fitnesse.wikitext.parser.Table.toTarget() level I believe it is possible to fairly accurately decide whether a test table has vertical or horizontal test flow and then add column or row coloring (and/or hooks for coloring via CSS).  For example here is a mockup of how some basic tables would look:
![image](https://cloud.githubusercontent.com/assets/1711265/5151999/e4704554-71b4-11e4-94cc-31201862f066.png)
![image](https://cloud.githubusercontent.com/assets/1711265/5152002/e66a42b0-71b4-11e4-9869-8c3e9985072d.png)

Please let me know if I can add table coloring logic to the codebase.  Suggestions or concerns are appreciated too.

amolenaar:
PR has been merged

PR #569, opened 2014-12-15
Issue #558 - Added table coloring based on Slim & Fit Fixture class.

mwarhaftig:
Wiki table coloring scheme rules based on Slim class type & Fit Fixture classes.

### DIFF
--- a/src/fitnesse/testrunner/WikiTestPage.java
+++ b/src/fitnesse/testrunner/WikiTestPage.java
@@ -49,8 +49,7 @@ public class WikiTestPage implements TestPage {
     String content = getDecoratedContent();
     ParsingPage parsingPage = new ParsingPage(new WikiSourcePage(sourcePage), variableSource);
     Symbol syntaxTree = Parser.make(parsingPage, content).parse();
-    return new HtmlTranslator(parsingPage.getPage(), parsingPage, 
-            sourcePage.getVariable(WikiPageIdentity.TEST_SYSTEM)).translateTree(syntaxTree);
+    return new HtmlTranslator(parsingPage.getPage(), parsingPage).translateTree(syntaxTree);
   }
 
   @Override
--- a/src/fitnesse/wikitext/parser/ColoredSlimTable.java
+++ b/src/fitnesse/wikitext/parser/ColoredSlimTable.java
@@ -126,11 +126,11 @@ public class ColoredSlimTable extends SymbolTypeDecorator{
     }
 
     public SymbolType isApplicable(Translator translator){
-        String testSystem = null;
+        Maybe<String> testSystem = Maybe.noString;
         if(translator instanceof HtmlTranslator){
-            testSystem = ((HtmlTranslator)translator).getTestSystem();
+            testSystem = ((HtmlTranslator) translator).getParsingPage().findVariable("TEST_SYSTEM");
         }
-        if(testSystem == null || !testSystem.equals("slim")){
+        if(testSystem.isNothing() || !testSystem.getValue().equals("slim")){
             return baseSymbolType;
         }
         return this;
--- a/src/fitnesse/wikitext/parser/HtmlTranslator.java
+++ b/src/fitnesse/wikitext/parser/HtmlTranslator.java
@@ -3,7 +3,6 @@ package fitnesse.wikitext.parser;
 public class HtmlTranslator extends Translator {
 
     private ParsingPage parsingPage;
-    private String testSystem;
 
     @Override
     protected Translation getTranslation(SymbolType symbolType) {
@@ -16,15 +15,8 @@ public class HtmlTranslator extends Translator {
 
     public ParsingPage getParsingPage() { return parsingPage; }
 
-    public String getTestSystem() { return testSystem; }
-
     public HtmlTranslator(SourcePage currentPage, ParsingPage parsingPage) {
-        this(currentPage, parsingPage, "fit");
-    }
-
-    public HtmlTranslator(SourcePage currentPage, ParsingPage parsingPage, String testSystem) {
         super(currentPage);
         this.parsingPage = parsingPage;
-        this.testSystem = testSystem;
     }
 }


amolenaar:
I think we need something like test system specific TestPage's. Each page being able to process test pages in a way appropriate for the tests system. Only I'm not excited by the idea that we introduce another factory class (maybe TestSystem should be extended to wrap it's test pages?).

I'm wondering: how can this work if I want to write a Cucumber or JBehave test system?

mwarhaftig:
Hi Arjan, I had some questions and confusion about integrating coloring with the test system.
* What would be gained by wrapping TestSystem around TestPage? - Don't TestSystems only get instantiated when a test is actually run while on the other hand TestPage is called by several responders?
* Would you be ok with using the WikiPage's TEST_SYSTEM variable to find the TestSystem for coloring?

If you could provide any more details on your original suggestion it would be appreciated.

Thanks.

amolenaar:
Hi Matt, sorry for the late response.

> What would be gained by wrapping TestSystem around TestPage? - Don't TestSystems only get instantiated when a test is actually run while on the other hand TestPage is called by several responders?

TestPage is also the interface through which pages are fed to the test system. TestSystems get instantiated only when a test is run. That's true. TestPages on the other hand are also created for normal wiki page rendering. They make sure SetUp and TearDown pages are added to pages with the Test flag enabled. Typically a SlimTestPage should include scenario libraries where FitTestPage shouldn't. The colouring you added is applicable for SlimTestPage's, not FitTestPage or CucumberTestPage (if we end up executing Cucumber tests from FitNesse, which would be a nice feature).

There needs to be a thing (factory?) that creates the appropriate type of TestPage for a given WikiPage. Should it be a new Factory class or can it be added to TestSystemFactory? Maybe TestSystem should take care of it (I think not, since it introduces a dependency from fitnesse.testsystems to fitnesse.wiki).

It's not clear to me what would be a nice solution.

> Would you be ok with using the WikiPage's TEST_SYSTEM variable to find the TestSystem for coloring?

I think that would be the variable of choice.

mwarhaftig:
Thanks for the detailed response Arjan.  I see your point about not coloring Scenario tables when in Fit test system and have done some research on your factory suggestion and other approaches.

WikiTestPage.getHtml() is the point that we parse the page content and translate it via HtmlTranslator (sub of Translator) to get the html.  When a table is part of the page's content then Translator.translateTree() will eventually get to a call of Table.toTarget() to generate the table's html.

With a TestPage factory and multiple TestPages (based on the test system) the same call to Table.toTarget() is still going to happen for each TestPage and the result in TestPage is still going to be the full page's html.  What I would propose instead of a factory is that we make add a TEST_SYSTEM property to HtmlTranslator and use that property in Table.toTarget() to decide what tables to color and what test system specific coloring to add.

What do you think about this idea?  If there is still some discussion about how to implement can we setup a quick chat (gchat, etc.) as I want to better understand your suggestions and ideas.

jediwhale:
FYI, fitSharp already does coloring on its result HTML, so there should 
be a way to turn off your coloring for users who don't need it.

On 2015-01-23 05:34, Matt Warhaftig wrote:
>
> Thanks for the detailed response Arjan. I see your point about not 
> coloring Scenario tables when in Fit test system and have done some 
> research on your factory suggestion and other approaches.
>
> WikiTestPage.getHtml() is the point that we parse the page content and 
> translate it via HtmlTranslator (sub of Translator) to get the html. 
> When a table is part of the page's content then 
> Translator.translateTree() will eventually get to a call of 
> Table.toTarget() to generate the table's html.
>
> With a TestPage factory and multiple TestPages (based on the test 
> system) the same call to Table.toTarget() is still going to happen for 
> each TestPage and the result in TestPage is still going to be the full 
> page's html. What I would propose instead of a factory is that we make 
> add a TEST_SYSTEM property to HtmlTranslator and use that property in 
> Table.toTarget() to decide what tables to color and what test system 
> specific coloring to add.
>
> What do you think about this idea? If there is still some discussion 
> about how to implement can we setup a quick chat (gchat, etc.) as I 
> want to better understand your suggestions and ideas.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/unclebob/fitnesse/pull/569#issuecomment-71187397>.
>

-- 
Cheers,
Mike Stockdale

/fit/Sharp <http://fitsharp.github.com>
Syterra Software Inc. <http://www.syterra.com>

amolenaar:
>What I would propose instead of a factory is that we add a TEST_SYSTEM property to HtmlTranslator and use that property in Table.toTarget() to decide what tables to color and what test system specific coloring to add.

I think you're pretty much spot on :). But instead of letting Table.toTarget() handle it, we might be able to pull it of with a decorator kind-of object instead(?), or override the toTarget() implementation with a custom one. I mean: for now the wiki text is used for Fit and Slim tables (we probably only want to add styling for Slim tables for a start).

Thinking aloud:
Say we add an interface with 2 methods: isApplicable() and toTarget() (identical to HtmlTranslator). We can attach instances of such objects to symbols. Those instances take precedence over the default translator. For the Table symbol type isApplicable() can then check if the TEST_SYSTEM == "slim". If so, the corresponding toTarget() method is called.

mwarhaftig:
Sounds good, the pull request has been updated to use a SymbolTypeDecorator and a Slim Table object having its own (colored) toTarget() method.  Non-Slim tables continue to use the original, unchanged Table.toTarget().

amolenaar:
Nice! I sent you a small PR (https://github.com/mwarhaftig/fitnesse/pull/1), that removes the test system getter from HtmlTranslator again. I think that's nicer :).

amolenaar:
Consider it done :)

mwarhaftig:
Thanks for giving it the final touch.  I am really happy this in the code base.

fhoeben:
@mwarhaftig I'm looking at FitNesse's codebase and I came across the colouring code as quite a performance bottleneck when starting the tests. 

Do you still remember some of your reasoning behind this colouring logic and can you maybe help me understand why it works the way it does?

To start of with: the code has been refactored a bit and it now resides in [fitnesse.wikitext.parser.decorator.SlimTableDefaultColoring](https://github.com/unclebob/fitnesse/blob/master/src/fitnesse/wikitext/parser/decorator/SlimTableDefaultColoring.java), but the question I have is on original intend which is still unchanged.

Most of the code is fine, but there is some reflection usage (`Class.forName()`) that is done often, and which is quite slow. (in code you added a comment `Unmarked decision tables aren't found by getTableType().  Color table if first row is valid class.` just before doing that)
I've taken the liberty of adding a small cache to prevent doing the class lookup for the same string over and over, but I can't quite understand why the class lookup is done in the first place. The test you wrote in the originally does not cover that part of the code as far as I understand. So I'm hoping you can clarify.

As I understand it you are trying to determine by the content's of a table's first row whether it is a fixture's class name by doing a lookup. 
That works for  FitNesse's own fixtures which are on the wiki's classpath. But in general (Slim) fixture classes don't need to be on the wiki's classpath, they need to be on the classpath of the test system when running the tests. This classpath can be quite different, and can even be on another machine. So in those cases the call to `Class.forName()` does not get you the information you are looking for, as it only finds classes included in FitNesse itself and in any plugins installed.

So my question is: what is the added value of doing a slow and unreliable check for fixture classes? Can you recall in which cases you had 'unmarked decision tables'?
Originally you listed quite a few classes by name, but that code got refactored. Given these I wonder whether this code was ever applicable for Slim (as it is currently not used for Fit runs)...
- fit.Comment
- fit.ImportFixture
- fit.ActionFixture
- fit.ColumnFixture

Removing the code altogether seems like a better option than just introducing a cache, but if it serves a purpose: let's add a test to see that it works... :-)

@joprsal You refactored this code recently. Do you know when this code is relevant?

PR #576, opened 2015-01-07
fixes 575

salandur:
When using a Query decission table, not all wrongs are reported in the JUnitRunNotifierResultsListener

when evaluating the missing & surplus rows, only ExecutionResult was based on the surplus rows, not the missing rows. changed the logic to check if either one has rows left and return the ExecutionResult status based on that

Change-Id: I1902a412e6e589982a4da48dd1603698ac641c44

### DIFF
--- a/src/fitnesse/testsystems/slim/tables/OrderedQueryTable.java
+++ b/src/fitnesse/testsystems/slim/tables/OrderedQueryTable.java
@@ -30,7 +30,9 @@ public class OrderedQueryTable extends QueryTable {
       }
     }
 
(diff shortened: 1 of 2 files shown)


amolenaar:
I'm getting a build error:

    [javac] /Users/arjan/Workspace/fitnesse/src/fitnesse/testsystems/slim/tables/OrderedQueryTable.java:33: error: incompatible types
    [javac]     return markSurplusRows(queryResults, unmatchedResultRows);

Can you have a look?

salandur:
I have fixed the compile error.

Issue #579, opened 2015-01-15
Failure Navigation don't work in the test history


bjornpalmqvist:
When checking the test history for a test that contains errors, the failure navigator does not find any failed tests. Hence the navigator can't be used to jump to the failures/errors/exception. This makes it hard to navigate to the failures directly when entering large tests.

To recreate this issue:
1. Create a test that will go wrong
2. Run it
3. Go into the test history for the failed test
4. Try to use the failure navigator

I will state that it can't find any failures, and navigates to test 1 och 0.

bjornpalmqvist:
Created pull request #580 for the changed that we used to fix this issue. I understand that i may need some more work when I'm not familiar with the FitNesse code base.

amolenaar:
By default the failure navigator is only enabled for test pages. Are you using a custom template that enables the Failure navigator in history pages?

PR #580, opened 2015-01-15
Issue #579 - Better filter for failure navigation

bjornpalmqvist:
This fixes the failure navigation for the test history.


amolenaar:
Hi,

When looking through this part of the code I came across a few inconsistencies. I fixed those and fixed in Failure navigator in the history page as well (by applying the `alternating_block` class to the content div.

Your commit put me on the right track. Thanks!

bjornpalmqvist:
Glad that it was helpful!

PR #585, opened 2015-01-18
Switch from commons-lang to commons-lang3

hansjoachim:
Switched to newer version of commons-lang, which basically consistet of tweaking the import statements. Apache has a [migration article/release notes for 3.0](http://commons.apache.org/proper/commons-lang/article3_0.html). It made some backwards-incompatible changes, most of which doesn't seem to affect FitNesse. 

The only one point I spotted was that `StringUtils.isNumeric` will now return false instead of true on empty Strings. I don't know whether the usage of isNumeric in FitNesse has made any assumptions regarding the old behaviour.

I have run `ant` and all tests are still passing.

One surprise was when I tried to remove the old commons-lang dependency, I noticed it was still fetched because velocity depends on it. (Found this through reading `ant dependencies` report.) Due to this I put it back, which means it will still be possible to use in the source code. I realize it's not ideal to have both in parallell, so I wonder if there's a better way to solve this. It might be possible to exclude it from velocity's dependencies, but I haven't tested the consequences of this, nor do I know whether it is a better idea.

### DIFF
--- a/ivy.xml
+++ b/ivy.xml
@@ -32,7 +32,7 @@
 			<exclude org="com.sun" module="tools" />
 		</dependency>
 		<dependency org="org.apache.velocity" name="velocity" rev="1.7" conf="default,maven->default" />
-		<dependency org="org.apache.commons" name="commons-lang3" rev="3.0" conf="default,maven->default" />
+		<dependency org="org.apache.commons" name="commons-lang3" rev="3.3.2" conf="default,maven->default" />
 		<dependency org="org.json" name="json" rev="20140107" conf="default,maven->default" />
 		<dependency org="org.apache.ant" name="ant" rev="1.9.4" conf="default" transitive="false" />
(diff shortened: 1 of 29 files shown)


amolenaar:
Hi,
The downside of using commons-lang3 is that it's introducing another dependency. Commons-lang (2) is required by Velocity. I feel the update is not worth an extra dependency (FitNesse only uses StringUtils and ArrayUtils).

hansjoachim:
No problem, that's perfectly understandable. As mentioned above I only realized commons-lang was a dependency of velocity when I tried to remove it. And at this point I had done the work and verified the tests still passed so I figuered I might as well hear what you thought. 

If nothing else, we know that it should be rather trivial to move to commons-lang3 if this becomes more desirable in the future (for instance if there's a new velocity release which has also switched to 3).

PR #588, opened 2015-01-20
Fixing some SonarQube-issues

hansjoachim:
Not sure if you're aware, but the people behind SonarQube (tool for code analysis) has a public instance where they scan various free and open source projects regularly, FitNesse included. (See http://nemo.sonarqube.org/dashboard/index/390646 for more details)

I took a look at this and started fixing some of the issues. Some of it is nitpick, but a lot are issues which should be resolved, and there seems to be some noise from the bundled Javascript libraries.

This branch is work in progress, but I thought I'd at least let you know the list of issues exists. I've started with some of the low-hanging fruits and plan to work on reducing the amount of open issues. :)

### DIFF
--- a/src/fitnesse/FitNesseVersion.java
+++ b/src/fitnesse/FitNesseVersion.java
@@ -49,6 +49,7 @@ public class FitNesseVersion {
     }
   }
 
+  @Override
   public String toString() {
     return version;
   }
--- a/src/fitnesse/components/ContentBuffer.java
+++ b/src/fitnesse/components/ContentBuffer.java
@@ -92,6 +92,7 @@ public class ContentBuffer {
     tempFile.delete();
   }
 
+  @Override
   protected void finalize() throws Throwable {
     delete();
     super.finalize();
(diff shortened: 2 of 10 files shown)


amolenaar:
I was not aware of this. I like it :).

Do you know if we can tailor the results by adding a config file to the fitnesse repo?

hansjoachim:
Hard to say about the config file. I only found the page from their website and don't know much about the setup/how they run the scans. Thus my question, I wasn't sure whether FitNesse had signed up somehow or if they have simply found a bunch of projects and scan them. Useful either way, though.

I would guess they are probably running the scans using the SonarQube Runner (http://docs.sonarqube.org/display/SONAR/Analyzing+with+SonarQube+Runner). It is possible to run it as an ANT task as well, though that seems to require defining it in build.xml. So I find it a bit unlikely that they would keep a patch for that around and re-apply it on top of latest HEAD every once in a while just to run scans.
If it is indeed the runner, it looks like it should be possible to create a `sonar-project.properties` file with various configuration (http://docs.sonarqube.org/display/SONAR/Analysis+Parameters). Of course this configuration might include information like which SonarQube server you would like to submit the results to and so on, but maybe they just override those when running the scans anyways.

I'll look into it and see if I can find anything (and continue fixing issues :)). Did you have any particular configurations in mind? I've been playing a bit around with SonarQube locally so I would anyways be able to test it out if I can get it working there.

PR #589, opened 2015-01-21
Deploy .gitignore to FitnesseRoot

linquize:
Deploy .gitignore file of FitNesseRoot

This facilitates user to create git repository inside FitNesseRoot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,3 @@
-TestDir
-properties
-updateList
-updateDoNotCopyOverList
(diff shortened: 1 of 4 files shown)


amolenaar:
Hi,

This is a good idea. I think FitNesseRoot/FitNesse should be added as well, for the distribution. Only not for the FitNesse repo itself.

linquize:
`FitNesseRoot/FitNesse` is already deployed before this pull request.

PR #594, opened 2015-01-24
Add https and SSL support for the wiki server and the slim test system

six42:
See FitNesse.UserGuide.AdministeringFitNesse.NetworkSecurity for instructions ow to use it.


amolenaar:
Hi,

This makes for a nice addition. Two questions: 
 * why do users need to create a class? Why not a configuration file (not plugins.properties, since those properties are also available to wiki pages)?
 * Should we prevent `javax.net.ssl.*` properties from being used in wiki pages?

six42:
Hi Arjan,

A) Why a class and not a configuration file:
1. For the intended users of FitNesse it should not be a problem to create a class as FitNesse is a tool which requires programer knowledge. Creating the class for a programmer is not more effort than creating a new config file.  
2. The SSL configuration requires a password. Storing passwords in clear text is not good practice. So we would need to encrypt or obfuscate it. Adding any encryption or obfuscation in an Open Source tool will not really work as everybody can look into the code. So it must happen in private code – having said that a class is required.
3. If somebody doesn't want to create a class he can use the default class and stick to the default password. I would not recommend this therefore I didn't documented it. But it is not much less secure than a simple configuration file. 
4. If somebody wants to store the information in a configuration file he can create once a SslParameters class which does this. So the current design with a class doesn't excludes the use of a configuration file. 

I know it is extra effort but security can't be achieved without effort.

But if you think it is a must have it can be added in this or the next release. 


B) Should we prevent javax.net.ssl.* properties from being used in wiki pages?

Which threat you see here? 
The current code sets and clears the values immediately after creating the SocketFactory. This happens before any test cases are executed.

amolenaar:
> Which threat you see here? 
> The current code sets and clears the values immediately after creating the SocketFactory. This happens before any test cases are executed.

System properties like `javax.net.ssl.keyStorePassword` can be accessed/displayed in Wiki pages, unless we do something about that.

I tried to merge and test the pull request. I added a few changes so the tests kind of work on Mac/Unix. See my amolenaar/fitnesse/ssl-support branch.

I tried to get the tests pass but so far without success. On Mac as well as Windows the test `fitnesse.slim.SslSlimServiceTest` keeps hanging (non-terminated slim server?).

six42:
Thanks for the Mac/Unix changes. 
I will check the tests once more but I am currently short of time, will take a while.
I assume you did "ant unit_tests".

bad49:
Does this change allow the FitNesse server to require SSL Client Authentication to users?  Also can the fitnesse.authentication.Authenticator.authenticate(...) method have a means to get the client certificate, that way a user can be authenticated just by their certificate.

amolenaar:
@six42 I always to `ant release`.

six42:
The foundation is there but not accessible from the outside at the moment.
If you want to add this you need to:

1) in the SocketService constructor call tryCreateServerSocket with third parameter as true to enforce client SSL authentication.

2) Create your Authenticator class which calls SocketFactory.peerName to get the user/owner of the certificate used.

bad49:
I will try to get something to you in the next couple of days.

six42:
Fantastic :)

bad49:
I got an initial go at the client auth.  I still need to write tests.  I am on a mac and hit the same issues amolenaar had with the tests blocking forever.  I got some of the tests working, but I think "fitnesse.socketservice.SslParametersNarcissus" is missing.  That class is referenced in fitnesse.slim.SslSlimServiceTest.  I need that class to know if I did not break anything in my changes.

The changes I made for the ssl clientAuth:
 * Looking for wiki.protocol.ssl.client.auth to be required in the FitNesseContext properties to enable clientAuth
 * Added the peerDn to Request that will be accessible in the fitnesse.authentication.Authenticator.authenticate(...)
 * did a little refactoring because constructors were getting out of hand. 

That is about what I did.  I pushed my changes to my clone of your repo if you want to look at them.  The next few days I will try to write some tests.

six42:
Can you let me know which testcase hangs forever?

I did a fresh install yesterday. On this Installation SlimServiceTest runs for a long time and dies then with an out of memory error. it can't bind to the slim port. There is already another client running on the port. Not really better than blocking forever. ;)
Still need to figure out what is going wrong.

Narcissus is will check.

bad49:
The issue on the hanging on Mac machines is in the following files:

fitnesse.socketservice.SslParametersWiki
fitnesse.socketservice.SslParametersAgent

it happens because they are using windows path separators.  This hangs the SslSlimServiceTest because of the following lines in the SlimServiceBaseTest:

  protected void createSlimService() throws InterruptedException {
    while (!tryCreateSlimService())
      Thread.sleep(10);
  }

  private boolean tryCreateSlimService() {
    try {
      startSlimService();
      return true;
    } catch (Exception e) {
      return false;
    }
  }

The startSlimService() throws an exception every time because it cannot find the keystores because they are using windows paths. If you look at the changes I made in my clone for the files I mentioned should work for both mac and windows.  Once I fixed this I get cannot find the fitnesse.socketservice.SslParametersNarcissus class.  I looked for it and could not find it either that is why I think it was not checked in.

The fitnesse.socketservice.SslParametersAgent is needed for a different test that fails on Macs because of the windows path seperators used.

six42:
I added two more commits. 
Should now work. At least under Windows it is now working for me. Bad49, I copied your path fix. Thanks for this.

Am 13. Februar 2015 23:31:05 MEZ, schrieb bad49 <notifications@github.com>:

bad49:
Your welcome.  I merged in your changes to my clone and the tests passed on my Mac.

six42:
:)

bad49:
I just finished testing the client authentication and it all works.  I also added a section in the NetworkSecurity page on how to enable client authentication.  Everything is pushed to my repo.  Six42 do you want me to pull request this into your repo?  Or should this be a separate pull request?  This side of github is new to me so I don't know what the proper way to go.  Thank you for your help.

six42:
I am neither an expert of Github. I assume that two seperate pull requests are required so that both of us get the credit for the contribution we have done to unclebob.fitnesse. @Arjan do you know this?

arjan:
guys, your developer is called @amolenaar . @arjan is my username, please stop mentioning me, thanks :)

six42:
Oh, sorry.

Am 16. Februar 2015 19:13:33 MEZ, schrieb Arjan Scherpenisse <notifications@github.com>:

amolenaar:
@bad49 can send a pull request to @six42. Then the pull request can be merged. The author information will remain the same. I did a similar thing with #617: send a pull request and let it be merged.
Make sure that when you send a pull request, the left hand side is correct (user/repo and branch).

amolenaar:
Is this PR ready for merge?

bad49:
I sent a pull request to @six42.  Let me know if there is any issues.

six42:
Created a new pull request as this one has merge conflicts. The new one is now finished and can be merged.

PR #602, opened 2015-01-29
Adding support for Android slim client

achmyr:
This changes intended to remove dependency to java.beans.\* package which is not supported by Dalvik VM and add packaging task for create fitnesse-slim.jar which can be added to Android app and execute fitnesse tests using slim.

Sample code to run fitnesse client inside android app

```
import fitnesse.slim.JavaSlimFactory;
import fitnesse.slim.SlimService;

SlimService.Options options = SlimService.parseCommandLine(new String[]{"-v", "-d", "8099"});
SlimService.startWithFactory(JavaSlimFactory.createJavaSlimFactory(options), options);
```

This will open socket on Android phone on port 8099 which can be accessed by Fitnesse server.


amolenaar:
Hi, Nice commit. It would be awesome to have Slim running on android. I noticed a few things, though:

1. The SecurityManager has been completely disabled. I assume it's not available on Android, but it would be nice to leave it in for plain Java tests.
2. The `java.beans` dependency is completely gone. I'm not sure how this will work out for users currently depending on the beans framework (are there any?).

smudgy:
Hi,
My comments below:
1.regarding SecurityManager - you're right, it is not implemented on Android - setSecurityManager() throws Exception, getSecurityManager() returns null.
What would you suggest to put as a condition to distinguish Java/Android runtime? 
I have an example, but personally I don't like this way to check. If you have some other suggestions, I would be happy to use it.
```
 final String vmName = System.getProperty("java.vm.name");
 if ("Dalvik".equalsIgnoreCase(vmName) || vmName.toUpperCase().startsWith("ART")) 
```
2. The java.beans package classes were used specifically in slim.converters package as internal implementation of text<-->Object conversion. Those new classes just copy of the real java.beans implementation, but some methods are not used at all. Would it be better to remove them or leave for compliance? Honestly, I would completely rewrite it just to have cleaner code.

P.S. This code is already used in a huge android project.

amolenaar:
Hi,

1. I was afraid the whole API was missing on Android. If it is there we can skip `setSecurityManager` if `oldSecurityManager` is null (defensive), or explicitly catch the exception on setSecurityManager() (make it a best effort operation / fire and forgive). I'd prefer the latter.

2. The `java.beans` package has a means to register editors (application wide) via the `PropertyEditorManager`. It's intended to support any converter registered this way. Replacing it with an implementation in FitNesse doesn't make sense, since the registered `PropertyEditor`s are not accessible anymore. However, I see that having a hard dependency on these beans-classes doesn't work for Android. Basically, not having `ConverterRegistry` depend on `java.beans` solves the issue, right?

I'd love to see how you make it work (with launching the android app and all). Do you have written a blog about it?

Do you have an example project on which we can experiment?

amolenaar:
@smudgy Do you have a test/demo app for Android?

smudgy:
Hi, Sorry for a delay - was busy writing fitnesse fixtures :) 

I've created test app showing primitive fixture running as a part of android app. Also, updated merge with changes for SecurityManager and with latest master.
Project is here [AndroidFitnesseExample](https://github.com/ACM64/AndroidFitnesseExample)

It is unclear to me still why Long is not supported as a type? Timestamps are great examples why we need it. 

Please shout if you need more info on a project

amolenaar:
Just an update: I plan to look into this, only it's not on the top of my back log.

brucetrask:
I ran the example from smudgy and it works great.  Thanks for that.
Dumb question on a slightly different subject but related.  Any know difficulties with running the Fitnesse Server on Android itself.  I would like to use some FitLibrary features and so am not using SLIM.  I believe to run FitLibrary I would have to run the whole fitnesse server on the Android device. Any known dependencies that would cause problems or should I just give it a go and see what pops up?

smudgy:
Hi,
I've started from intention of making server run on android, but gave up
after couple hours of adding dependencies. It may changed for M, give it a
go and let me know how it goes.
On 23 Apr 2016 20:02, "brucetrask" <notifications@github.com> wrote:

I ran the example from smudgy and it works great. Thanks for that.
Dumb question on a slightly different subject but related. Any know
difficulties with running the Fitnesse Server on Android itself. I would
like to use some FitLibrary features and so am not using SLIM. I believe to
run FitLibrary I would have to run the whole fitnesse server on the Android
device. Any known dependencies that would cause problems or should I just
give it a go and see what pops up?

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/unclebob/fitnesse/pull/602#issuecomment-213806511>

fhoeben:
I just came across this PR. Are you still using FitNesse, still happy with the approach in this PR?

achmyr:
@fhoeben 
Yes, I'm happy with the approach. I've double checked with the latest Android API, and these changes still make sense

build.xml
```diff
@@ -68,6 +68,23 @@
 		</jar>
 	</target>
 
+	<target name="slim-jar" depends="compile-slim" description="generate slim only jar file">
```
fhoeben: Can you convert this to the build.gradle currently in use?

src/fitnesse/slim/converters/LongConverter.java
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2003-2009 by Object Mentor, Inc. All rights reserved.
```
fhoeben: Why this copyright message?

src/fitnesse/slim/converters/PropertyEditorManager.java
```diff
@@ -0,0 +1,117 @@
+/*
```
fhoeben: Why this copyright message?

src/fitnesse/slim/converters/LongConverter.java
achmyr: I believe this copyright was in the [IntConverter.java](https://github.com/unclebob/fitnesse/blob/30b496e330add41ab36b7fa04b21f1a6e8fefecd/src/fitnesse/slim/converters/IntConverter.java), so it was inherited by copy here

src/fitnesse/slim/converters/PropertyEditorManager.java
achmyr: The PropertyEditorManager was copied over from original source in ASF. I'm open for suggestion what to put in here.

build.xml
achmyr: 👍

src/fitnesse/slim/converters/PropertyEditorManager.java
fhoeben: If there are no modifications, just leave it as-is. Otherwise add a line to indicate copyright of FitNesse contributors.

achmyr:
The JVM version update itself is not a problem. The only restriction for Android that it doesn't have the Streams API and once it has been introduced into the code, it may stop working at all.

achmyr:
@fhoeben I've made changes you requested, also, the test-app was updated to the latest Android SDK. Please review

src/fitnesse/slim/instructions/SystemExitSecurityManager.java
```diff
@@ -26,9 +26,19 @@ private SystemExitSecurityManager(SecurityManager delegate) {
   public static void activateIfWanted() {
     if (isPreventSystemExit()) {
       SecurityManager currentSecMgr = System.getSecurityManager();
-      SystemExitSecurityManager systemExitSecurityManager = new SystemExitSecurityManager(
+      if (currentSecMgr != null) {
```
fhoeben: I believe the security manager will be null in plain java also, therefore the system exit one will not be added.
What I've seen (in slf4j) is a check for android by getting system property: `java.vendor.url` turning that into lowercase and checking whether it contains `android`...

fhoeben:
On 2nd thought there seems to be little point in having a custom PropertyEditor interface. If one can adapt an existing PE to implement the new interface, one might as well just implement the Converter interface, and register the class directly, right?

achmyr:
All the changes are in now.

PR #606, opened 2015-01-30
Some fixes to insert image

linquize:
Some fixes to insert image.


amolenaar:
Thanks for, ehhm, cleaning up after me :).

PR #608, opened 2015-02-02
quick fix for javadoc errors in java8

denisko:
Java 8 is very strict on javadoc usage of HTML. This pull request fixes most of the errors, and some warnings.


amolenaar:
Thanks :)

PR #617, opened 2015-02-07
Plugins without properties file changes

fhoeben:
Allow plugins to be loaded without the need to edit plugins.properties.

This change allows plugins to register themselves using Java's ServiceLocator mechanism, instead of the user having to edit the property file of his/her FitNesse installation.

To use this new feature a plugin must:
* Create a class implementing fitnesse.plugins.PluginFeatureFactory (probably easiest by subclassing fitnesse.plugins.PluginFeatureFactoryBase) and in this class return the symbols, Slim tables, etc that the plugin contains from the appropriate methods
* Add a file called fitnesse.plugins.PluginFeatureFactory in its jars META-INFA/services directory, containing the fully qualified class name of the PluginFeatureFactory implementation

When a jar containing this file and class is placed in FitNesse's plugin directory it will be automatically registered when FitNesse loads its plugins.

The existing mechanism (using the plugins.properties file) still works. Information from the property file is processed first before the PluginFeatureFactories are used.


amolenaar:
I think it would be nice if `PluginFeatureFactory` has the same interface as the current "plugin" classes (although that interface is weak and discovered through introspection). That way the transition to discoverable plugins is simpler and the interface is more concise.

Do we really need to set the component factory? It's probably better to pass it as argument.

fhoeben:
I decided against the 'old' plugin interface because it puts the responsibility of actual registration with plugin. With the new interface the plugin author only has to return what they created, How to actually use that is internal to FitNesse. Furthermore not all existing plugins use that interface (they might just provide a symbol (like the maven-classpath one) or a slim table. Then they don't need (or currently have) the required methods. They can just override the one method they need and return their Symbol.

The 'old' interface is still fully supported, they can be returned from the getPlugins() method. So plugins that have that can just override that method from PluginFeatureFactoryBase and be done with almost no work also.

I think this interface will make for an easy transition and gives a nice separation of concerns between the PluginsLoader and the actual plugins.

PR #618, opened 2015-02-07
Prefer isEmpty() when checking for emptiness

hansjoachim:
Converts various instances of (!)size > 0 to use isEmpty instead. In most cases, it's more readable, it doesn't need to count all the elements like length() so there's a (probably not noticable) performance gain and it's one step closer to a happy SonarQube.

At the same time I did some simplifications and minor fixes in related parts of the code. See individual commits for details.

### DIFF
--- a/src/fitnesse/testsystems/slim/tables/QueryTable.java
+++ b/src/fitnesse/testsystems/slim/tables/QueryTable.java
@@ -113,7 +113,7 @@ public class QueryTable extends SlimTable {
     markMissingRows(unmatchedTableRows);
     markSurplusRows(queryResults, unmatchedResultRows);
 
-    return unmatchedTableRows.size() > 0 || unmatchedResultRows.size() > 0 ? ExecutionResult.FAIL : ExecutionResult.PASS;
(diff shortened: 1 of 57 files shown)


src/fitnesse/testsystems/slim/tables/QueryTable.java
```diff
@@ -113,7 +113,7 @@ protected ExecutionResult markRows(QueryResults queryResults, Iterable<MatchedRe
     markMissingRows(unmatchedTableRows);
     markSurplusRows(queryResults, unmatchedResultRows);
 
-    return unmatchedTableRows.size() > 0 || unmatchedResultRows.size() > 0 ? ExecutionResult.FAIL : ExecutionResult.PASS;
+    return !unmatchedTableRows.isEmpty() || !unmatchedResultRows.isEmpty() ? ExecutionResult.FAIL : ExecutionResult.PASS;
```
hansjoachim: I considered flipping this ternary to (isEmpty && isEmpty) ? PASS : FAIL, but I'm not sure whether that makes it more readable or less.

(Same thing with OrderedQueryTable above.)

Issue #523, opened 2014-09-16
Enhancement: historize error logs the same way as test results


jfpoilpret:
Currently, running the same test suite several times will historize the results as expected (one XML file per test run, the file name is based on test run time and is stored in `FitNesseRoot/files/testResults` test subdirectory).

However the stdout/stderr that go to `FitNesseRoot/files/ErrorLogs/...` for the test suite get overwritten after each execution of that suite. Of course, if FitNesse was launched with WIKI pages version history (ie without `-e 0` command-line option), then the log will be historized as a normal WIKI page.

However, when you browse the history of runs of a given test page and select any result in that history, there is no direct way from teh test result page to see the error log for that execution.

You generally have to type the URL manually, e.g. `http://localhost/ErrorLogs.MyFirstSuite` and then find out the right version that matches the date of the test run.

It would be nice if FitNesse could:

1. Historize ErrorLogs the same way as test run results are stored (potentially inside `FitNesseRoot/files/testResults` same location as the test result XML themselves)
2. Have a link to the right ErrorLogs version from the test result page you got from the history

This is particularly useful when you have continuous builds running your tests, and you see build (n-1) failed but not build (n) and you want to see the test results along with its log for build (n-1). 

Today you can't do that easily (i.e. you need to type the URL then find the right version, as mentioned above), 

Sometimes you may even be completely unable to do that, i.e. if you launched FitNesse with `-e 0` option (which is what I do for my setup).

amolenaar:
fixed by #627.

PR #627, opened 2015-02-11
Add execution log to test history

amolenaar:
This PR lets FitNesse store the execution log (executed command, stdout, stderr) in the test history. This way the test execution log is always available (e.g. for analysis) along with the test results.

As a result ErrorLogs files are no longer created.

See #523.


ranjith1882:
Hi,

I have updated with latest fitness version, after this upgrade, when I click on "Tests Executed OK" link to see execution log, an empty page is showing and i am not able to see any log info.

amolenaar:
Hi, I checked, but was not able to reproduce. Does the same error happen when you execute a test from the FitNesse acceptance test suite?

ranjith1882:
yes, same issue with fitnesse acceptance test suite also.

ranjith1882:
Hey,

Its workinf fine, Issues is because of some template pages and css files are overrided in existing code base, replace with this version resource files and issue got resolved.

Thanks.

PR #636, opened 2015-02-15
Allow for test history cleanup after execution

amolenaar:
Changed the SuiteResponder so it can do (optional) cleanup actions on the
suite's test history. This prevents the test history from filling up the file
system. In the configuration file a property `test.history.days` can be added
and assigned a number, the number of days for which to maintain the history.

Is a request parameter is also desired? It might make it easier from CI's to
execute tests and clean up afterwards.

Also the user manual still needs to be updated.


floppyrom:
Great!

It would be very useful that after the clean-up happens and old stuff gets deleted, we should still keep the last successful test run log, even if this is older than those certain number of days. Like that you can always have a reference in your test history with an OK run!

Issue #610, opened 2015-02-04
Replacement of symbolic links is no longer working


ufroehling:
I have a symbolic link with name=XyTeam and path=.MyApp/XyTeam/TrunK. Now i try to replace it i by entering the existing name and the new desired path = .MyApp/XyTeam/ReL11 and click on the Create/Replace button. I then get the error message "...  already has a child named XyTeam ". 

Version is the latest from 20150114.
In our live version 20131110 the replacement is still possible and used regularly.

mwarhaftig:
Hi Ufroehling, thanks for the detailed info on the bug, was able to recreate and will look into the cause.

mwarhaftig:
Commit b47bea tightened the Symbolic Link behavior but also introduced the bug you encountered.  Submitted pull request to resolve.

mwarhaftig:
Did a little more research and don't think b47bea introduced the issue. Most likely it was later changes to FileSystemPageFactory.createSymbolicPage (bug only started happening with release 20140630). The fix remains the same.

amolenaar:
fixed by Matt

PR #639, opened 2015-02-18
Issue #610 - Allow Sym Link rename/replace on existing Sym Link.

mwarhaftig:
Commit b47bea tightened the Symbolic Link behavior but also introduced this bug by checking SymbolicLinkResponder.isValidWikiPageName() in SymbolicLinkResponder.addSymbolicLink().

Fix is to allow Symbolic Link rename/replace to proceed if the target page is itself a symbolic link but error if target page is actually child WikiPage.


mwarhaftig:
Did a little more research and don't think b47bea introduced the issue.  Most likely it was later changes to FileSystemPageFactory.createSymbolicPage (bug only started happening with release 20140630).  The fix remains the same.

PR #640, opened 2015-02-18
Display port of FitNesse-server in test-result-pages

t-echnics:
When running multiple FitNesse-server with different ports on one machine, you need to know, which server created the test-results.

### DIFF
--- a/src/fitnesse/reporting/history/ExecutionReport.java
+++ b/src/fitnesse/reporting/history/ExecutionReport.java
@@ -16,6 +16,7 @@ import fitnesse.util.XmlUtil;
 public abstract class ExecutionReport {
   private String version;
   private String rootPath;
+  private String port;
   private TestSummary finalCounts = new TestSummary(0, 0, 0, 0);
   public Date date;
   private long totalRunTimeInMillis = 0;
@@ -24,9 +25,10 @@ public abstract class ExecutionReport {
     version = new FitNesseVersion().toString();
   }
 
-  public ExecutionReport(FitNesseVersion version, String rootPath) {
+  public ExecutionReport(FitNesseVersion version, String rootPath, String port) {
     this.version = version == null ? "null" : version.toString();
     this.rootPath = rootPath;
+    this.port = port;
   }
 
   public void tallyPageCounts(ExecutionResult result) {
@@ -47,6 +49,8 @@ public abstract class ExecutionReport {
       return false;
     else if (!StringUtils.equals(version, e.version))
       return false;
+    else if (!StringUtils.equals(port, e.port))
+        return false;
     else if (!DateTimeUtil.datesNullOrEqual(date, e.date))
       return false;
     else if(!finalCounts.equals(e.finalCounts))
@@ -69,6 +73,7 @@ public abstract class ExecutionReport {
   }
 
   protected void unpackCommonFields(Element documentElement) {
+    port = XmlUtil.getTextValue(documentElement, "FitNessePort");
     version = XmlUtil.getTextValue(documentElement, "FitNesseVersion");
     rootPath = XmlUtil.getTextValue(documentElement, "rootPath");
     String dateString = XmlUtil.getTextValue(documentElement, "date");
@@ -109,6 +114,10 @@ public abstract class ExecutionReport {
 
   public String getVersion() {
     return version;
+  }
+
+  public String getPort() {
+    return port;
   }
 
   public long getTotalRunTimeInMillis() {
--- a/src/fitnesse/reporting/history/SuiteExecutionReport.java
+++ b/src/fitnesse/reporting/history/SuiteExecutionReport.java
@@ -29,8 +29,8 @@ public class SuiteExecutionReport extends ExecutionReport {
     unpackXml(xmlDocument);
   }
 
-  public SuiteExecutionReport(FitNesseVersion version, String rootPath) {
-    super(version, rootPath);
+  public SuiteExecutionReport(FitNesseVersion version, String rootPath, String port) {
+    super(version, rootPath, port);
   }
 
   @Override
--- a/src/fitnesse/reporting/history/SuiteHistoryFormatter.java
+++ b/src/fitnesse/reporting/history/SuiteHistoryFormatter.java
@@ -35,7 +35,7 @@ public class SuiteHistoryFormatter extends BaseFormatter implements Closeable {
     super(page);
     this.context = context;
     writerFactory = source;
-    suiteExecutionReport = new SuiteExecutionReport(context.version, getPage().getPageCrawler().getFullPath().toString());
+    suiteExecutionReport = new SuiteExecutionReport(context.version, getPage().getPageCrawler().getFullPath().toString(), "" + context.port);
     totalTimeMeasurement = new TimeMeasurement().start();
   }
 
--- a/src/fitnesse/reporting/history/TestExecutionReport.java
+++ b/src/fitnesse/reporting/history/TestExecutionReport.java
@@ -23,8 +23,8 @@ import java.util.List;
 public class TestExecutionReport extends ExecutionReport {
   private List<TestResult> results = new ArrayList<TestResult>();
 
-  public TestExecutionReport(FitNesseVersion version, String rootPath) {
-    super(version, rootPath);
+  public TestExecutionReport(FitNesseVersion version, String rootPath, String port) {
+    super(version, rootPath, port);
   }
 
   public TestExecutionReport(InputStream input) throws IOException, SAXException {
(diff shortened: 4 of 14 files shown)


amolenaar:
I'm not clear on your use case. If you're running multiple servers, each server has it's own FitNesseRoot, right?

t-echnics:
No, actually not.
We are running multiple servers with different configurations (network-, database-connections, etc.) e.g. for test/production-systems. The configurations are passed as system-properties.
The tests itself are the same - the FitNesseRoot is shared among all servers.
In this constellation this patch is very helpful and so I did it some years ago.
With the last updates these files often changed and I had to refactor my patch over and over...

linquize:
Could you squash some similar commits (SuiteExecutionReport)?

t-echnics:
Sorry for these multiple commits - I was really struggling with the git-plugin of my eclipse.
I was not able to squash any commits, when already pushed to my fork...
Finally I used git directly from command line and now used just only commit comment, as all the changes are done due to displaying the port on the test-result-pages.

amolenaar:
Wouldn't it be more useful to "tag" a test run as being a run against an acceptance or a production environment? When you display only the port number, you still need to figure out which port maps to which environment.

Also, wouldn't it be helpful to display the target environment on the wiki pages as well?

t-echnics:
We are using FitNesse for approx. 10 years now, while using this feature for several years (>5) without any further needs and running approx. 15 servers on different ports on the same machine.
You just log into the server running on the specific port and there you are... nothing else is needed.
You don´t need to tag anything and you don´t have to show anything special on the wiki-pages...
Each server is identified by the port - each server has its special logo (prod, test,...) displayed via css - so nothing more is needed.
You just need differ the test-results stored on the common space...

amolenaar:
I said "tag" which is kind of misleading. I ment an identifier. Now you're using the port as the identifier in the test history. I'm just curious if this approach can be generalized somewhat. Using a port is so, well, low level(?).

> Each server is identified by the port - each server has its special logo (prod, test,...) displayed via css - so nothing more is needed.

This is what I ment: an identifier that tells you on which server you are. How do you load the different css then? via a system property?

t-echnics:
I don´t quite understand your approach of generalizing something just so simple - in my opinion it is already very generalized this way.By just displaying the port like the version in the test results you exactly know which instance did run this test. The port is needed to connect to the corresponding FitNesse-server - so your "low level info" is a mandatory info... If showing an other info instead, you would need some sort of mapping (to the port), which would  not gain any value (for us), but would cause extra effort and make it more complex.

We really did think about our approach - it is just so simple and works without any problems or further needs, so keep it simple, stupid...without the need of much to changed - also for future versions.

The different css is just loaded via plugins.properties, where every fitnesse-server-instance has its own one.

six42:
Hi t-echnics and Arjon

In my opinion t-echnics is right that more information should be displayed in the test history about the test environment used.
On the other hand Arjon is right that the current implementation is very specific to your needs t-echnics and doesn't gives any benefit to other users. 

Possible solution:

Add another special page (TestSystemSetUp) to Fitnesse. This page is run once each time a test system is started. This page captures information about the test environment (like the port number of fitnesse, the fitnesse version, the version of my code, the name of the person running the test, ...)


Am 2. März 2015 10:26:18 MEZ, schrieb t-echnics <notifications@github.com>:

t-echnics:
I don´t think that this very minor change would cause any trouble, but could help someone else like me/us. (What about the version of FitNesse itself - this information is also displayed in the test-history - I would guess that this information is not used by the majority of users...)

If you run multiple FitNesse-servers on one machine, you have to use different ports.
Why not displaying this information in the test-history?
My proposed change is only displaying the already existing information in the test-history.
I wouldn´t like to use a special setup-page for this simple/native information about the FitNesse-server itself - it should be created by itself, like displaying the version.

Regarding your examples: Why would you like to display the name of the person running the test?

amolenaar:
Regarding the port number: since you already mentioned you change the theme by loading FitNesse with different configuration files, I think it would make sense to make the directory where test results are stored also configurable. That way you can define one directory per environment (= FitNesse instance). Also, you do not have to dig through the test results in order to find the ones for the environment you're looking for. 

I don't feel much for adding (yet another) special page.

We already have an issue for adding user information to the test history.

To get back to the version/user information: those have functional value. Things change between FitNesse versions, so I want to know which version I used to run a set of tests. Likewise, it tests fail it would be nice to know who executed those (what did he change, for example, and why?). 

A port on the other hand is a technical feature of the application (of the wiki server only). It's hard to explain, but from an engineering point of view, for me it just doesn't feel right (it smells). We're using the wrong tool (=port) to solve a problem (running tests for multiple environments on one wiki). I can see it solved your need. But it feels like a workaround, not a solution. Would it be an improvement to make the test results location configurable?

t-echnics:
There is no problem to be solved about using one wiki for multiple servers/environments - it just works without any problems, there is no need to change anything. From my point of view, showing the port number does not smell. It just shows the (actually needed) information which is used anyway, which can be useful - I already explained why...

Things do change in FitNesse-versions, but you also do not need to know this information in the test-history-pages. I assume you know which version you are running and I assume, if you see a failed test, you fix it (for interactive testing)... (You could also figure out the previous run version, by using the timestamp, if you previously changed the version (for automatically testing)) But of course, showing the version might make life easier.
Same for the port number, just imagine running multiple servers on one machine. You have to choose a different port number, otherwise you are not able to run them! So why shouldn´t this information be displayed...

Why should anyone need to know the name of the one which did run the test?
Running a test does not change the result. Modifying a test (or the domain code) could change the result, but the responsibly person for this should be clear. Showing the name of the user who started the test is unneccessary information, which also may lead to the wrong person... 

Using different directories for each server would be possible, but does not gain a benefit for us - but we would have to manage multiple directories, which wouldn´t be very nice, I guess...

six42:
I think we agree that everybody wants to see different information about the test run. And there are good reasons for this and there is no wrong or right.

So we should have a flexibel solution.

The good news is there is already one. :-)

The "SuiteSetUp" page can do the job.
Create one in your root folder with the content:
!2 Fitnesse Port ${FITNESSE_PORT}


... and you get what you want.

Give it a try and let me know what you think.

The rendering of the suite history might need to be alligned to the rendering of the test page history.

If you like it some hardcodings for fitnesse version, date and port could be removed.

t-echnics:
I am a bit wondering, why a pull request is discussed like a feature request (to find a solution for a possible, theoretical usecase). 

Of course one could also imagine different approaches to get the port number displayed.
But why spending time on figuring out different ways of achieving the goal - there is already a working implementation, I just wanted to share...

I wanted to keep things simple - using the given implementation. (KISS)
I don´t want to get one information this way and the other that way.
One could also imagine to refactor the whole testresults-history-thing in the future, but actually I think there is currently no need for that.

amolenaar:
A PR is a feature request, with code attached. I wonder what you mean by

> but we would have to manage multiple directories, which wouldn´t be very nice, I guess...

@six42: that's a nice and simple solution :)

t-echnics:
In my opinion a PR is way more than a FR.
A PR is a FR with a full working implementation, with an existing use case - no theoretical nice to have feature, which may not be used at all in the future.

When using/configuring different locations for the test-results (one of your proposals), we would also need to manage the different locations - e.g. notifications/backup/... nothing unsolvable, but unneccessary overhead, so I do not prefer this way.

When using setup-pages (one for suite and one for test!), you would get the version displayed by code and the port by the wiki-page rendering. Of course it is possible, but these are two different ways and the port is displayed not in the same way - e.g. the port is displayed below the version, not in the same row. Personally I would prefer the same way of displaying the server information and get this right out of the box without the need to explicitly setting this info in your setup-pages - so I would call this a workaround.

amolenaar:
> When using/configuring different locations for the test-results (one of your proposals), we would also need to manage the different locations - e.g. notifications/backup/... nothing unsolvable, but unneccessary overhead, so I do not prefer this way.

You can put the directories side by side. The store requirements themselves do not change. If you define `test.history.days=x` in your configuration file, results older than *x* days will be cleaned up automatically.

I see no need to add port field.

t-echnics:
It´s obvious to me, that the store requirements do not change regarding the size, but the locations does change (different directories) and I want to have one single directory.
I really doubt, that anybody is going to use your test.history.path property - I do not have a need for that.

Strange thing to tell me I am just using a "smelly" workaround, when using the same mechanism like for showing the version, but showing the port via wiki-page-rendering is quite nice...

Sorry to say, but if you do not want to pull requests and doubt their needs, but implement stuff you think people need, although there was no request for this, then this discussion was 3 weeks of timewasting and you should just ignore further pull requests...

amolenaar:
I'm sorry you feel that way.

PR #643, opened 2015-02-20
see mail from rene jehle

reneintersys:
Hi  Mike Stockdale,
 
I wanted to send this E-Mail to Uncle Bob, but his Mail-account is out of order. 
 
Can you handle our contribution ?
 
we use FitNesse very extensively and like to contribute.
 
We wrote some widget code which support often needed functions.
 
Doc below is in german lang, I can translate if you like. Java Code is not really of best quality, but it works.


jediwhale:
We need to improve the code a bit before we merge it.
-  All code in FitNesse must have unit tests.
- We do not use Java doc style comments
- Do not swallow exceptions

amolenaar:
You can also package them separately as a plugin and add a reference to fitnesse.org.

PR #648, opened 2015-02-22
Sync with unclebob:master to identify conflicts

six42:


six42:
sorry wrong direction :(

PR #649, opened 2015-02-23
Add https and SSL support for the wiki server and the slim test system

six42:
See FitNesse.UserGuide.AdministeringFitNesse.NetworkSecurity for instructions ow to use it.


amolenaar:
@six42 @bad49 What's your real name so I can mention it in the release notes?

bad49:
Brian Davis

Issue #656, opened 2015-02-27
Better message on slim.timeout


fhoeben:
When the slim.timeout period is exceeded an error is sent to the log of the wiki process
`SEVERE: Killing slim client as exception occurred:java.net.SocketTimeoutException: Read timed out`

And in the wiki page no results of the current table are shown, but 'Errors Occurred' is shown on the top of the page. Clicking on that shows the execution log where an 'Internal Exception:' is shown: `Read timed out`

This is not very helpful (as indicated in #653). 

* The message should mention the `slim.timeout` property as a way to address this, and the current timeout value. 
* Furthermore I would prefer the resulting wiki page to be similar to what happens on a 'StopTest'  or 'StopSuite' exception. Meaning that all the tables are rendered, and all results that were obtained are shown and all assertions not executed marked as such. This also gives the user an indication when exactly the timeout occurred.

amolenaar:
I think the web interface is a lot more informative now, with #660 merged in.

Issue #653, opened 2015-02-26
"Read timed out" killing Slim client with FitNesse v20150225


reprive:
I've built the last FitNesse release 20150225 and am getting random Read Timed out exceptions on Slim tests, which completely kill the test pages (they don't get finished - not even with errors/exceptions). The status bar becomes "Testing was interrupted and results are incomplete. Test Pages: 0 right, 0 wrong, 0 ignored, 0 exceptions     Assertions: 0 right, 0 wrong, 0 ignored, 0 exceptions (11.499 seconds)".

This happens on tests that work perfectly on v20150114. I haven't been able to figure out what exactly causes this, but the tests that have the issue are not very fast (driving browsers). 

Has a new timeout value been introduced recently? If so how can I set that to a larger value?

I see in the log: 
SEVERE: Killing slim clint as exception occured:java.net.SocketTimeoutException: Read timed out
java.net.SocketTimeoutException: Read timed out
        at java.net.SocketInputStream.socketRead0(Native Method)
        at java.net.SocketInputStream.read(Unknown Source)
        at java.net.SocketInputStream.read(Unknown Source)
        at java.io.BufferedInputStream.fill(Unknown Source)
        at java.io.BufferedInputStream.read1(Unknown Source)
        at java.io.BufferedInputStream.read(Unknown Source)
        at java.io.FilterInputStream.read(Unknown Source)
        at util.StreamReader$2.read(StreamReader.java:218)
        at util.StreamReader.readUntilFinished(StreamReader.java:140)
        at util.StreamReader.preformRead(StreamReader.java:119)
        at util.StreamReader.readBytes(StreamReader.java:77)
        at util.StreamReader.read(StreamReader.java:70)
        at fitnesse.slim.SlimStreamReader.getLengthToRead(SlimStreamReader.java:34)
        at fitnesse.slim.SlimStreamReader.getSlimMessage(SlimStreamReader.java:55)
        at fitnesse.testsystems.slim.SlimCommandRunningClient.invokeAndGetResponse(SlimCommandRunningClient.java:168)
        at fitnesse.testsystems.slim.SlimTestSystem.processTable(SlimTestSystem.java:124)
        at fitnesse.testsystems.slim.HtmlSlimTestSystem.processAllTablesOnPage(HtmlSlimTestSystem.java:48)
        at fitnesse.testsystems.slim.SlimTestSystem.runTests(SlimTestSystem.java:101)
[...]

PR #660, opened 2015-03-01
Remove global slim timeout during testing

### LINKED ISSUE #657: slim.timeout default value
The current default value for Slim timeout is 10 seconds. This is quite short, can this be set to 30 seconds or a minute?

six42:
This fixes #657 #656 #653 
also fixes a bug with multiple arguments in SLIM_FLAGS.

I started preparing automatic test cases. They will follow shortly.


amolenaar:
I created a few acceptance tests in the timeout branch on my fork. That might be useful.

amolenaar:
Are you limiting the slim.timeout property to only be used while loading the test system?
I would expect it to be used when executing tests as well.

six42:
I exected this as well. But testing shows it is not the case. There is a workaround using SLIM_FLAGS. But also then the timeout doesn't seems to be useful as every testcase will fail afterwards.
See HowtoEnforceA_TimelimitPerTestCommand

I added the usage of slim.timeout while waiting for the SlimHeader message. The remaing usage of slim.timeout is unchanged.

six42:
Fixed the merge conflict to allow automatic merge.

src/fitnesse/testsystems/slim/SlimClientBuilder.java
```diff
@@ -92,9 +94,10 @@ protected String defaultTestRunner() {
     	arguments = ArrayUtils.add(arguments, "-ssl");
     	arguments = ArrayUtils.add(arguments, useSSL);
     }    	
-    String slimFlags = getSlimFlags();
-    if (!"".equals(slimFlags))
-    	arguments = ArrayUtils.add(arguments, slimFlags);
+    String slimFlags[] = getSlimFlags();
+    if (slimFlags != null)
+    	for (String flag : slimFlags)
+    		arguments = ArrayUtils.add(arguments, flag);
```
amolenaar: The method ClientBuilder.parseCommandLine() could be used to parse the flags. That will take into account flags containing quotes.

src/fitnesse/testsystems/slim/SlimClientBuilder.java
```diff
@@ -179,12 +182,12 @@ protected String determineSlimHost() {
...
   }
 
-  protected String getSlimFlags() {
+  protected String[] getSlimFlags() {
     String slimFlags = getVariable("slim.flags");
     if (slimFlags == null) {
       slimFlags = getVariable(SLIM_FLAGS);
     }
-    return slimFlags == null ? "" : slimFlags;
+    return slimFlags == null ? new String[] {} : slimFlags.trim().split("\\s");
```
amolenaar: There is a smarter parser in ClientBuilder.parseCommandLine(). That will take into account quoted values.

six42: Thanks for the hint. Changed it.

PR #668, opened 2015-03-06
MarkupPageInclude: one funny typo fixed: change -> chain

TiborVyletel:


amolenaar:
Nice catch. Thanks for correcting.

PR #670, opened 2015-03-07
Change See widget to handle aliases

markhop:
Whilst updating some of the documentation I noticed that the See widget does not handle aliases correctly and due to this the !see [[alias text][AliasLink]] is not rendered properly.

I am not too enamoured with the current solution due to the amount of casting so if someone could suggest a better way of doing this I'd be grateful.


jediwhale:
I don't think this will work.  e.g. try a test with:
!see [[starts like alias but is not

jediwhale:
I think you can avoid the cast with next.getType().getWikeRule().parse(...)

markhop:
@jediwhale  thanks form the feedback.  I've amended the code to remove the casting as advised and also added a few unit tests for a malformed alias.  All seems to be fine although I'm still not 100% convinced this is correct.

jediwhale:
Looks pretty good.  Maybe has a method isNothing so you don't have to dig inside it looking for nulls.  (Maybe was my clumsy attempt to wrap null values - I've since found Nat Pryce has a much better implementation: http://www.natpryce.com/articles/000776.html)

Maybe<Symbol> alias= next.getType().getWikiRule().parse(next, parser);
if (!alias.isNothing())
  return new Maybe<Symbol>(current.add(alias.getValue()));

markhop:
Again thanks for the suggestion.  You were right it was the null checking that was bugging me and I must admit to have been confused with all the Maybe<T>'s lying around.  The fog is now beginning clear thanks!

PR #679, opened 2015-03-12
Make the "bootstrap" theme the default

amolenaar:
The bootstrap theme looks nicer and, since the website is using the same theme, it makes sense to use this theme also for (first time) downloaders.


linquize:
Bootstrap UI is cool.

PR #681, opened 2015-03-15
Remove extra character which broke half of the paragraph

hansjoachim:
With the extra character in place, the remainder of the paragraph was not shown. 

I stumbled across this after a colleague asked if I knew why both FitNesse and FitLibrary contains the fit package with various fixtures. I looked a bit into it, but didn't find any definitive answer, though I registered FitLibrary had a similar question on their mailing list recently[1]. Is there a historical reason why both projects contain this package?

[1] http://sourceforge.net/p/fitlibrary/mailman/fitlibrary-user/?viewmonth=201501


amolenaar:
That was before my time. If it's in FitLibrary, I'd expect it can be removed from FitNesse (?)

hansjoachim:
>If it's in FitLibrary, I'd expect it can be removed from FitNesse (?)

I don't know. I was mostly curious why the package appears in both projects. It does sound like it should be in only one place, but it's not clear to me which one is the authoritative version of the package.

PR #687, opened 2015-03-21
Revert old behaviour for Slim String converter

amolenaar:
String converter used to pass strings as is. The 20150226 version converts empty Strings to null values.

This PR reverts the old behaviour.

### DIFF
--- a/src/fitnesse/slim/converters/StringConverter.java
+++ b/src/fitnesse/slim/converters/StringConverter.java
@@ -13,6 +13,6 @@ public class StringConverter implements Converter<String> {
   }
 
   public String fromString(String arg) {
-    return !StringUtils.isBlank(arg) ? arg : null;
+    return arg;
   }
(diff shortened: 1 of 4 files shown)


mgaertne:
What about a feature flag to smooth out the transition here?

Best Markus

-- 
Dipl.-Inform. Markus Gärtner
Test-Driven Development

http://www.shino.de/blog
http://www.mgaertne.de
http://www.it-agile.de
Twitter: @mgaertne

davidmagill:
Are you suggesting that this should be the intended behavior? If so, I would ask the following: 

If the setter method expects a string value, why should an empty cell pass it null? Also, I'm not sure how you would pass an empty string if that was required behavior... 

The new behavior breaks backward compatibility.

amolenaar:
Merged, as agreed by @antoine-aumjaud.

PR #704, opened 2015-04-02
Added ability to show first title (!1...!6) of the page instead of its name in the TOC

andrey012:
I propose to add ability to use first title in the TOC. My main issue, which caused this feature was inability to use nonASCII chars in the TOC, because nonASCII can not be used for page names. 

Here is the demo: 
http://andrey012.github.io/fitnesse/FirstTitleInToc.mp4


six42:
Nice Feature!

Which name is shown in the test results?
How fast is this, for example if you apply it to the FullReferenceGuide?

andrey012:
six42 wrote:
> Nice Feature!
>
> Which name is shown in the test results?
Sorry, didn't understand question. If you mean the acceptance test, that I've added - FitNesse.SuiteAcceptanceTests.SuiteWidgetTests.TestContentsWithFirstTitles - then, the idea is, that only GrandChild is filled in this test, so we give title to GrandChild - !1 before You're too young to be grandparents
*
Next create a grandchild*
script
start 	Page Builder
line 	!1 You're too young to be grandparents
page 	ParentPage.FirstChild.GrandChild



And then verify below:
Response Examiner.
type 	pattern 	matches?
contents 	a href="ParentPage.FirstChild" class="suite">FirstChild</a 	true
contents 	a href="ParentPage.FirstChild.GrandChild" class="static">You're too young to be grandparents</a 	true




> How fast is this, for example if you apply it to the FullReferenceGuide?

Good point.

I optimized it a bit now - and now to stop translation after we have got first title. Now, when working out of disk cache (repeating several tries - all files are already cached) - it is 1.17 vs 0.87 seconds on my i3 machine. We can store a copy of first title in properties in a denormalized manner, I didn't actually go this way, because I do not see any denormalization in the FitNesse so far, and normal data models add simplicity and easiness (for example you can modify contents.txt in other software without need to keep anything in sync).

AG

six42:
I mean the test history:

In the TOC I see:
Deutschland
Grossbritanien
Vereinigte Staaten

It is the same in Test History? Or do I see there
Deutschland
United Kingdom
Vereinigte Staaten

IMHO display name should be identical on all pages.
Should be a property of each page. Could be derived from the first title on the page or given explicit. 
Managing the display name via a TOC parameter feels for me that the global look&feel becomes inconsistent.

If somebody adds a feature to search pages by title. Must he search the page name or by the name he sees in the TOC.

andrey012:
No, this way I have done it - it only affects TOC, so it a cosmetic tweak for TOC rather then functional.

In order to do all the features you suggest (search, test history, etc) - I suggest to add another checkbox in page properties - "Use first title as name of page".
And in this case - I'll try to patch the getName() function in order to honor this feature.
Do you agree with such approach?

AG


six42 wrote:

andrey012:
Please let me know what you think about suggested (below) approach. I'll need this feature anyway (this way as I have done it now, or other way I suggest below), and I'm interested in merging my tweaks into mainstream, in order to use FitNesse build from Maven repository instead of maintaining my own build for my projects.

Thank you.

AG

Andrey Grinenko wrote:

six42:
Hi Andrey,

As a user I would love this feature as you described it below with the flag.

I can't comment on the technical solution. Have not looked much at this area of the code. Somewhere the distinction between filename and display name must be done. Feel you know better than me.

What do others think?

amolenaar:
Just wondering.. Are characters like the ß allowed in URL's? In that case we can just broaden the allowed characters for page names.

andrey012:
Arjan, I've made some experiments - tweaked JS and backend validation, added urldecode for processing request URL, tested result under Linux and Windows - so far it works. To be honest I was afraid, and still a bit afraid of using nonAscii in filenames on filesystem - it is too OS- and locale-dependant.
First thing, that comes on mind - URLencoding filenames themselves - is not a good way, because some filesystems has particular path length limit and urlencoding triples length.

The cons of this approach - is that we still can not have arbitrary page name in TOC and other places. Arbitrary - I mean having not only letters but other characters which are not wiki name.

So, if you could suggest a decision which way to go - I'll try to code it.
Now we have 3 options:
1. as it is done now - only for TOC
2. as six42 suggested - add option to use first title everywhere and make it property of a page
3. as you suggest - allow particular non-ascii letters in wiki names/paths

AG

Arjan Molenaar wrote:

andrey012:
After thinking a bit more, I suggest to keep my current implementation.

1. first of all it is an option. It is switched on by flag, if it is switched off - it does not add any impact to current system at all (no performance impact, no any other visible changes). It is switched off by default, so nobody will suffer.
2. I'll explain my particular reason for this feature - it is for non-tech users, who may be not good with English. Developers are anyway using English almost everywhere, but users are different. Users will not run tests themselves, and wiki will be used as a reference for users, when they are uncertain about the behavior of the application. In this case I was going to have a separate TOC for users to be used as a reference.
3. adding a parameter to properties.xml will definitely result in need to check its value in various places of code, which will impact performance. I can see, that properties.xml are already loaded for TOC, but in other cases - properties.xml are not always read when it comes to rendering page name. For example when we simply reference a page in our wiki test - properties.xml is not read. So I'm afraid adding a feature of switching which title to use will definitely result in extra file reads.

AG

six42 wrote:

six42:
Hi @andrey012, sorry for coming back late.

As you mentioned your change is rather cosmetic than functional I am less a fan of it.
For me it would be important to have a consistent view and naming schema on all pages. 

There is already a cosmetic implementation with the -g (graceful) printing of file names in the TOC.
Another cosmetic feature is adding the -h switch in the TOC to show the help text as well. Not sure if you tried this but this would allow to show Großbritannien in your TOC to guide your users. 
If only a cosmetic solution is requested a simpler to maintain approach could be a new TOC switch (-H) to replace the page name in the TOC by the help text. I think the help text is not much used and can be misused as page title without much collateral damage.
Or a better disgracer with more encoding as @amolenaar proposed. (The German wikipedia has a page Großbritannien. Not sure how far you can go with a disgracer and a Chinese title) 

Going away from cosmetic to functional changes a general page switch which shows the help text instead of the page name as page title everywhere could be an approach. 

I agree with you that there is gap in FitNesse due to the limitation of page names.
For class names the user can enter any text and FitNesse constructs a valid class name from the free text.
Unluckily for page names it is the other way round. The user has to give a valid file name and many cosmetic features are added to construct a user understandable page title.
 
I would like to see a functional extension which closes this gap and allows a user to specify any page title and FitNesse takes care to make it a valid file and link name.

amolenaar:
@andrey012 agree with point 1 and 2. Esp. point 2 seems like a good reason to me.
@six42 You were just ahead of me :). I agree with your reasoning.

I still have second thoughts on how the title is retrieved. As far as I know all (German) characters are part of ISO-8859-1, so there should be no problem in putting them into file names nowadays. I think I prefer the ability to use those characters in the page name, instead of via the page name. Even more since this feature is only applicable for the ToC, hence it will be a hassle to search in things like the test history, or page renaming.

I'd like to see how far we can get with page names. Also the "-H" switch sounds like a more viable alternative (it's more explicit), probably with a fallback on the page name if no help text is available.

andrey012:
I can build the -H option as you say. The only thing I'd like to have as a compliment for -H is ability to place help text inside the page without need to copy-paste it.

I will do in another branch and will create separate PR then.

AG

Arjan Molenaar wrote:

six42:
@andrey012 

 !help is the command to show the help text into a page.

And yes if no help text is defined and -H option is given the page name should be displayed in the TOC. if a help text is defined it might also be an option to show the the page name in small letters after the help text. So swapping the text and format compared to -h.  

@amolenaar 
Extending the allowed characters for pages names could work for the file name but I think it then fails with the link names. Link names are encoded by the browser before sending to the server if I understood this correctly. 
So I see in my browser the link http://de.wikipedia.org/wiki/Großbritannien 
But the server sees: http://de.wikipedia.org/wiki/Gro%C3%9Fbritannien 
@andrey012 maybe you want to try it.

andrey012:
Agreed not no move this way, and implement content -H option instead

PR #706, opened 2015-04-04
Problem: when Fitnesse is launched to run a test or a suite (with -c Wik...

andrey012:
Problem: when Fitnesse is launched to run a test or a suite (with -c WikiName?test) - and if WikiName does not exist (e.g. mistyped in build file) - there were NullPointerException, and java returned exit code = 0, which made Maven think, that build have succeeded.

Fix: track the fact, that responder tries to return 404 and throw an exception, then catch exceptions in FitNesseMain.main() and set exit code to 1


amolenaar:
Hi,

I think this is a nice fix. However, can you please add a test case that describes this behaviour?

andrey012:
I'd be glad to, but can you, please give me one hint. 
The problem which I try to solve is following: when I have a Fitnesse test job in build file (e.g. Maven), it looks like launching java with -c pathToSuitePage?suite or something like that. 
This means, that to try behaviour we need to fork new java process. 

I've walked through existing Fitnesse testsuite (controlled by JUnit) - and I couldn't find any similar scenarios there. I mean - in all the tests I have found we do not fork java. 

So, before building such a new thing, I want your advice. 
As I understand, I can: 
1. either add another taget to Ant build.xml, and expect exit code = 1 for launching java -jar fitnesse-standalone.jar -c FitNesse.NonExistentTestSuite?suite
2. either add JUnit test for this case
3. either try to add a page to AcceptanceTests suite (not sure if there is possibility to fork process and expect exit code right now). 

Which way will be better? 

Thank you.

andrey012:
Sorry, I should be more thoughtful. 
I've added test case FitNesseMainTest.runningCommandWithNonExistentAddressResultsInError()
Please take a look.

amolenaar:
That's a nice and concise fix. Thanks!

PR #707, opened 2015-04-05
Issue 703: making prevention of System.exit() calls configurable.

benhamidene:
Introduced new system property prevent.system.exit to control the activation of the SystemExitSecurityManager. Default behaviour is that the SystemExitSecurityManager is activated if the property is not found

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/SystemExitIsPrevented/content.txt
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/SystemExitIsPrevented/content.txt
@@ -1,5 +1,13 @@
+!|script|SystemExitTableConfiguration                 |
+|set system property|prevent.system.exit|to|true|
+
+
 !|SystemExitTable                                        |
 |System exit code|exception message?                     |
 |0               |prevented system exit with exit code 0 |
 |1               |prevented system exit with exit code 1 |
 |42              |prevented system exit with exit code 42|
+
+
+!|script|SystemExitTableConfiguration                |
+|restore original system property|prevent.system.exit  |
--- a/FitNesseRoot/FitNesse/UserGuide/AdministeringFitNesse/CommandLineArguments/content.txt
+++ b/FitNesseRoot/FitNesse/UserGuide/AdministeringFitNesse/CommandLineArguments/content.txt
@@ -1,3 +1,5 @@
+!1 Command line arguments
+
 A subset of configuration options can be provided through the command line. Command line properties take precedence over properties set in the configuration file and system properties (''-Dproperty=value''). 
 
 | ''Command!-&nbsp;-!line!-&nbsp;-!parameter'' | ''Configuration!-&nbsp;-!property'' | ''Default'' | ''Description'' |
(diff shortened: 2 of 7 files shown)


amolenaar:
This looks good. Thanks. Only when I tried an "ant release", the test process was abruptly terminated. It seems like the test case `fitnesse.junit.FitNesseSuiteExampleTest` is failing. Can you have a look please?

benhamidene:
I have fixed the problem. All ant tasks should run now without errors

amolenaar:
Shouldn't the default be to prevent system exit? Since that's the old behaviour.

benhamidene:
well I am a fan of falling fast .. and clean coding:)

And I have changed the behaviour only for test run as Junit-Tests , not from a fitnesse server.

I can not see why we should tolerate/handle calling System.exit in Junit-Tests. If a developer makes such a thing then he should notice it soon and should fix it. If it is part of the test then it could be set for that test case.

I don´t think it is critical if the test run is interrupted when a System.exit() is called in a junit test but it might be when the test is run from the UI of a central FitNesse instance.

bknobloch:
I'm also interested in this pull request as I am experiencing the same performance issue when running tests for i/o intensive applications.

And I second benhamidene's optinion on preferring failing fast and better runtime performance in case of the junit-test runner instead of defaulting to the setting for the wiki-mode.

benhamidene:
Hi amolenaar,

any chances to merge this commit ? not convinced by my answer :) ?

amolenaar:
:) I think you're right. The only doubt I have is that the default behaviours differ when executing a suite from the wiki vs. executing the suite from jUnit.

mgaertne:
I thought that I should provide some background information on why I initially introduced that feature. At a time there were lots of folks asking why FitNesse wasn't providing information on why a test was aborted. Most of these cases ended up as "production code called System.exit() for error handling". Eventually the amount of questions and people blaming FitNesse for the problems was so high, that we incorporated this feature. It seemed to be a good idea at that time.

Now, be aware that changing the default behavior may have people coming back for the same reasons.

Best Markus

benhamidene:
Hi ,
I understand your arguments .But.

* It is a fact the SystemExitSecurityManager is responsible for massive performance leaks. We have a spring based integration test whose execution duration went down from 380 sec (!!!) down to 25 sec after removing the  _SystemExitSecurityManager_ .
* My commit changes the default behavior __exclusively__ for tests run by the __JUnit-Runner__. The behavior for test run from an FitNesse instance stayd _unchanged_.
* I am still convinced that using System.exit() in an application or a test is __not__ a good practice. If it has to be used than it is possible to activate the  _SystemExitSecurityManager_ as I made for the PreventExit Test. I would recommend to consider that as the __exception__ and not the rule.

If you are not convinced and because the performance issue is very critical for us I could turn back the default behavior of the JUnitRunner and make a separate pull request for it that could be voted .

amolenaar:
I agree with your reasoning as far as point 1 and 2 go. Considering point 3, since we're executing tests in a separate process, it's harder to figure out what went wrong.

It thing it would be fine if `FitNesseRunner.shouldPreventSystemExit()` returned `true` by default, since that will be the slower, but more secure option. Power users like yourself can add the annotated when needed. Also, there should be no difference in execution, depending on whether your executing tests from the web interface or from JUnit.

mgaertne:
I understand the reasons why you changed the behavior. I also agree that System.exit() in applications is bad. Nevertheless people are doing terrible things out there, and I wanted to make you aware of the side-effects to your changes.

I certainly agree with Arjan that the behavior should be consistent throughout the application no matter how you run it.

Best Markus

benhamidene:
OK. You win :)

I have updated the commit and added a short description for the property. Ant builds run without error. It should be ready to go now   ;)

Best regards .
Anis

PR #711, opened 2015-04-14
Alias for fixture names

six42:
Write more readable requirements and segregate them from the
implementation with alias names for fixtures.
See FitNesse.UserGuide.WritingAcceptanceTests.SliM.DefineAlias for
samples and details.

Also relax the naming convention for slim tables types names. A
semicolon is now optional. Example: addTableType("dt", DecisionTable.class);


amolenaar:
Indeed, this is a nice way to decouple fixture names from table headers. This can make table headers more meaningful.

PR #715, opened 2015-04-17
Also call testComplete on exception

fhoeben:
Ensure listeners are informed when SlimTestSystem execution of a test threw an exception


amolenaar:
no brainer :)

Issue #714, opened 2015-04-17
JUnitRunNotifierResultsListener should inform RunNotifier of the fact testing was aborted because of an exception 


fhoeben:
If a Slim test run aborts (for instance due to a slim timeout) this is not reported to jUnit by `JUnitRunNotifierResultsListener`. This means that in the jUnit results the fact that pages are skipped because the testSystem was stopped is NOT REPORTED at all.

I think its `testSystemStopped()` implementation should be extended with a call to the notifier reporting a problem while running `mainClass`. 

I'm not sure whether `fireTestAssumptionFailed` or `fireTestFailure` is the most appropriate method of jUnit's `RunNotifier` to call. I'm leaning towards `fireTestAssumptionFailed`.

fhoeben:
I believe a good fix should include `JUnitRunNotifierResultsListener` implementing `TestsRunnerListener`, and tracking how many tests were completed, so that on testSystem abort it can report how many tests were not executed.

fhoeben:
I just looked up [jUnit's documentation on assumptions](http://junit.org/apidocs/org/junit/Assume.html), and now I think we should use `fireTestFailure`

amolenaar:
Closing this issue since #716 has been merged.

PR #716, opened 2015-04-17
Notify jUnit of test system exceptions

fhoeben:
Pass information on test system exceptions to jUnit's RunNotifier, also reporting if tests were skipped because of this, #714.
Furthermore allow subclasses of FitNesseRunner to control which listeners are used in a test run.

### DIFF
--- a/test/fitnesse/junit/FitNesseRunnerExtensionTest.java
+++ b/test/fitnesse/junit/FitNesseRunnerExtensionTest.java
@@ -2,8 +2,16 @@ package fitnesse.junit;
 
 import fitnesse.FitNesseContext;
 import fitnesse.components.PluginsClassLoader;
+import fitnesse.testrunner.MultipleTestsRunner;
+import fitnesse.testrunner.WikiTestPage;
+import fitnesse.testsystems.*;
 import org.junit.runner.RunWith;
+import org.junit.runner.notification.RunNotifier;
 import org.junit.runners.model.InitializationError;
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertNull;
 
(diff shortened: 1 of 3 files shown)


test/fitnesse/junit/FitNesseRunnerExtensionTest.java
```diff
@@ -27,4 +40,35 @@ protected FitNesseContext createContext(Class<?> suiteClass) throws Exception {
       return super.createContext(suiteClass);
     }
   }
+
+  public static class ListenerExtension extends JUnitRunNotifierResultsListener {
```
amolenaar: Why this subclass?

amolenaar:
Looks good. How can I test it?

test/fitnesse/junit/FitNesseRunnerExtensionTest.java
fhoeben: I actually used it to test (with debugger) that my behavior was called. I apologize for such a rude form of testing. I could have just deleted it.

fhoeben:
I was wondering about that myself. In my real-world environment I had a test that caused Slim to be aborted. But I don't think I can introduce that in the unit-test set of FitNesse, without breaking its unit test suite.

Maybe we can use an in-memory wiki-page, combined with a mock notifier. But I saw no way to easily wire that up, not in any way that was proportional in complexity to the change itself. Do you see a way to do that? 

I did not think just calling the method directly (i.e. NOT in context of a run by MultipleTestsRunner) and seeing that the notifier was called (i.e. a true unit-test) was a test that added much value. That could easily be done, of course.

PR #718, opened 2015-04-18
content -H option -- shows subWiki page's help text as title in TOC and ...

andrey012:
...shows subWiki page's name with small letters (i.e. swaps subWiki page's name and help text in TOC)

Main reason - to let make titles in TOC contain pieces of text, that are not allowed in page name (i.e. non-ASCII chars).

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteWidgetTests/ContentsUsage/content.txt
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteWidgetTests/ContentsUsage/content.txt
@@ -6,6 +6,8 @@ Use !contents to list all or some of the child pages of the current page along w
     -f   ...show suite filters--define FILTER_TOC {true} for global;
     -g   ...show graceful names in the list--define REGRACE_TOC {true} for global;
     -h   ...show help property text--define HELP_TOC {true} for global;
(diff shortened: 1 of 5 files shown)


six42:
Wow, this PR was fast.

andrey012:
This one was much easier then fetching headers from pages.
Actually -- thank you for -H idea.

AG

six42 wrote:

amolenaar:
That's fast indeed.

I was checking the User guide full reference page with the new option and I love it:

screen shot 2015-04-18 at 22 37 17

I think the ': page name' can go, though, since it does not add useful information, or maybe put it in parenthesis,
Now I got confused with the normal help text:

screen shot 2015-04-18 at 22 37 36

andrey012:
As for me - I'd leave the page name for those who maintain Wiki - it can be quite useful when moving and referencing pages, managing commits in Git-controlled wiki and similar things, especially if names will not be disgraced.
My initial idea was of giving better understanding for non-tech users, which, I think, is completely satisfied with this option. Page name, written with small letters, should not confuse them.
I can remove page name, or make them optional (in this case - please suggest a proper option to use).

AG


Arjan Molenaar wrote:

amolenaar:
I think there's little need in showing the page name. Wiki maintainers can use the URL directly.

andrey012:
done.

AG
Arjan Molenaar wrote:

PR #721, opened 2015-04-21
Fixes #719 by adding System.exit(0) call to SlimService.java

### LINKED ISSUE #719: SlimService is missing System.exit(0) which prevents the process from being shut down correctly, if any non-damon threads have been spawned as part of test fixture.
When running integration tests via FitNesseMain, it may occur, that non-daemon threads are spawned as part of the SlimService process, e.g. RMI registry threads. This prevents the SlimService process from being shut down properly. While SlimService.main() contains calls to System.exit() in case of an error, a System.exit(0) is missing for successful fixture execution.

maverick1601:
fixes #719


amolenaar:
Looks okay, given the System.exit() already in the catch clause.

PR #724, opened 2015-04-24
Add interceptor support to Slim's MethodExecutor

fhoeben:
Adds the option for a Java Slim fixture to 'intercept' all calls made to it by Slim, by implementing `fitnesse.slim.fixtureInteraction.InteractionAwareFixture`.

This allows a fixture to implement some generic behavior without having to add it to all of its methods invoked by Slim.

Examples of such behavior:
* exception handling allowing exceptions from any of its methods to be converted to SlimExceptions (without having to implement that in each method invoked), or 
* waiting for the system under test to be in a stable state before invoking the next method (again without having to add that waiting to each method).

Personally I have 2 specific wishes I hope to implement based on this. I'm hoping to extend my Selenium based fixture so that
1. it will include a screenshot in the FitNesse generated HTML each time a Selenium exception is thrown, and
2. implement waiting for Angular DOM manipulation to complete before having my fixture's method interact with the browser.


six42:
Hi Fried,

sounds like a nice feature which helps to reduce the amout of glue code 
to be written.
Could you add a FitNesse page which explains and demonstrates the new 
feature with an example.
This will assure that many people can benefit from your new feature.

An example which adds a screenshot or an image to the test results would 
be really cool.
But not sure how many dependencies this adds.

fhoeben:
I'll try to get around to writing the wiki page. 

The screenshot functionality I use, have in mind, adds a Selenium dependency which I fear will be a bit too much to have in FitNesse itself. Adding screenshots is quite a powerful feature in my experience. I already have it in my BrowserTest fixture (invoked as `|show|take screenshot|myName|`) and even developed a custom Slim table, 'storyboard' (https://github.com/fhoeben/hsac-fitnesse-plugin/wiki/3.-Storyboard-Example), that does it for each row.

I have exception handling that takes a screenshot currently implemented with a custom 'fixture interaction' (https://github.com/fhoeben/hsac-fitnesse-fixtures/blob/master/src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java), but this change will make it easier to implement and will remove the need for this customer fixture interaction (and the need to configure it in the wiki using a slim.flags variable).

PR #725, opened 2015-04-26
StackTraceEnricherTest.shouldParseDirectories is red on Windows

2xmax:
Whoa! I checked out the repo on Windows and found that the unit tests had been broken. In a word, the assertion tried to find a substring **C**:\fitnesse in ..some prefix..**c**:\fitnesse...some suffix.... Then I fixed and calibrated it both on Windows 8.1 and Ubuntu 14 and ensured that it works correctly.

I hope it is useful for you. Any comments are welcome!


amolenaar:
Thanks. The readability of the test improved a lot as well.

PR #728, opened 2015-05-01
fixes #698: tag filtering is case insensitive

### LINKED ISSUE #698: SuiteFilters are not consistantly case (in)sensitive
When figuring out why a suite of us was not executing, I found out that the SuiteFilter behaves differntly on terms of case sensitivity when using the runTestsMatchingAnyTag (case insensitive) and runTestsMatchingAllTags (case sensitive) filters.

What is the correct & expected way of handling them?

For details see lines [163](https://github.com/unclebob/fitnesse/blob/master/src/fitnesse/testrunner/SuiteFilter.java#L163) and [173](https://github.com/unclebob/fitnesse/blob/master/src/fitnesse/testrunner/SuiteFilter.java#L173)

salandur:
in the original code the tag filter was caseinsentive and with the introduction of the exclude filter this was implemented wrongly

Change-Id: I7ec37f74e5f3015a647651d4f61ae8273d2bf2a7

### DIFF
--- a/src/fitnesse/testrunner/SuiteFilter.java
+++ b/src/fitnesse/testrunner/SuiteFilter.java
@@ -16,6 +16,9 @@ import fitnesse.wiki.WikiPage;
 import fitnesse.wiki.WikiPagePath;
 import org.apache.commons.lang.StringUtils;
 
+import static org.apache.commons.lang.StringUtils.equalsIgnoreCase;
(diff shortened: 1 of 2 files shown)


amolenaar:
I think this behaviour requires a test :)

salandur:
Some unit tests are added

Issue #631, opened 2015-02-13
Install CodeMirror plugin for page editing


DmitriyKartashov:
http://codemirror.net/ - jQuery plugin for highlighting text in textarea.
It provides a lot of features such as code folding, bracked matching , multiline editing and so on.
Example of the plugin work can be found here: http://runnable.com/UmGTLn0iIYljAAEu/codemirror-example-change-theme

amolenaar:
You want to use this as a replacement for the plain text area? Sounds like a plan, but a custom syntax is needed I suppose(?)

DmitriyKartashov:
Yes, replacement for the plain text area.
You're right. The custom syntax is a part of installation process. Actually the main one.
If you don't mind later I'll discribe which features of FitNesse will be supported in the text editor. I'm only started to work with it.
Btw.
This IDEA plugin will be used as an example of code highlighting: https://plugins.jetbrains.com/plugin/1143
And code folding will be provided for collapsible blocks in the first version as well.

amolenaar:
merged

PR #730, opened 2015-05-01
Install CodeMirror plugin for page editing #631

DmitriyKartashov:
Implementation of task #631

Code ready for review.
Tested in IE 11, Google Chrome 41.0.2272.118, Firefox 37.0.2, Opera 29.0.


amolenaar:
Nice patch! CodeMirror works pretty well. The colouring makes the wiki text more readable.

I have two suggestions:

1. I think `src/fitnesse/junit/JavaFormatter.java` doesn't need changing, since it only shows test results.
2. The CodeMirror editor could use a border, so it stands out more.

DmitriyKartashov:
1. Fixed.
2. Partially fixed. I couldn't set up blue border with editor is focused. Is it critical?

Don't I need to do some work regarding licenses?

amolenaar:
Thanks. I don't think we need to do anything regarding the license. It would be nice though to have a page explaining the components used along with their licenses.

DmitriyKartashov:
Page was added. Please, review.

amolenaar:
Nice, I'll find some time to merge this.

PR #732, opened 2015-05-02
Make SlimExceptionResult recognize message with a newline

### LINKED ISSUE #731: SlimExceptionResult should recognize message with a newline
SlimExceptionResult parses exception text to determine whether it has a 'message'. At the moment it does not see exception messages including a newline as having a message. I believe it should.

At the moment the unit test below fails

```java
  @Test
  public void messageWithNewlineFound() {
    SlimExceptionResult result = new SlimExceptionResult("key", "message:<<Bad things\nhappened>>");
    assertTrue(result.hasMessage());
    assertEquals("Bad things\nhappened", result.getMessage());
  }
```

fhoeben:
Fixes #731

Besides allowing newlines in the message it also ensures that the pattern matches the whole of the exception text (i.e. the exception text should not have more content)


fhoeben:
The extra 'fix' I thought up to not allow pre- and postfix is a bad idea, so I'm removing that.

PR #739, opened 2015-05-18
Improved slim hash handling

### LINKED ISSUE #738: Improve display and handling of Maps in Slim (for Java)
Currently Slim supports supports sending Maps to fixtures based on HashTables (as described in http://fitnesse.org/FitNesse.FullReferenceGuide.UserGuide.WritingAcceptanceTests.SliM.DataTypes), and formatting of Map results by fixtures as HashTables. 

This is great but quite limited at the moment.
* The support by Slim is not mentioned in the HashTable's documentation (http://fitnesse.org/FitNesse.FullReferenceGuide.UserGuide.FitNesseWiki.MarkupLanguageReference.MarkupHashTable), this only mentions `DoFixture`
* formatting of nested Maps in fixture results is not supported (i.e. a nested map is rendered as `{a=1, b=2}` instead of as a nested table, which is supported by the HashTable widget)
* the formatting is only applied if the declared return type of the method is `Map`, it is not based the actual returned value's type (this means methods declaring subclasses as return type do not get the feature, and neither do methods that declare `Object` when they actually return a `Map` implementation).

fhoeben:
Fixes #738 

I'm not quite sure of the the change to `MethodExecutionResult.returnValue() `, it was needed to keep the unit tests green, BUT I don't really grasp why Lists were treated this special before and whether my change capture the intent of the previous code.

What I believe happens with the current (and previous) code is that all output objects are sent back to FitNesse using their converter's toString() result, except Lists (which is always sent using its own toString() result). Why prohibit defining a custom toString() for lists using a converter?
As in-fact the conversion of Lists can not be controlled, shouldn't that be made explicit somewhere in the documentation (with reason)?

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/TestCustomConverter/content.txt
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/TestCustomConverter/content.txt
@@ -1,0 +1,27 @@
+We check whether a custom converter for lists can be used to change how lists are returned to the wiki.
+
+Whether a converter is applied to a method's result depends on its declared return type:
+
+ * List (exactly, no sub-interface or class implementing it): converter is '''not applied''' (this is needed for some Slim tables, such as !-QueryTable-!).
+ * All return types: converter is applied, based on the actual class of the object returned, or on its declared type when !style_code[null] is returned 
+
+
+|import            |
+|fitnesse.slim.test|
+
+|script   |!-TestSlimWithConverter-!|
+|check    |get object     |[a, b, c]|
+|check    |get list       |[a, b, c]|
+|check    |get array list |[a, b, c]|
+|same list|[a, b, c]                |
+|set list converter                 |
+|check    |get object     |{a, b, c}|
+|check    |get list       |[a, b, c]|
+|check    |get array list |{a, b, c}|
+|same list|{a, b, c}                |
+|same list|[a, b, c]                |
+|remove list converter              |
+|check    |get object     |[a, b, c]|
+|check    |get list       |[a, b, c]|
+|check    |get array list |[a, b, c]|
+|same list|[a, b, c]                |
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/TestCustomConverter/properties.xml
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/TestCustomConverter/properties.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<properties>
+<Edit>true</Edit>
+<Files>true</Files>
+<Properties>true</Properties>
+<RecentChanges>true</RecentChanges>
+<Refactor>true</Refactor>
+<Search>true</Search>
+<Test>true</Test>
+<Versions>true</Versions>
+<WhereUsed>true</WhereUsed>
+</properties>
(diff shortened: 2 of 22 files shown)


fhoeben:
Not only MethodExecutor's result creation and MapConverter's `toString()` now uses the actual type of its elements, but ArrayConverter's and GenericCollectionConverter's `toString()` as well.

amolenaar:
> What I believe happens with the current (and previous) code is that all output objects are sent back to FitNesse using their converter's toString() result, except Lists (which is always sent using its own toString() result). Why prohibit defining a custom toString() for lists using a converter?

I think this has to do with the way lists are treated special on both sides (since statements are all converted as lists). It would make sense to have a custom converter for lists that are output of fixture invocations. That way you can properly format the list (e.g. strip the square braces, or quote the elements).

> As in-fact the conversion of Lists can not be controlled, shouldn't that be made explicit somewhere in the documentation (with reason)?

Is it not possible to make a ListConverter at all?

fhoeben:
You can make a Converter for lists, but it will not be used for the results of methods returning a List...

With my new code it WILL be used for lists nested inside arrays, collections or maps. But still not for the top-level result.

fhoeben:
Sample of current behavior using a new fixture class and wiki page (which is red because list is not rendered using custom formatter):

## Wiki
```
We check whether a custom converter for lists can be used to 
change how lists are returned to the wiki.
But the call to send the list in the alternative approach works.

|import            |
|fitnesse.slim.test|

|script   |!-TestSlimWithConverter-!|
|check    |get list    |[a, b, c]   |
|same list|[a, b, c]                |
|set list converter                 |
|check    |get list    |{a, b, c}   |
|same list|{a, b, c}                |
|remove list converter              |
|check    |get list    |[a, b, c]   |
|same list|[a, b, c]                |
```

## Fixture

```java
package fitnesse.slim.test;

import fitnesse.slim.Converter;
import fitnesse.slim.converters.ConverterRegistry;
import fitnesse.slim.converters.DefaultConverter;
import fitnesse.slim.converters.GenericCollectionConverter;

import java.util.Arrays;
import java.util.List;

public class TestSlimWithConverter {
    private Converter<List> standardConverter = null;

    public void setListConverter() {
        standardConverter = ConverterRegistry.getConverterForClass(List.class);
        ConverterRegistry.addConverter(List.class, new ListConverter());
    }

    public void removeListConverter() {
        ConverterRegistry.addConverter(List.class, standardConverter);
    }

    public List getList() {
        return Arrays.asList("a", "b", "c");
    }

    public boolean sameList(List otherList) {
        return getList().equals(otherList);
    }

    private class ListConverter implements Converter<List> {
        private GenericCollectionConverter<Object, List<Object>> converter;
        public ListConverter() {
            converter = new GenericCollectionConverter<Object, List<Object>>(List.class, new DefaultConverter());
        }

        @Override
        public String toString(List o) {
            String s = converter.toString(o);
            return s.replace("[", "{").replace("]", "}");
        }

        @Override
        public List fromString(String arg) {
            String replace = arg.replace("{", "[").replace("}", "]");
            return converter.fromString(replace);
        }
    }
}
```

fhoeben:
Updating `MethodExecutionResult`'s returnValue method to the code below makes the wiki page above green (but breaks unit tests in `HashWidgetConversionTest`):

```java
  public Object returnValue() {
      return toString();
  }
```

fhoeben:
It look like (at least the QueryTable) is written to expect to get it lists as lists (and not as a single string to be parsed). And making it parse the string is hard when it is not aware of the converter used (which I believe it can't be aware of since that converter is in the Slim server process and not in the Wiki process where the QueryTable class lives).

So it looks like converters (at least for lists) are actually in the wrong place at the moment (the Slim server process) they should be in the wiki process.

fhoeben:
My previous code did change the handling of List results: it never converted a result to string if the result was an instance of List. The old behavior only did not convert if the method was defined to return a List (and not a subinterface or subclass) and never if the method was defined to return Object or a Collection.

With my latest commit the old behavior is back: results of methods declared to return a List are not converted, all other results are.

fhoeben:
What I did not alter (yet) in this pull requests is how arrays and collections are handled in general. For these the behavior is still what it was: using the `GenericCollectionConverter` and `GenericArrayConverter`. 

Should that also be made pluggable? I was thinking we could add an extra method to register an ArrayConverter  (based on the element type) and one to register an CollectionConverter (based on the Collection and element type). Would that add value?

amolenaar:
I was (finally) about to merge this PR, but ran into a test failure:

```
    [junit] Testcase: FitNesse.SuiteAcceptanceTests.SuiteSlimTests.TestCustomConverter took 0.029 sec
    [junit] 	FAILED
    [junit] [[a, b, c]] expected [{a, b, c}]
```

Can you reproduce this?

fhoeben:
Which suite/target were you running?
The fact that the ConverterRegistry is a (static) singleton has caused me problems while testing before

fhoeben:
When I ran the Ant unit_tests target today I got an failure also (another than you got). My latest commit addressed that by ensuring the ConverterRegistry was in its default state before and after.

I don't really see how the page you got an error from would fail (it already tries to clean up after itself).

For me that page is green when I run it via the wiki, and `ant unit_tests` also encounters no failures or errors. `ant` also completes without errors.

Do you still have the issue?

Issue #630, opened 2015-02-13
Install DataTables jQuery plugin for tables


DmitriyKartashov:
http://www.datatables.net/ - poverfull jQuery plug-in for decorating html tables. The plugin allows to perform searching inside the tables, filtering the data and other useful features.
I suggest to install it by default. It will simplify work with the tables.

woodybrood:
What does this actually get a FitNesse user? It's some pretty powerful
table functionality, but what does the user actually do with this in a
typical setup?

Keep in mind that "Big Tables" are often unhealthy in FitNesse tests, as
they are slow and consume a lot of memory when processed/executed.

So, how does it simplify work with the tables?

Daniel A. Woodward
-------------------------
FitNesse using Slim and Java
@woodybrood
http://whotestedthis.co

DmitriyKartashov:
The user gets sorting, better style, filtering and searching. I suggest this list as a default set of features.

In fact it really helps already with 8+ rows tables.
I'm a FitNesse user too. Sure, most of our tables have less than 6 rows of data. Here the features might be superfluous. I agree. But we have to work with 10, 16 and even 50 (in extreme cases up to 100) rows table. I'm not sure that they aren't "Big Tables" but I really miss filtering when I'm btw checking test cases.

I'd also suggest to disabled the features for tables with less than N rows. N = 8 by default. Sorry that I didn't mention it in the origin issue.

woodybrood:
Can it be done in a way that would be compatible with the coloring changes
that folks have implemented?

How would it work with ScriptTable or TableTable?

DmitriyKartashov:
>>Can it be done in a way that would be compatible with the coloring changes
that folks have implemented?
What do you mean? If you are talkng about test results highlighting - answer is yes. Does FitNesse have some other coloring?

As far as I remember tables have no html classes.
I hope it will be possible to define separate class for every type of table. In that case user will be able to configure witch types of tables will have DataTables support.
By default it can be On for DT, DDT and Off for other.

amolenaar:
I don't feel comfortable with this. It would make for a nice plugin. I think it's better to make the UI more pluggable.

PR #742, opened 2015-05-20
Install DataTables plugin #630

DmitriyKartashov:
Implementation of task #630

Sorry for large pull request.
Please, pay attention to changes made PropertyBasedPluginFeatureFactory. I'm not sure that I've done them in proper way.

I've run tests locally (ant.all task). I've got 48 test fails.
The messages were like "Method createPageWithContent[2] not found in fitnesse.fixtures.HashFixture." With different method and class names.

Code ready for review.
Tested in IE 11, Google Chrome 41.0.2272.118, Firefox 37.0.2, Opera 29.0.


amolenaar:
I'm not sure about this PR. What problem is it solving? I mean: it's a big chunk of javascript...

DmitriyKartashov:
I'd say it helps to create and maintain big test tables.
For example, with this tool it's easy to check is there a duplication of test cases or not.
The user can just enter the test case values in the "filter by" fields and see is such a case already created (and how many times).

Big tables doesn't mean 100+ rows.
For DDT if we have four columns as input data even if they are boolean it will require 16 test cases to cover all the flows. And what will be if they are, say, integers?...

I showed that tool localy to my team and all of them (including devs and BAs) are liked it.
It should mean something...

amolenaar:
I still have my doubts. I ran into some issue while testing this branch (e.g. pages like `FitNesse.SuiteAcceptanceTests.SuiteResponderTests.SuiteRenameResponder.TestRenameChildPageWithSiblingReference` appeared empty).

This might make for a nice plug-in, I'd say. For that we have to make some adjustments to FitNesse, but that would open the possibility to create plug-ins that can add behaviour to the front-end, which would be awesome.

amolenaar:
I don't feel comfortable with this. It would make for a nice plugin. I think it's better to make the UI more pluggable.

PR #745, opened 2015-05-24
Unit test classes should not write files at parent directory

linquize:
Unit test classes should not write files at parent directory

e.g. repository is at `/home/user/code/fitnesse`
but unit test created a folder `/home/user/code/target`
It should be  `/home/user/code/fitnesse/target`


linquize:
ping

PR #755, opened 2015-06-20
SlimTestContext modified to provide current TestPage

andreptb:
Following guidelines for contributing, I just posted a [question on stackoverflow](http://stackoverflow.com/questions/30959027/fitnesseroot-dir-access-from-a-custom-slim-table).

This pull request is my own approach to resolve that need. By adding the current TestPage instance to SlimTestContext, from my custom SlimTable I'm able to fetch variables from SystemVariableSource, such as *FITNESSE_ROOTPATH* and *FITNESSE_ROOT_DIR*.

### DIFF
--- a/src/fitnesse/testsystems/slim/SlimTestContextImpl.java
+++ b/src/fitnesse/testsystems/slim/SlimTestContextImpl.java
@@ -7,6 +7,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import fitnesse.testsystems.ExecutionResult;
+import fitnesse.testsystems.TestPage;
 import fitnesse.testsystems.TestSummary;
 import fitnesse.testsystems.slim.tables.ScenarioTable;
 
@@ -14,6 +15,15 @@ public class SlimTestContextImpl implements SlimTestContext {
   private final Map<String, String> symbols = new HashMap<String, String>();
   private final Map<String, ScenarioTable> scenarios = new HashMap<String, ScenarioTable>();
   private final TestSummary testSummary = new TestSummary();
+  private TestPage pageToTest;
+
+  public SlimTestContextImpl(TestPage pageToTest) {
+    this.pageToTest = pageToTest;
+  }
+  
+  public SlimTestContextImpl() {
+    super();
+  }
 
   public String getSymbol(String symbolName) {
     return symbols.get(symbolName);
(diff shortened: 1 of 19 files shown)


src/fitnesse/testsystems/slim/SlimTestContextImpl.java
```diff
@@ -14,6 +15,15 @@
...
   private final TestSummary testSummary = new TestSummary();
+  private TestPage pageToTest;
+
+  public SlimTestContextImpl(TestPage pageToTest) {
+    this.pageToTest = pageToTest;
+  }
+  
+  public SlimTestContextImpl() {
+    super();
+  }
```
amolenaar: I think the default constructor can go. It should no longer be needed, right?

cc: @fhoeben

andreptb: No, it's no longer needed. I tried to remove, but then I got a few compilation errors on unit tests that instantiated this class to test something else. Since TestPage would serve as optional "context" data I figured to just leave the default constructor and unit tests as it is.

Want me to pull request again removing it? I'll propably have to mock a TestPage instance for unit tests that are instantiating SlimTestContextImpl.

fhoeben: I think deleting the default constructor and just have the unit tests pass in `null` is better (but more work, maybe your IDE can help you there).
Furthermore I would suggest making the field `pageToTest` final.

andreptb:
Ok @amolenaar removed default constructor and made pageToTest final. About adapting the unit tests some tests creates an instance variable referencing root page, in those I used:

```java
SlimTestContextImpl testContext = new SlimTestContextImpl(new WikiTestPage(root));
```
Other tests that didn't create this reference, so I used:

```java
SlimTestContextImpl testContext = new SlimTestContextImpl(new WikiTestPage(new WikiPageDummy()));
```
Just ran the tests, everything seems to be OK.

fhoeben:
The fix you made is most welcome, and I can use it in my custom Slim tables when merged, BUT is not a complete solution to https://github.com/fhoeben/hsac-fitnesse-fixtures/issues/15.
I think an extension to the Slim protocol is required to correctly address the issue of fixtures creating and/or accessing files from the wiki's files section. Because the fixtures might run on another machine than the wiki server. (I gave a very brief description of what could be done to allow files to be created and referenced by fixtures in https://github.com/fhoeben/hsac-fitnesse-fixtures/issues/15#issuecomment-116254628)

andreptb:
@fhoeben,

You're right. I think we can use this issue to address local fixtures and maybe create another one proposing this extension?

Issue #758, opened 2015-06-26
Supporting labeled constructor argumemnts


benhamidene:
Hi,

our business experts using fitnesse would like to be able to describe the constructor arguments of fixtures

So for instance instead of having such a table

|script|Fixture with parametrized constructor | argument 1 | argument 2 |
|...|

They would like to write s.th like:

|script|Fixture with parametrized constructor | #first argument | argument 1 | #second argument | argument 2 |
|....|

I will send a pull request for that.

PR #759, opened 2015-06-26
#758: Supporting labeled constructor arguments.

benhamidene:
see issue 758

https://github.com/unclebob/fitnesse/issues/758

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/TestWeSupportLabeldConstructorArguments/content.txt
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/TestWeSupportLabeldConstructorArguments/content.txt
@@ -1,0 +1,36 @@
+| import |
+| fitnesse.slim.test |
+
+
+
+!2 Constructor without arguments 
+
+|script|Fixture with parametrized constructor |
+|check|saved porperties are | null |
+
+
+!2 Constructor with one argument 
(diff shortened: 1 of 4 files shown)


benhamidene:
Hi,
I agree with your arguments. But I am still convinced that we need some solution. Sometimes we have fixtures with multiples arguments in the constructor. That is not really leasable / understandable.

@sergebug I did not get your second point. Do you have an example how to use the second type of constructors?
That could be sufficient ..

sergebug:
My point is FitNesse already has more than one option to get what you need. 
Regrading Scenarios please look here for details: 
http://www.fitnesse.org/FitNesse.FullReferenceGuide.UserGuide.WritingAcceptanceTests.SliM.ScenarioTable 

To give you an example you could place this in library:
```
!|Scenario|Constructor with First Argument|arg1|Second Argument|arg2|
|Constructor;|@arg1|@arg2|
```
Then inside you test you can write script like this
```
!|Script|
|Constructor with First Argument|value1|Second Argument|value2|
```
During run-time it will be transformed and executed as scenario specified above.

mgaertne:
You could also provide a constructor with four parameters instead of two, and ignoring the two with the documentation conpletely in your custom fixture.

Best Markus

-- 
Dipl.-Inform. Markus Gärtner
Test-Driven Development

http://www.shino.de/blog
http://www.mgaertne.de
http://www.it-agile.de
Twitter: @mgaertne

Issue #762, opened 2015-07-02
jUnit run using FitNesseRunner fails test if multiple test systems are used


fhoeben:
When a FitNesse Suite containing tests for multiple test systems (e.g. Slim and Fit) is run as part of a jUnit test run (using `@RunWith(FitNesseRunner.class)`) the test throws an error, claiming not all tests are executed, after the tests for the first test system are completed.

At this point in time this is correct, as the tests by the other test system(s) have not been executed yet, but no reason to stop the test execution. The check whether all tests have been executed is only relevant after all test system's tests have been executed.

The problem is in `JUnitRunNotifierResultsListener.notifyOfTestSystemException()`

PR #763, opened 2015-07-03
Allow multiple test systems during jUnit run

fhoeben:
Please consider this for #762


amolenaar:
Thanks! I suppose you're the first one trying this :p

PR #775, opened 2015-07-27
Various issues

hansjoachim:
Fixes a couple of issues. See individual commit messages for details. `ant` ran succesfully.

I have a small question at the end, since I noticed that BaseFormatter and PluginFeatureFactoryBase both have protected final Loggers. Is this in order to have the subclass hierarchy use the same logger? I'm more used to each class having it's own specific logger, so I'm curious about this approach. :)


hansjoachim:
Thanks for merging. Any thoughts on the loggers?

PR #779, opened 2015-08-02
Fix bugs in Symbol text replacements

six42:
This PR makes usage of Symbols in TableTables very easy.

It fixes the bugs descibed in
http://fitnesse.org/FitNesse.SuiteAcceptanceTests.SuiteSlimTests.TableTableSuite.TestTableTableImplementingStatementExecutorConsumer.TestKnownLimitations

For more complex  symbol requirements the fixture can now handle the
full symbol management itself.

@jplambert this will break your tests pages you have to add another statement to make your tests pages work. Unluckily it is not possible to fix the bugs without breaking your code as you benefit from the bugs. See the new version of TestTableTableImplementingStatementExecutorConsumer for details.

The bug was mainly that the symbol pattern also matches the symbol assignment pattern.
So if $A was 44 and a cell contains the assignment $A= it was replaced by 44=. This was nothing the fixture could understand. 
Only if the symbol was new no replacment happenend and the assignment was visible in the fixture.


jplambert:
Hi @six42, and thanks,

That's the kind of thing I wanted to do but was put off by the fact that it would break compatibility with existing implementations of TableTable.

I see that you suggest to support both ways by using a slim symbol to configure the expected behavior. That's a nice idea, though I find it rather tricky: it is not self-explanatory, setting the value is not very test-friendly, and mixing both kind of TableTable in the same page could lead to something ugly :-) Still that should work, it is an interesting proposal.

I would suggest to have the old behavior as the default behavior. That way, no change would be needed for existing TableTable implementations. Also the cases that get fixed are rather uncommon. Most people won't care.

Finally, I like the way you factored symbol stuff in SlimSymbol. This is a welcome clean-up.

What do the FitNesse maintainers think?

six42:
Hi Jean Piere,

thanks for taking the time to look at this PR.
It was more work than I expected to do the re factoring and in the end I rolled back some standardizations to keep backward compatibility.
Good to hear that you like it.

I made some updates to simplify the usage and increase compatibility.
I wanted a solution where the fixture code steers the behaviour and not the test page.
Now I found one:

I moved the assignment of the switch in the fixture code itself.
Two additional lines do the magic. 
No changes required to test pages.

How many fixtures you have which implement StatementExecutorConsumer? They might still need a change.

test/fitnesse/testsystems/slim/tables/SlimTableTest.java
```diff
@@ -113,8 +113,8 @@ public void replaceSymbolsFullExpansion_ShouldMatchFullSymbolName() throws Excep
   public void replaceSymbols_ShouldReplaceConcutenatedSymbols() throws Exception {
     SlimTable table = new MockTable();
     table.setSymbol("x", "1");
-    table.setSymbol("y", "1");
-    assertEquals("this is $x->[1]1 and $y->[1]1", table.replaceSymbolsWithFullExpansion("this is $x1 and $y1"));
+    table.setSymbol("y", "a");
+    assertEquals("this is $x1->[11] and $yb->[ab]", table.replaceSymbolsWithFullExpansion("this is $x1 and $yb"));
```
amolenaar: Is this okay? Shouldn't it say "$yb->[a]b"? Or am I missing something?

six42: Please look at issue #780.
This doesn't works. How we display it doesn't matters.
to keep the code simple it would be best to remove it.

jplambert:
@six42, that's another nice trick but still... I find it, well... I find it tricky! Assigning a symbol in the context from the TableTable implementation, sure it will work but it heavily smells like it never was designed to be used that way.

Now of course the problem here is some kind of fundamental flaw in FitNesse: symbols and context handling in SLIM is sort of a mess -- it works pretty well with FIT but that's because it relies on static methods which on the other hand is BAD (have a look at the unit tests). So in my opinion "the right way" to do it would imply to rewrite quite a few things, and certainly to break compatibility with existing TableTable implementations.

So from this point of view, since I don't see it happening, your solution is not that bad. However it should be proprely documented. Basically your comments in setStatementExecutor and doTable of TableTableIncFirstCol explain it all and should be put in some place that would be easy to find. In the FitNesse doc of TableTable for instance! However I can understand that you wish to have some positive feedback on the PR before spending a non-negligible amount of time writing a top-notch doc ;-)

(Don't mind my harsh words on FitNesse, I love this tool which works great, but still there are many many things that seems badly designed or that don't fit well together and that I'd make differently if I were to write FitNesse again from scratch... Which I don't think will happen!)


As for how many fixtures I have which implement StatementExecutorConsumer, I'd say none personnally :-) But I spent some time adding support for the RestFixture: https://github.com/smartrics/RestFixture

So far that's the only implementor of StatementExecutorConsumer I know of, but that's the whole point of compatibility: we don't know how many people are using it in the wild... On the plus side, I would call implementing StatementExecutorConsumer and un-documented feature of FitNesse, and most people aren't using TableTable at all.

six42:
Hi @jplambert,

question regarding your RestFixture. Do you use the Slim Symbol Assigment Pattern "\\A\\s*\\$([A-Za-z]\\w*)\\s*=\\s*\\Z" in your fixture to assign symbols or is this only done with the "let" command of the RestFixture. And same question for getting symbol values. You use the slim symbol pattern "\\$([A-Za-z]\\w*)" or your pattern "\\%([a-zA-Z0-9_]+)\\%".

If your fixture doesn't uses the standard slim patterns than you will have no issues with this pull request and need not to change anything. :+1: 

In this case I could remove the 'tricky' switch to support the advanced symbol handling. I just added it to don't lock you out.  
FitNesse would simply not support this and if somebody needs it the way to do it is use other patterns as you did.

There have been no comments from anybody else which likely confirms that not many people use table table with symbols.

six42:
Hi @amolenaar ,

Thanks for marking this for the next release. Any date already?
#791 breaks this PR now :(
will you fix it? Any other change required before the merge?

hcf:
HI @six42 and @amolenaar. You are indeed correct in that my PR is not correctly working. I have added failing tests now, and should be easy enough to fix in my PR, or should I create a new PR or in this one?

amolenaar:
Just add a fix and push it to your branch is enough. You might want to rebase/squash, to ensure all commits result in working code. That makes it easier to look for problems (via `git bisect`)

hcf:
#796 created @amolenaar

jplambert:
Hi @six42, sorry for answering so late,

I believe the most important is to notify @smartrics as he is the owner of RestFixture. That way he'll be able to cope with any change introduced in FitNesse.

six42:
Done. This PR contains also all changes from #791 #796.

hcf:
Hi @six42  and @amolenaar. I closed my pull request.

@six42 could you maybe add some of my tests to the SlimTableTest?

```
@Test
public void replaceSymbolsShouldMatchSingalUTF8SymbolName() throws Exception {
    SlimTable table = new MockTable();
    table.setSymbol("Æ", "ae");
    table.setSymbol("Ø", "oe");
    String actual = table.replaceSymbols("$Æ $Ø");
    assertEquals("ae oe", actual);
}

@Test
public void replaceSymbolsShouldMatchUTF8SymbolName() throws Exception {
    SlimTable table = new MockTable();
    table.setSymbol("Ære", "honer");
    table.setSymbol("Høre", "hear");
    String actual = table.replaceSymbols("$Ære $Høre");
    assertEquals("honer hear", actual);
}

@Test
public void replaceSymbolsFullExpansion_ShouldReplaceUTF8Symbols() throws Exception {
    SlimTable table = new MockTable();
    table.setSymbol("æ", "1");
    table.setSymbol("ø", "1");
    assertEquals("this is $æ->[1]1 and $ø->[1]1", table.replaceSymbolsWithFullExpansion("this is $æ1 and $ø1"));
}
```

amolenaar:
@hcf @six42 Nice to see such collaboration. I expect the tests could be added and then all can be merged?

FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/SlimSymbols/NamingConvention/content.txt
```diff
@@ -0,0 +1,31 @@
...
+!3 Slim symbols names are case sesitive
+
+!|test slim          |
+|string|getStringArg?|
+|lower |$x=          |
+|upper |$X=          |
+|$x    |lower        |
+|$X    |upper        |
+
+!3 Slim symbols names also support UTF8 names
```
six42: Hi @hcf and @amolenaar

I added this test page with this UTF8 test: FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/SlimSymbols/NamingConvention

Think this should cover the required testing 
and in addition provides the documentation to the community that this new feature has been added.

I am a big fan of "Eat your own dog food" and prefer adding wiki tests then Junit tests.
And I think this is a nice example that wiki tests can be more efficient than Junit as it takes less lines to express the feature in a wiki page than in Junit.

If you see any feature not covered than let me know and I can add it.

FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/SlimSymbols/NamingConvention/content.txt
```diff
@@ -0,0 +1,31 @@
...
+
+!|test slim   |3            |
+|string       |getStringArg?|
+|ØÅÆØÅÆ_ØÅÆØÅÆ|$Æ_ØÅ=       |
+|$Æ_ØÅ        |ØÅÆØÅÆ_ØÅÆØÅÆ|
```
six42: Hi @hcf and @amolenaar

I added this test page with this UTF8 test: FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/SlimSymbols/NamingConvention

Think this should cover the required testing
and in addition provides the documentation to the community that this new feature has been added.

I am a big fan of "Eat your own dog food" and prefer adding wiki tests then Junit tests.
And I think this is a nice example that wiki tests can be more efficient than Junit as it takes less lines to express the feature in a wiki page than in Junit.

If you see any feature not covered than let me know and I can add it.

FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/SlimSymbols/NamingConvention/content.txt
```diff
@@ -0,0 +1,31 @@
...
+|foo   |$x3_=        |
+|$x3_  |foo          |
+
+!3 Referencing undefined Slim symbols names raise no erro but return the symbol name
+
+!|test slim              |
+|string    |getStringArg?|
+|$UNDEFINED|$UNDEFINED   |
+
+!3 Slim symbols names are case sesitive
```
hcf: Typo, sesitive -> sensitive

FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/SlimSymbols/NamingConvention/content.txt
```diff
@@ -0,0 +1,31 @@
...
+|fitnesse.slim.test|
+
+!3 Slim symbols names must start with a letter, followed by letters, digits and underscores
+
+!|test slim          |
+|string|getStringArg?|
+|foo   |$x3_=        |
+|$x3_  |foo          |
+
+!3 Referencing undefined Slim symbols names raise no erro but return the symbol name
```
hcf: erro -> error

FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/SlimSymbols/NamingConvention/content.txt
hcf: I don't know the inner workings of this table, but will it fail if someone should remove \\p{L} from the regular expressions by mistake?

src/fitnesse/slim/SlimSymbol.java
```diff
@@ -0,0 +1,81 @@
+package fitnesse.slim;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+abstract public class SlimSymbol {
+  public static final Pattern SYMBOL_PATTERN = Pattern
+      .compile("\\$([A-Za-z\\p{L}][\\w\\p{L}]*)");
+// This would be a better pattern as it allows to define the end of a symbol name with another $ sign
```
hcf: Why leave the outcommented pattern here? If it is a better pattern, why has it not been used? Maybe include the reason for not using it at least if the comment should stay here.

six42: The discussion for this commented code is in #780.

Yes the table would fail if you remove the new pattern. You can try with an older version of Fitnesse to prove it.

I will update the typos.

src/fitnesse/slim/StatementExecutor.java
```diff
@@ -131,7 +140,52 @@ private MethodExecutionResult getMethodExecutionResult(String instanceName, Stri
...
+   *  
+   * @return true is the fixture will handles symbols assignments and lookups itself.
+   *         This should be a rare exception and is not recommended
+   *         
+   *  Would be nice to use the classname but we don't know it at this point.
+   *         
+   */
+  
+  private Boolean ignoreSymbols(String instanceName, String methodName){
+	  try{
```
hansjoachim: This method (and the next) catch generic Exceptions. Though since the null values are handled, it's not clear to me what else could throw an exception here. Could you elaborate, and perhaps catch a more specific Exception? :)

src/fitnesse/slim/StatementExecutor.java
```diff
@@ -122,8 +127,12 @@ public Object callAndAssign(String variable, String instanceName, String methodN
 
   private MethodExecutionResult getMethodExecutionResult(String instanceName, String methodName, Object... args) throws Throwable {
     MethodExecutionResults results = new MethodExecutionResults();
+    Boolean ignoreSymbols = ignoreSymbols( instanceName,  methodName);
+	if (!ignoreSymbols){
+		args = context.replaceSymbols(args);
+	}
     for (int i = 0; i < executorChain.size(); i++) {
-      MethodExecutionResult result = executorChain.get(i).execute(instanceName, methodName, context.replaceSymbols(args));
+    	MethodExecutionResult result = executorChain.get(i).execute(instanceName, methodName, args);
```
hansjoachim: This line seems to have gotten some additional whitspace before and after.

src/fitnesse/slim/StatementExecutor.java
six42: Hi Hans Joachim,

Thanks for your valid comments.
Swallowing all exceptions without any message is not best practice.

I want these methods to never fail!

Just as a background:
I don't recommend anybody to use the flag  SLIM_AGENT_FIXTURE_HANDLES_SYMBOLS
and these methods. The better solution is a different table type for your fixtures which is processed line by line. 
I added this as I feared to break   https://github.com/smartrics/RestFixture. Their code is currently only working because of a bug in the symbol handling. This PR fixes the bug. I wanted to give them a simple path to continue to work without changing all their code.
In the mean time I expect they even don't need it as they use a different symbol pattern in their fixture.

But as long it is not merged and I get feedback I can be only 99.99% sure.

In case I am wrong and there is a wider use for this the exception handling can be improved. 

In case I am right it is rather something to get removed in the next release. It should just be a backward compatibility option.

I can also remove it right away.
It comes to the general question how to handle possible incompatibilities.

Issue #787, opened 2015-09-07
Query table not matching fields from left to right as intended according to documentation


hcf:
According to http://www.fitnesse.org/FitNesse.UserGuide.WritingAcceptanceTests.SliM.QueryTable, when matching rows in a query table "the fields are matched left to right". This is not the case, as each row is scored for every field, so given:

x | n | 2n
------------ | ------------- | ------
1 | 2 | 4
2 | 3 | 6

According to the documentation, a result list with x=1, n=3 and 2n=6, should give:

x | n | 2n
------------ | ------------- | ------
1 [OK] | 2 [Was 3] | 4 [Was 6]
2 [Missing] | 3 | 6

but at the moment we are getting

x | n | 2n
------------ | ------------- | ------
1 [Missing] | 2 | 4
2 [Was 1] | 3 [OK] | 6 [OK]

What is the intended behaviour here? The way it is, or the way it is documented?

amolenaar:
Closed, since PR has been merged

PR #788, opened 2015-09-07
Fixes issue #787

hcf:
The query table is not matching from left to right as described in the
documentation.


hcf:
Anyone have the time to have a look at this PR? @amolenaar?

amolenaar:
I think your patch is okay.

PR #791, opened 2015-09-17
Utf 8 characters in symbols

hcf:
Making sure we are able to parse symbols containing utf-8 characters.


six42:
Hi @hcf ,

How do you retrieve the symbols you set with UTF  character?
if I didn't missed anything you just changed the assignment pattern but not the retrieval pattern.
What is your use case?  Can you provide a wiki testpage?

hcf:
Hi @six42 and @amolenaar. You are indeed correct in that my PR is not correctly working. I have added failing tests now, and should be easy enough to fix in my PR, or should I create a new PR or in this one?

Issue #784, opened 2015-09-01
Problems with CodeMirror plugin for page editing


philippe-granet:
Syntax highlighting in plain text editor is great but there is few bugs:

* On some larges tests, cursor disappear when scrolling (Chrome and Firefox), Workaround : click on "Format" button, then cursor  reappear

* problem with horizontal scroll on Firefox (not Chrome) when some lines length are superior to body width size, horizontal bar appears at the bottom of page,  workaround : override CodeMirror css with this:
.CodeMirror {
  position: absolute !important ;
  top: 200px !important ;
  left: 0 !important ;
  right: 0 !important ;
  bottom: 0 !important ;
}

* Ctrl+F doesn't work well, could you add this codemirror feature : https://codemirror.net/demo/search.html

* codemirror.js is heavy to load by navigator : 314Ko and not minimized

* it uses version 4.12.0 of codemirror , it seeems to be an old version : http://codemirror.net/doc/releases.html , lots of bugs have been patched since this version

staniere:
Hello, I confirm problem of searching text with new higligth syntax (tested on internet explorer 9.0 and firefox 40.0.3)

DmitriyKartashov:
The problem with text searching is that CodeMirror is rendering only part of the text.
I extended the renderable area to infinity. Now it will work.
On my office PC I tested a page with 11k lines of text. I didn't find any performance problems. But for sure, in same cases it could become slower.

amolenaar:
Is this issue resolved now?

staniere:
hello, 
how could I test the solve ? (it seem no new release has been sent since 20150814)
could I get a beta version of fitnesse-standalone.jar	 somewhere ?

hansjoachim:
>could I get a beta version of fitnesse-standalone.jar somewhere ?

Look under Edge on the [download page](http://fitnesse.org/FitNesseDownload). It contains the latest green build from the development branch. :)

staniere:
hello, 
I confirm CTRL+F is working fine on entire page now (build #912)

amolenaar:
Cool! Thanks.

PR #793, opened 2015-09-17
Plain text editor fixes.

DmitriyKartashov:
Pull request for
Problems with CodeMirror plugin for page editing #784
Fix Ctrl+F problem

Correct regexes a bit/
For example !content looked like:
**!c**ontent


amolenaar:
Nice fix!

PR #795, opened 2015-09-21
autocomplete for moving of tests

gmeiser:
Hello,

I created an HTML 5 autocomplete feature to use with the "move" function.
The feature requires a newer jquery version and the jquery UI files.

Now the move function shows available suites as predefined path if you type in more than 1 char.

The autocomplete function creates a list of all available suites and a list of all available tests (which is not used at this moment).

Kind regards

Gordon


amolenaar:
Autocompletion is a great addition.

It needs some work still, though:
`SimpleResponse` should remain a simple response object. Writing a file with page names is not what I would expect. Instead I would expect a service (e.g. `NameWikiPageResponder`) to serve the names.
The css contains to much styling.

gmeiser:
Hi Arjan,
I will try to implement the autocompletion feature the way you suggested it.
As I am new to the Fitnesse framework (and unfortunately found no detailed implementation guidline or something like that) it would help me a lot if you could state more precisely how you would implement it.

I understood the following:
- you would not use the jquery-ui file as this results in too much overhead only for the one feature
- SimpleResponse should stay clean, the implementation should be done in a separate file
- best way should be to implement the feature as new service
- writing the available suites and tests to a file should be circumvented

amolenaar:
Hi G.,

Thanks a lot for looking into this. Having autocomplete on several locations would be a huge addition. Maybe we can even add it to the edit page at some point :)

> On 28 Sep 2015, at 10:00, gmeiser <notifications@github.com> wrote:
> 
> Hi Arjan,
> I will try to implement the autocompletion feature the way you suggested it.
> As I am new to the Fitnesse framework (and unfortunately found no detailed implementation guidline or something like that) it would help me a lot if you could state more precisely how you would implement it.

In general, I think it’s a good idea to keep FitNesse minimal on the web site (the edit page being a notable exception to the rule ;) ). I was looking for how to get auto-complete functionality in a HTML 5 way and they have invented datalists for that. I think that would be a nice starting point, eventually adding a shim library for the browsers that do not support datalists.

At some point it would be great to get rid of JQuery all together. I mean: modern browsers support all the features we use from JQuery out of the box. Only we need some shims to support older browsers.

> I understood the following:
> 
> 	• you would not use the jquery-ui file as this results in too much overhead only for the one feature

If we can pull it of with datalists, that would be (more) awesome

> 	• SimpleResponse should stay clean, the implementation should be done in a separate file

Indeed. SimpleResponse is just a response object, so it contains the HTTP headers and response body.

> 	• best way should be to implement the feature as new service

Either that, or the pages should be parsed when the page is rendered, which might be the simpler approach for a start.

> 	• writing the available suites and tests to a file should be circumvented

Yes. I think there is no reason to put this information in a file. Let’s first prove there is a memory/speed issue with just processing the pages and add improvements later.

Hope this helps,

Regards,

Arjan

six42:
Slightly of topic but as @amolenaar mentioned autocomplete for page edit.
The cod mirror editor which is now used by fitnesse has autocomplet functionality.
Maybe somebody is interested to give it a try.

gmeiser:
Hey Arjan and six42,
the messages of you helped me a lot to focus in the right direction.
I will start to evaluate the possibilities to implement the autocomplete feature to Fitnesse.

First step will be to implement the autocompletion feature when "moving" tests in the way Arjan recommended it (HTML5 without usage of jquery, datalists, ...).
Second step on my site will be the evaluation of the places where autocomplete could be useful (maybe one of you already knows the places so I can skip this step :) ).
And finally I will take a look at the possibilities of the Code Mirror editor. Currently I am using a python script (which runs at every build process of fitnesse) to generate a jstree with all methods implemented in the different fixture files. This tree is searchable and the selection of a method copies the "fitnesse-usable" call of the method to the clipboard and can be directly inserted/pasted to the fitnesse editor. 
Although using the hint addon/feature of the Code Mirror editor would be a much better way supply the user with the methods which are available.

Last but not least we should think about the support of older browsers (as Arjan mentioned) although I personally think that HTML5 is coming up more and more and nearly all browsers will support it completely in the near future. 

As soon as I have something to show you, I will let you know.

Greetings Gordon

gmeiser:
Opened a new pull request here:

https://github.com/unclebob/fitnesse/pull/867

PR #801, opened 2015-09-29
System under test annotation to support multiple annotated fields

SamCooper:
This small change allows a Slim fixture to contain more than one annotated field and search across all of them.

Previously it would just return the first field with the correct annotation, now it checks to see if it contains a compliant method otherwise it moves to the next field.

It allows you to create orchestration style tests that verify several components are behaving correctly together.

Another nice feature now is that it allows you to create fixtures via aggregation rather than inheritance, where test functionality is supplied as components and aggregated by a top level fixture class. Although the annotation name 'SystemUnderTest' does not make so much sense then.

It might make sense therefore to change the name of the annotation to make it more apparent of this extra behaviour?

### DIFF
--- a/src/fitnesse/slim/SystemUnderTestMethodExecutor.java
+++ b/src/fitnesse/slim/SystemUnderTestMethodExecutor.java
@@ -17,7 +17,7 @@ public class SystemUnderTestMethodExecutor extends MethodExecutor {
     } catch (SlimError e) {
       return MethodExecutionResult.noInstance(instanceName + "." + methodName);
     }
-    Field field = findSystemUnderTest(instance.getClass());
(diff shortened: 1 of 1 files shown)


amolenaar:
Nice change! Good to know the `@SystemUnderTest` annotation is actually used. :)

six42:
Hi Sam
This sounds very powerfull.

Would you mind to add a wikitest page that demonstrates the feature? 
This would ensure that many people would start using your feature.
Also prevents that any future code change breaks it.

SamCooper:
I can try if you point me at how to do that.

What are your thoughts on changing the annotation name? Probably not desirable but the pedant in me gets upset by things like badly named variables! :-)

PR #802, opened 2015-09-30
Unused imports

hansjoachim:
Removed unused imports.


amolenaar:
Thanks! Are those my left-overs?

hansjoachim:
Probably :p A couple more showed up, they should be taken care of in #808.

PR #805, opened 2015-10-01
update markup variables page to include underscores

a1dutch:
updates the markup variables page to include underscores as valid characters as per the define wiki parser.


amolenaar:
Thanks!

PR #811, opened 2015-10-03
Replace use of StringBuffer with StringBuilder and String

hansjoachim:
Replaced use of StringBuffer since it has synchronized methods which has a negative impact on performance. Additionally, a couple of places could use a simple String instead.

(As a bonus, also a few places where foreach could be utilized.)

I left some cases alone, like Logger.formatLogLine(LogData) and various toString-methods where it wasn't clear if a single String would increas or hinder readability.

### DIFF
--- a/src/fit/Fixture.java
+++ b/src/fit/Fixture.java
@@ -257,7 +257,7 @@ public class Fixture {
   }
 
   public static String camel(String name) {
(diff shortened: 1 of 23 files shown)


amolenaar:
I appreciate your eye for detail. Thanks!

hansjoachim:
Thank you. :)

PR #813, opened 2015-10-04
Issues with modifiers and related

hansjoachim:
Mostly concerning modifiers:
* List modifiers in consistent order
* Removed implicit modifiers from interfaces. Implementations will need to set the Overriden methods as publicly visible in order to fullfil the interface contract, so they are mostly redundant. I did leave the static modifiers though, since they may count as an implementation hint even though it isn't strictly enforceable.

As I mentioned in #812, it seems I was missing out on testscope locally, so I've went over and did some more StringBuffer -> StringBuilder fixes there. In a couple of cases it was used to provide similar functionality to StringUtils.join() so I simply replaced these.

Otherwise see individual commit messages for (more) details.
Particularly note the interface CompositePageFinder which is unused, and everything builds fine without it. It is part of the public API though and may have external use, so I'm a bit vary of simply removing it.

### DIFF
--- a/src/fit/FixtureListener.java
+++ b/src/fit/FixtureListener.java
@@ -4,7 +4,7 @@
 package fit;
 
 public interface FixtureListener {
-  public void tableFinished(Parse table);
+  void tableFinished(Parse table);
 
-  public void tablesFinished(Counts count);
+  void tablesFinished(Counts count);
 }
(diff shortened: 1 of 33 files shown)


amolenaar:
You're on a roll! :)

PR #817, opened 2015-10-10
Ensured loggers are private static final

hansjoachim:
As found by SonarQube, made loggers private static final. This is to ensure that 1) other classes cannot use them to write log messages, 2) there is a common logger for the class regardless of the number of objects and that 3) it is initialized in one place only.

Note that a couple of these were protected, but not used in subclasses in FitNesse itself. Though of course it is possible that external subclasse may have picked them up. Normally I'd be wary of breaking existing public API, but I think creating a more specific logger in those subclasses should be reasonably straigth-forward. 

Note that ContextConfigurator and FitNesseContext both have additional fields which holds loggers, but I haven't touched these.

### DIFF
--- a/src/fitnesse/reporting/BaseFormatter.java
+++ b/src/fitnesse/reporting/BaseFormatter.java
@@ -9,10 +9,8 @@ import fitnesse.testsystems.TestSystem;
 import fitnesse.wiki.WikiPage;
 
 import java.io.IOException;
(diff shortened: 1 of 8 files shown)


amolenaar:
Given proper loggers were introduced not so long ago, I'd expect them not to be used. But as you already mentioned, giving them their own logger probably makes more sense anyway.

PR #818, opened 2015-10-10
Another batch of issue fixes

hansjoachim:
Fixed some more issues:
* Use Map.entrySet() instead of keySet() for places where both key and value are used. Should be easier to just iterate through the pairs instead of having to look up the values all the time. Fixed in most places where the for loop accessed both keys and values except from MemoryFileSystem.getPayload(File). In this case it will immedatially return the value, and looks like it will necessarily do this each time. So I suspect always loading all values might make things worse rather than better. (I haven't profiled it though, so this is just guesswork.)
* Put some modifiers in consistent order
* Placed array designator on the type instead of the variable to make it easier to see what the type is.
* Added yet more Override annotations. These were found by Eclipse, but not IntelliJ and should hopefully be the last ones.
* Replaced for loops with more readable foreach loops.
* And some unused imports

A couple of questions:
*FileUtil.writeLinesToFile takes in a list of apparently anything. In FitNesse, it is only called with a list of Strings though, which makes sense. I wonder whether the list of lines should/could be anything but String? (It is a public method though, so I don't know how it might be called outside of FitNesse.)
*TableResultExtractor looks completely unused?

### DIFF
--- a/src/fit/FixtureName.java
+++ b/src/fit/FixtureName.java
@@ -21,6 +21,7 @@ public class FixtureName {
       this.nameAsString = tableName;
   }
 
+  @Override
   public String toString() {
     return nameAsString;
   }
--- a/src/fit/FixtureName.java
+++ b/src/fit/FixtureName.java
@@ -4,7 +4,6 @@
 package fit;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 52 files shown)


amolenaar:
`TableResultExtractor` is used in an acceptance test `SlimSymbolManipulation`.

amolenaar:
It looks to me like `FileUtil.writeLinesToFile` can only handle strings, since inside the loop each line entry is cast to a String anyway.

hansjoachim:
>TableResultExtractor is used in an acceptance test SlimSymbolManipulation.

Ah, right. I keep forgetting those usages don't show up. :p

PR #821, opened 2015-10-14
Class Path with Space

six42:
This should fix the issues described in #803 and #726

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteTestSystems/ClassPathWithSpace/content.txt
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteTestSystems/ClassPathWithSpace/content.txt
@@ -1,0 +1,10 @@
+!3 A space in the classpath must still allow to start the test system
+!path abc xyz
+
+!3 A trailing space in classpath definition  must still allow to start the test system
+!define TS (trailingspace )
+!path ${TS}
+#!path trailingspace 
+
+|script              |
+|check|echo|pass|pass|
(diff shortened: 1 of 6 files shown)


six42:
Hi @Thurmann and @hansjoachim,

can you try this PR if it fixes your path problem in all situations.

hansjoachim:
Thanks for making this patch. I haven't reviewed it in detail, but at a glance, it looks like this would help.

Unfortunately I haven't been able to properly test this. (#726 seem to indicate the problem only occurs on Windows and I have had problems trying to reproduce it on other platforms. And on the system where I am able to reproduce it, I don't have a build environment set up for FitNesse. If this ends up being merged, I believe FitNesse has some development/daily builds, so I could probably test with that. (Or if it comes down to it, I could possibly build it and transfer the jar or something...))

amolenaar:
I'm fine with merging this. The patch looks fine.

amolenaar:
Ouch, this PR makes the build fail (on unix). I'm not sure why, the PR seems okay.

six42:
:-( 
Any error message?

amolenaar:
It's complaining about a class not found for VelocityContext. That's the last jar on the class path, hence it also has the closing quote attached to it.

amolenaar:
I now changed it in such a way that the quoting is only applied for Windows. All parameters that contain spaces are wrapped in quotes. For Unix, there's no such need.

hansjoachim:
I've now had the time to test with a recent edge build of FitNesse, and things were working great! :) The classpath looked as expected, and I had no problems running the test suite. Thank you.

Since this was switched to a Windows-only fix/"feature", is there still a test in place for this? Just wonder whether we'll be able to catch regressions related to this or whether one needs to be on the lookout?

PR #824, opened 2015-11-01
Detailed diff in case of a failed Test

jdufner:
For description of the patch see: https://github.com/jdufner/fitnesse/wiki/Comparison-of-Test-Results


six42:
Hi Juergen,
Nice feature,
thanks for adding this.

Could you add your wiki page to the acceptance test suite of Fitnesse and  add more examples to your wiki how the result will look like:

If a symbol is used
regular expression compare is used
Number compare is used
Custom comparator is used
First letter is different
Last letter is different
no matching text at all
No obvious match: Hello World vs Wellcome Lord

jdufner:
Hi six42,

I added an acceptance test in my fork https://github.com/jdufner/fitnesse/tree/feature/detailed_diff_failing_tests . I took screenshots and put they on my wiki page: https://github.com/jdufner/fitnesse/wiki/Comparison-of-Test-Results

I didn't update the pull request yet. Send me feedback what to do now, please.

test/fitnesse/html/HtmlDiffUtilTest.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @Test
+  public void testBuilderClosingTag() {
+    String actual = "abc";
+    String expected = "axc";
+    String htmlDiff = null;
+    htmlDiff = new HtmlDiffUtil.ActualBuilder(actual, expected)
+        .setClosingTag("</span class=\"diff\">").build();
+    Assert.assertEquals("a<span class=\"diff\">b</span class=\"diff\">c", htmlDiff);
```
amolenaar: Shouldn't this read:

    Assert.assertEquals("a<span class=\"diff\">b</span>c", htmlDiff);

(no class in closing tag)?

ivy.xml
```diff
@@ -35,7 +35,7 @@
 		<dependency org="commons-lang" name="commons-lang" rev="2.6" conf="default,standalone,maven->default" />
 		<dependency org="org.json" name="json" rev="20140107" conf="default,standalone,maven->default" />
 		<dependency org="org.apache.ant" name="ant" rev="1.9.4" conf="default" transitive="false" />
-		<dependency org="com.googlecode.java-diff-utils" name="diffutils" rev="1.3.0" conf="default,standalone,maven->default"/>
+		<dependency org="com.googlecode.java-diff-utils" name="diffutils" rev="1.3.0" conf="default,slim,standalone,maven->default"/>
```
amolenaar: Slim should not require diff-utils. You do not want to pull in extra dependencies on the slim server (`fitnesse.slim`)

test/fitnesse/html/HtmlDiffUtilTest.java
jdufner: You are right, closing tags don't need a class attribute. I put it in the closing tag to demonstrate and test the builder pattern. The default value of the closing tag is still `</span>`.

Do you have another suggestion for this test case?

amolenaar: Maybe a completely different tag? Or a close tag with a custom tag (`</span><b>extra</b>`). The class in the close tag confused me.

ivy.xml
jdufner: Could you explain that? I dont get it.
* Do you want to copy or rewrite diff-utils? Fitnesse has already a dependency to diff-utils.
* Do you want, that Slim has no dependencies? Slim has already a dependency to htmlparser.

Do you have any other suggestions to implement this feature?

test/fitnesse/html/HtmlDiffUtilTest.java
jdufner: Nice idea. I'll update the PR.

six42:
Hi Juergen,

Wow, impressive list of tests. Thanks.
And all show meaningful formatting with your extension.
The combination with symbols could be improved, but in my opinion not urgent.

To make sure that no future PR breaks your feature you should add your page to the accptance test suite.
Have a look at the below two pages to see how you can setup a test with expected failed results.


http://fitnesse.org/FitNesse.SuiteAcceptanceTests.SuiteSlimTests.IgnoredTestsShouldColorGray


http://fitnesse.org/FitNesse.SuiteAcceptanceTests.SuiteSlimTests.TestCustomComparators

ivy.xml
amolenaar: I mean that the diff-utils are not used by the slim service (the bit that's executing stuff on a SUT).

Instead, all comparison is handled by FitNesse itself (the `default` configuration). So this change can be reverted.

jdufner: The interface `fitnesse.testsystems.TestResult` is independent from the subsystems either `fit` or `slim`, but the the visual presentation is in `fitnesse.testsystems.slim.HtmlTable.Cell.formatTestResult`. My patch concerns the visual representation in Slim only.

jdufner:
Hi six42,
as you suggest, I implemented a acceptance test similar to http://fitnesse.org/FitNesse.SuiteAcceptanceTests.SuiteSlimTests.IgnoredTestsShouldColorGray.

six42:
Great,

Can you share the link.

jdufner:
The wiki content is https://github.com/unclebob/fitnesse/pull/824/files#diff-d79763babcd41404d8002f1134af096d 

Did you mean that?

jdufner:
Hi amoleenar,
Hi six42,

I recently updated my pull request. Could you please make clear your expectations to accept this pull request. I understood that you dislike the dependency of slim to diff-util. Please tell me your preferred solution.

Thanks in advance

amolenaar:
I like the PR, I like the feature.

I don't think we need the screenshots, unless they're in documentation in which case they should be put in `FitNesseRoot/files` (or they won't be packaged with the distribution).

The dependency of slim to diff-utils is not needed, since 'slim' only covers the package 'fitnesse.slim' (the slim server). Class `HtmlDiffUtils` can be moved to `fitnesse.testsystems.slim`.

I sent you a pull request with my proposed changes: https://github.com/jdufner/fitnesse/pull/2.

six42:
Hi Juergen,
This is really a nice feature.
Thanks also for the great acceptance test and even fixing the display with symbols.

jdufner:
Hi Six42,
thank you for compliment. Do you know, when you will do the next release?

2015-11-26 2:01 GMT+01:00 six42 <notifications@github.com>:

six42:
Hi Juergen, 
this is for @amolenaar to decide.
I hope it is only after my PRs are merged ;)

amolenaar:
:D Normally I try to merge all outstanding pull requests before doing a release. I plan on releasing before Christmas, depending on my spare time.

Issue #746, opened 2015-05-30
FTBFS with openJDK9


hansjoachim:
I've tried to test a bit with one of the early openJDK9 builds on Ubuntu Wily and discovered that FitNesse fails to build with JDK9. When running `ant compile`, I get the following error message:
```
    [javac] /home/ubuntu/fitnesse/src/fitnesse/slim/converters/ConverterRegistry.java:48: error: incompatible types: Converter<CAP#1> cannot be converted to Converter<T>
    [javac]     Converter<T> converter = getConverterForClass(clazz, null);
    [javac]                                                  ^
    [javac]   where T is a type-variable:
    [javac]     T extends Object declared in method <T>getConverterForClass(Class<? extends T>)
    [javac]   where CAP#1 is a fresh type-variable:
    [javac]     CAP#1 extends T from capture of ? extends T
``` 

As mentioned this is an early build (specifically 9~b64-1ubuntu1), so the bug is not necessarily in FitNesse, but I thought I'd report it just in case.

hansjoachim:
The lines in question was recently refactored/simplified, but FWIW I still get the same error message
```
    [javac] /home/ubuntu/dev/fitnesse/src/fitnesse/slim/converters/ConverterRegistry.java:53: error: incompatible types: Converter<CAP#1> cannot be converted to Converter<T>
    [javac]     return getConverterForClass(clazz, null);
    [javac]                                ^
    [javac]   where T is a type-variable:
    [javac]     T extends Object declared in method <T>getConverterForClass(Class<? extends T>)
    [javac]   where CAP#1 is a fresh type-variable:
    [javac]     CAP#1 extends T from capture of ? extends T
```
(OpenJDK9 9~b80-2 on Ubuntu Wily)

I have to admit I'm not quite certain of what the error message is trying to say...

amolenaar:
Does this have to do with T being a generic type itself?

hansjoachim:
Might be.

I read a bit on http://stackoverflow.com/questions/25914289/incompatible-types-required-classt-found-classcap1-where-t-is-a-type-vari and have a simple fix/workaround now. For now, I'm just casting the value to Converter<T> before returning which might not be the best, but I can look a bit more into it. With that in place, all the source code builds successfully with JDK9, though unfortunately I ran into a couple of test failures:
```
    [junit] Tests run: 12, Failures: 3, Errors: 0, Skipped: 0, Time elapsed: 0.097 sec
    [junit] 
[...]

PR #835, opened 2015-11-21
Make FitNesse buildable with openJDK9

hansjoachim:
Adding a cast in order to build with openJDK9 (tested with 9~b88-1 on Ubuntu Xenial)

While FitNesse now builds, there are test failures, so #746 isn't completely fixed by this. I'm looking more into what's causing the failing tests.

Also built successfully with openJDK 8 to verify it also work for older JDKs.

### DIFF
--- a/src/fitnesse/slim/converters/ConverterRegistry.java
+++ b/src/fitnesse/slim/converters/ConverterRegistry.java
@@ -49,9 +49,8 @@ public class ConverterRegistry {
     }
   }
 
-  public static <T> Converter<T> getConverterForClass(Class<? extends T> clazz) {
-    // While the casting may be redundant, it is required for JDK9.
-    return (Converter<T>) getConverterForClass(clazz, null);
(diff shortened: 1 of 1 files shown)


src/fitnesse/slim/converters/ConverterRegistry.java
```diff
@@ -50,7 +50,8 @@ protected static void addStandardConverters() {
   }
 
   public static <T> Converter<T> getConverterForClass(Class<? extends T> clazz) {
```
amolenaar: I found it can also be fixed by changing `Class<? extends T>` to `Class<T>`. That might be nicer than introducing another cast.

hansjoachim:
You're right. I've restricted the parameter instead.

I've also looked more into the version number thing with the tests I mentioned and it doesn't seem to be FitNesse-specific. Still not sure if it is a bug or not, but I have a minimal test case for it which displays the same behaviour.

If I ignore StackTraceEnricherTest, FitNesse builds successfully now with JDK9. :)

PR #837, opened 2015-11-23
In-process Slim

amolenaar:
This PR makes Slim work really in-process (with the `&debug` option or when executed from JUnit). Previously a socket was opened to communicate between FitNesse and the SUT, although they were running in the same process. The socket has been replaced by a `MockSocket`.

### DIFF
--- a/src/fitnesse/slim/JavaSlimFactory.java
+++ b/src/fitnesse/slim/JavaSlimFactory.java
@@ -35,7 +35,11 @@ public class JavaSlimFactory extends SlimFactory {
   }
 
   public static SlimFactory createJavaSlimFactory(SlimService.Options options) {
-    return new JavaSlimFactory(options.statementTimeout, options.verbose);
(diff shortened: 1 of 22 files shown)


jplambert:
Hi,

I guess this is only for Java SUT? Or would it work with other Slim server implementations?

Thanks in advance.

amolenaar:
Hi,

This only works for a Java SUT. The SUT needs to run on the JVM at least.

jplambert:
That's what I thought.

I was asking because we are doing PHP and even though the Slim protocol over sockets provides a nice abstraction, in practice the PHP is always local and it would significantly improve the run time to avoid all the network stuff.

So far we thought that we could define a whole new Test System in FitNesse to do exactly that.

amolenaar:
You can create your own test system, indeed.

I think using stdin/stdout instead of a socket would be a nice alternative.

PR #838, opened 2015-11-24
Smart Search Screen

six42:
The search results screen shows now also the search input form.
If a user didn't found what he was looking for he can modify the
search and start a new search with one click.
The search form also allows to define the search scope: the area of the
wiki that will be included in the search. (marked yellow in the screenshot)
In the past the text search was always global and the properties search limited to the current page and its childs. This was confusing.
To save space on the screen the search form uses now tabs and tooltips
smartsearchscreen


amolenaar:
Really nice. I think the tabs could be made a little more explicit, but I'll merge it so that anyone that wants to use the edge version can play with it.

six42:
Thanks for merging.

I wasted two days to get the tabs more visible but had no success. Would be happy if any css guru improves this.

amolenaar:
I do not consider myself a css guru, but can have a look ;)

PR #839, opened 2015-11-25
Improvements for pull #643

reneintersys:
We made following improvements to fullfill the requirements by jediwhale in https://github.com/unclebob/fitnesse/pull/643

- Added FitNess unit tests
- Removed Java doc style comments
- Do not swallow exceptions

### DIFF
--- a/FitNesseRoot/FitNesse/UserGuide/FitNesseWiki/MarkupLanguageReference/EnhSysDate/content.txt
+++ b/FitNesseRoot/FitNesse/UserGuide/FitNesseWiki/MarkupLanguageReference/EnhSysDate/content.txt
@@ -1,0 +1,25 @@
+'''!-!sysdate-!''' defines a variable which holds a time/date in a defined format and calculates a given offset
+
+'''Syntax'''
+{{{!sysdate variablename ({definition of the time/date format} definition of offset)}}}
+
+|'''Markup Text'''                                       |'''Displays as'''                                   |'''Variable defined'''|'''Remark'''                                         |
+|!-!sysdate date0 ({dd.MM.yyyy HH:mm:ss})-!              |!sysdate date0 ({dd.MM.yyyy HH:mm:ss})              |${date0}              |current date/time in the defined format              |
+|!-!sysdate date1 ({yyyy-dd-MM'T'HH:mm:ss})-!            |!sysdate date1 ({yyyy-dd-MM'T'HH:mm:ss})            |${date1}              |current date/time in the defined format              |
+|!-!sysdate date2 ({yyyyMMddHHmmss})-!                   |!sysdate date2 ({yyyyMMddHHmmss})                   |${date2}              |current date/time in the defined format              |
+|!-!sysdate date3 ({01.MM.yyyy 00:00:00})-!              |!sysdate date3 ({01.MM.yyyy 00:00:00})              |${date3}              |first day of the current month                       |
+|!-!sysdate date4 ({dd.MM.yyyy 23:59:59} last_of_month)-!|!sysdate date4 ({dd.MM.yyyy 23:59:59} last_of_month)|${date4}              |last day of the current month                        |
+|!-!sysdate date5 ({dd.MM.yyyy HH:mm:ss} +1h)-!          |!sysdate date5 ({dd.MM.yyyy HH:mm:ss} +1h)          |${date5}              |current date/time with thime shift +1 hour           |
+|!-!sysdate date6 ({dd.MM.yyyy HH:mm:ss} -4m +3h)-!      |!sysdate date6 ({dd.MM.yyyy HH:mm:ss} -1y -4m)      |${date6}              |current date/time with thime shift -1 year -4 minutes|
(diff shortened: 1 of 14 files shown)


six42:
Hi Rene,

sounds interesting.

I am not sure if I understood how and when to use these features.
can you add pages to the user guide or acceptance test suite which demonstrates the usage and the benefit.

something similar to:

http://fitnesse.org/FitNesse.FullReferenceGuide.UserGuide.FitNesseWiki.MarkupLanguageReference.TodaysDate

http://fitnesse.org/FitNesse.SuiteAcceptanceTests.SuiteWidgetTests.TestInclude

reneintersys:
Hi @six42 

We added documentation to the user guide - could you please have a look?

http://localhost:8001/FitNesse.FullReferenceGuide.UserGuide.FitNesseWiki.MarkupLanguageReference.EnhSysDate
http://localhost:8001/FitNesse.FullReferenceGuide.UserGuide.FitNesseWiki.MarkupLanguageReference.RandNumber
http://localhost:8001/FitNesse.FullReferenceGuide.UserGuide.FitNesseWiki.MarkupLanguageReference.RandString

amolenaar:
Just one remark: the class SymbolProvider.java has been completely reformatted. Basically opening curly bracket should be placed after the statement (e.g. `if (something) {`). The complete reformat of the file makes it really hard to figure out what changed.

nextremos:
@amolenaar Corrected the formatting for SymbolProvider.java

reneintersys:
@amolenaar 
We corrected the formatting for SymbolProvider.java

six42:
Thanks so much

No I understand how to use it.

Your markup combines value calculation and assignment to a variable in a single step.

I think keeping this seperate would increase the flexibility

Your first example could also be written with the existing markup like this:

!define date0 {!today}

Any reason why you didn't extended the !today markup with your additional and very useful calculation options?

How do you like the idea of just creating markup for the calculation of random elements and leaving the assignment to !define?

reneintersys:
@six42 

> How do you like the idea of just creating markup for the calculation of random elements and leaving the assignment to !define?

One advantage is readability, from our perspective it's easier to read
*!sysdate time ({dd.MM.yyyy HH:mm:ss})*
than 
*!define time {!sysdate ({dd.MM.yyyy HH:mm:ss})}*

Another aspect is, that the variable is directly initialized and the value is visible on the page. As we have a lot of nested pages it's pretty useful to regain a defined variable. Of course this is only a reason for values which change not too often (E.g. dates)

*!sysdate zeit ({dd.MM.yyyy} +1m)*
**sysdate defined: zeit=02.12.2015** // Here we can simply take the value "02.12.2015" and search in the parent pages for the defined value

*!define time {today}*
**variable defined: time=today** // No idea what the value is

> Any reason why you didn't extended the !today markup with your additional and very useful calculation options?

This is a historical reason. As we use FitNesse for a long time and very intensive - we started to build our own set of widgets and other tools which seemed helpful for us.

At the end of the day, we would like to contribute to the project as FitNesse is a major framework in our company. This is our first trial and we learn the coding guidelines, the process to commit and much more.

amolenaar:
I agree with @six42 that it is more consistent to keep value calculation as assignment separated. Also it would make sense to merge !sysdate and !today.

jplambert:
Guys,

All these discussions about time are so familiar.

Here too we defined our own variation of 'today' which was able to format the time in any format.

Truth is, as much we found it nice at first, as much now we are moving away from it and favoring fixtures that output time so that we can assign it in a symbol.

Why that? Simply put, because variables in FitNesse is equivalent to static text replace. That means that using the variable twice, will lead to evaluate the current time twice. Which may have changed meanwhile. As a result, the test fails even though the code behaves properly.

Obviously, using symbols solves the issue, but that requires custom fixtures specific to your project, and it is less efficient than generating the time by the Wiki engine.

I am curious to know if the "all-in-one compute time and assign" method by reneintersys helps solving this issue.

In any case, I can only can everybody of the pitfalls of writing tests manipulating time, especially in FitNesse. Using !today to greet the readers with the current day is fine; using it to test stuff is not.

jplambert:
... I just remembered of another issue we had with time tests in FitNesse.

Let's say you need the same time, but in two or three different formats. This could happen easily in some of our tests, mixing ISO 8601 for dates in API, SQL date_time format in database, and unix-time/second since Epoch from many other places in the code.

Here we are not even using several times the same variable; we are asking for the time once for each time format. Again, time can change between each time generation, leading to random false-positives...

To handle such cases, we use a first fixture to generate time in Epoch format and store it in a symbol. Then we use another fixture to convert any Epoch format time to any other format, ensuring that both time representations are indeed denoting the same time.

Advantages of this solution:
- it works as we expect it to work (no counter-intuitive stuff with the variables)
- very flexible
- rather ready to implement, actually

I would not say the same of adding new keyword into the FitNesse Wiki....

reneintersys:
Thanks for all your comments and recommendations. We decided to not continue with this pull request as we ran out of time and budget. Please feel free to take over parts of the code you like.

Pull request can be closed.

PR #854, opened 2015-12-27
Make negative tests more readable

six42:
As a FitNesse developer
To test and see the behaviour of FitNesse with negative tests 
I run FitNesse in a sub instance 
and I need an easy way to see if the test did what I expected.

Added a scenario to extract the result of a single cell
Added a scenario to extract and match the result summary  
Added a scenario to show the result and input of a negativ test case in
HTML format
Added helper scenarios to hide the setup of the sub instance

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/ScenarioLibrary/content.txt
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/ScenarioLibrary/content.txt
@@ -4,6 +4,7 @@
 
 !|scenario|given page|page|
 |given page|@page|with content|nothing|
+|$CONTENT=|echo||
 
 !|scenario|given test page|page|
 |given page|@page|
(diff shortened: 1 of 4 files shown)


amolenaar:
This makes for a nice readability improvement. Is there a particular reason why you refactored those tests?

amolenaar:
... only the tests fail now :( Any clue why? A short investigation on my side did not show a solution.

six42:
strange, works nice for me.
Did you rebuild the Fitnesse-Standalone.jar and run the test with this?
The only thing I changed is removing the additional classpath elements.
can you send me the executionlog and testresult.xml?

amolenaar:
I started FitNesse with `ant run`. Which class path elements did you remove? (I can't make it up from the changes).

six42:
I will check once more with a clean build. Maybe I missed something.

six42:
Now it works. 
The classpath was the issue. During junit testing the standalone.jar is not build and therefore the classpath to the classes folder is required.

six42:
Closing as same code is also in #857.

PR #877, opened 2016-02-22
Resetting min. Java version needed to 1.7

benhamidene:
Many of our enteprise projects are still using Java 7 (and even Java 6). Upgrades to java 8 are planned but not in the near future as they are related to major changes in our runtime and application stacks.
That´s why we would appreciate to postpone the upgrade of FitNesse to Java 8 as it would push us either to stick to the december release or to fork FitNesse. We think that both ways are not necessary and not wished.

[Review on Reviewable](https://reviewable.io/reviews/unclebob/fitnesse/877)

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -6,7 +6,7 @@
 
     <property environment="env"/>
     <property name="env.FITNESSE_TEST_PORT" value="8080"/>
-    <property name="min.java.version" value="1.8"/>
+    <property name="min.java.version" value="1.7"/>
     <property name="src.dir" location="src"/>
     <property name="test.src.dir" location="test"/>
     <property name="classes.dir" location="classes"/>


bknobloch:
I second the Java 7 compatibility request. 

Around here we're running fitnesse tests for several development teams in Java 7 maven builds using the junit runner. For that we need all the required runtime components to be compatible to java 7 not only the slim client.

Surely Java 8 opens new options, but I believe that there ist much room for new or improved things in fitnesse while staying Java 7 compatible for some time.

nbyl:
+1

A lot of our clients are still running on Java 7. It is still publicly supported until the end of april (http://www.oracle.com/technetwork/java/eol-135779.html) and for enterprise customers even longer. Therefore preserving Java 7 as the compiler target would help a lot of people staying                                on recent fitnesse releases.

gruenewald:
+1

benhamidene:
I support the suggestion of @amolenaar in #872.

slu-it:
+1

mahamoti666:
+1

dbe-it:
+1

andifalk:
+1

annahitaroehrig:
+1

benhamidene:
Hi folks,

a question: what is the value of restricting fitnesse to Java 8 ?

Is there more than being able to use new cool features of Java 8 ?

Thr price/unvalue would be to exclude all those who are not able to move fast towards the new world. Is it worth it?

As I said: I believe FitNesse deserves major refreshs / refactorings even redesign an that would be great place to start with Java 8  ;)

amolenaar:
Hi,

Good question. My basic idea is that we have to move the project forward. Switching to a newer JDK is one way of doing it. One of our dependencies (org.json) moved to Java 7, so it's about time we also make a step forward. The replies so far seem to be in favour of a more gradual upgrade path.

I planned on using streams and delegate functions, but we can postpone that for a year or so :)

PR #884, opened 2016-03-07
Programmatically add alias & default table type

LudovicVa:
Pull request associated to issue 882.

### DIFF
--- a/src/fitnesse/testsystems/slim/tables/SlimTableFactory.java
+++ b/src/fitnesse/testsystems/slim/tables/SlimTableFactory.java
@@ -163,6 +163,10 @@ public class SlimTableFactory {
       tableTypeArrays.put(fixture, makeTableType(tableSpecifier));
     }
   }
+  
+  public void addDefaultTableType(String fixture, String tableType) {
+      tableTypeArrays.put(fixture, tableType);
+  }
 
   private String makeTableType(String tableSpecifier) {
     tableSpecifier = tableSpecifier.replace(':', ' ');
--- a/src/fitnesse/testsystems/slim/tables/SlimTableFactory.java
+++ b/src/fitnesse/testsystems/slim/tables/SlimTableFactory.java
@@ -165,7 +165,11 @@ public class SlimTableFactory {
   }
   
   public void addDefaultTableType(String fixture, String tableType) {
-      tableTypeArrays.put(fixture, tableType);
+	 tableTypeArrays.put(fixture, tableType);
+  }
+
+  public void addAlias(String alias, String fixture) {
+	  aliasArrays.put(alias, fixture);
   }
 
   private String makeTableType(String tableSpecifier) {


six42:
Hi @LudovicVa,

thanks for the PR.

can you explain how you plan to use these new functions?

Can you give some realistic sample code?

You can't call it from your fixture code. From where do you call them?

LudovicVa:
Hi @six42,

Those methods would be useful in a plugin (in the "registerSlimTables" method).
In my case, I made a plugin that, among other things, 'redirects' one fixture to another (like an alias but with also with some method (input/output) redirection (with conversion or not)). To do so, I use a DDT to call a get and a set on a _generic_ fixture that is created using a custom Fixture Interation (in slim) that then invoke the _real_ fixture method. Those "redirections" are read from a config file.

Thus, it would be pratical to have my plugin being able to read this file as well and set the default table type to DDT to these "fixtures" (as I use the same project for both the FixtureInteraction AND the plugin, it makes sense). It is not very elegant as it requires a Slim and Fit setting but at least it is not necessary to repeat these "fixtures" name in the set up of each test (via define table type). And in any case, to make it work you need settings for both FiT and Slim (via the Fixtures Interaction and by calling with the correct table type).

Hope it is clear enough :)

amolenaar:
I can see why this is useful to do programmatically, next to the table definition.

Can you create an acceptance test that will be using this feature? You can add test fixtures for `fitnesse.slim.test`. That way this functionality will be documented and covered (so I don't rename or delete it by accident :) ).

`FitNesse.SuiteAcceptanceTests.SuiteSlimTests.SetUp` contains an example of how to configure FitNesse with extra plugins.

six42:
Hi @LudovicVa,

Wow, redirection sounds powerful.
But I am still a bit lost with your description. Especially when you mention you are mixing Slim and Fit.
As you mention yourself this is not very elegant I would like to better understand how it will work. Maybe there is an elegant solution :)

Do you plan to publish your plugin so that other could use it as well? Is the source on github?

Why is it better to have the redirection instruction in a config file than on a wiki page? Putting it in a SuiteSetup page would avoid that you have to repeat it on every page as well.

Could you share a wiki page which uses your feature?
This would help to fully understand the problem you are solving.

LudovicVa:
Hi !
I will try to take some time to write the tests over the week.

Unfortunately, I develop the plugin for my company so I wont be able to share it "as this". However the principle is simple emough to recode it quickly.
Basically, the plugin works like this: there is a xml file that looks like this:
`<fixtures>
   <fixture name="Alias" renames="oldFixture">
      <redirect name="add" to="addToList" transform="conversion"/>
   </fixture>
</fixtures>`
If you now call the "alias" _fixtures_, it will call the fixture called "oldFixture". And if you call "add" it will apply the conversion to the input (or the result, depending on how you call it). The conversion uses the same kind of interface as a "Converter<?>' for datatypes.

Thus, I just want to avoid to have a xml config file AND some definition at the beginning of each tests (as we are a team of approx 50 users, it may be complicated to maintain both files, which are not even in the control version tree).

Thanks

amolenaar:
Thank you, @LudovicVa. Please keep the (acceptance) test simple, as it should run from a standalone fitnesse installation without extra dependencies.

PR #891, opened 2016-03-17
Change json content type from text/json to applicatioin/json

nitrok:
At this moment all json data has incorrect content type, text/json. 
The MIME media type for JSON text is application/json. Source: http://www.ietf.org/rfc/rfc4627.txt


amolenaar:
Thanks!

PR #896, opened 2016-03-19
Remove unnecessary code from SlimTestSystem

akruijff:
A list was unnecessary duplicated. There are no side effects to this
change, because the list is never altered.


amolenaar:
Thanks. Nice bit of cleanup :).

PR #907, opened 2016-04-12
SystemUnderTest now accepts method calls from subclasses.

GJatKvK:
Dear FitNesse team,

I've used @SystemUnderTest a while ago. This enables to construct a complex object model in a readable manner. While dusting off this project and use this again with the current version of fitnesse I discovered @SystemUnderTest only accepts calls to the declared class. I use methods from the subclasses a lot. I consider this the powert of SystemUnderTest.

This small addition enables this again.


amolenaar:
Nice addition! To be able to call methods on subtypes of the one used to assign to SystemUnderTest sounds valid.

GJatKvK:
Thanx.

Issue #908, opened 2016-04-14
JUnitRunNotifierResultsListener should fireTestFinished even when fireTestFailure is called.


sgo:
I've been having issues using the FitNesseRunner with the Gradle build tool as it was simply ignoring any test failures reported by the JUnitRunNotifierResultsListener.

At first I thought this was a bug with Gradle but they claim to support a strict interpretation of the RunNotifier contract meaning you should call fireTestFinished after every fireTestFailure.
See: https://discuss.gradle.org/t/with-custom-junit-runner-failures-are-ignored/2496

After I read this I modified JUnitRunNotifierResultsListener locally and sure enough the test failures are now being picked up by Gradle.

I'd submit a patch but I noticed the JUnitRunNotifierResultsListenerTest is a bit sparse and I'm still having trouble getting IvyIDEA to work properly so I'll try again later.

amolenaar:
#909 merged

PR #909, opened 2016-04-23
JUnit run notifier will fire test finished as well after test failure…

sgo:
I created a patch for issue #908 to help out.

In summary the FitNesseRunner doesn't work properly with the Gradle build tool. Reason for this is they expect an invocation of fireTestFinished for every fireTestStarted even if you already invoked fireTestFailure.
They do this as suggested by the Javadoc of RunNotifier.fireTestFinished (http://junit.sourceforge.net/javadoc/org/junit/runner/notification/RunNotifier.html#fireTestFinished%28org.junit.runner.Description%29)

The effect is that any failures or errors observed in a fitnesse run are ignored by Gradle's test result listener.

I added tests to JUnitRunNotifierResultsListenerTest to cover this behavior.

Finally, I'd like to bring your attention to the fact I modified the 2 existing tests slightly to illustrate the summaries passed into testComplete don't matter. They only matter when the test result succeeds and the summary still insists there were failures or errors of some kind.
I don't know if this use case makes sense so I made sure to cover both approaches when adding my own tests.

### DIFF
--- a/src/fitnesse/junit/JUnitRunNotifierResultsListener.java
+++ b/src/fitnesse/junit/JUnitRunNotifierResultsListener.java
@@ -51,13 +51,20 @@ public class JUnitRunNotifierResultsListener
     increaseCompletedTests();
     if (firstFailure != null) {
       notifier.fireTestFailure(new Failure(descriptionFor(test), firstFailure));
+      fireTestFinishedFor(test);
     } else if (testSummary.getExceptions() > 0) {
       notifier.fireTestFailure(new Failure(descriptionFor(test), new Exception("Exception occurred on page " + test.getFullPath())));
+      fireTestFinishedFor(test);
     } else if (testSummary.getWrong() > 0) {
       notifier.fireTestFailure(new Failure(descriptionFor(test), new AssertionError("Test failures occurred on page " + test.getFullPath())));
+      fireTestFinishedFor(test);
     } else {
-      notifier.fireTestFinished(descriptionFor(test));
+      fireTestFinishedFor(test);
     }
+  }
+
+  private void fireTestFinishedFor(TestPage test) {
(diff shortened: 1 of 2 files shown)


src/fitnesse/junit/JUnitRunNotifierResultsListener.java
```diff
@@ -51,15 +51,22 @@ public void testComplete(TestPage test, TestSummary testSummary) {
...
     } else if (testSummary.getExceptions() > 0) {
       notifier.fireTestFailure(new Failure(descriptionFor(test), new Exception("Exception occurred on page " + test.getFullPath())));
+      fireTestFinishedFor(test);
     } else if (testSummary.getWrong() > 0) {
       notifier.fireTestFailure(new Failure(descriptionFor(test), new AssertionError("Test failures occurred on page " + test.getFullPath())));
+      fireTestFinishedFor(test);
     } else {
-      notifier.fireTestFinished(descriptionFor(test));
+      fireTestFinishedFor(test);
     }
```
amolenaar: We could do one single fireTestFinishedFor() after the if statement I suppose. The docs say a fireTestFinished event should be executed for every fireTestStarted event.

amolenaar:
Thanks for the PR. I have one small remark, apart from that it looks nice.

src/fitnesse/junit/JUnitRunNotifierResultsListener.java
sgo: You're right. I remember thinking that and then somehow concluded it'd have more impact than I wanted but there should be no issue to do so.
I'll update the PR when I get home tonight.

sgo:
I updated the PR. If there's anything else I'll get to it when I find some time.

amolenaar:
Awesome!

PR #911, opened 2016-05-06
Fixture interaction received method as a String rather than "method" object directly

LudovicVa:
While developing a Fixture interaction, I don't understand why the "Fixture Interaction" interface is not able to find the actual methods itself, rather than being given as an argument.
It is not symetrical with the way it can create new instances (as classname is given as a string, rather than a "class" object directly).

In my particular case, it will make my Fixture interaction more elegant (as I need the ability to basically "rename" methods. For now, I use a FixtureProxy which wraps a decision table into a dynamic decision table).

I also added a "cached interaction" class which I believe, can improve performance in some cases (as reflexion is not the fatest thing in Java). It simply caches method, class, and constructors.


LudovicVa:
I just realized it makes the "InteractionAwareFixture" a bit odd. I'm not sure where we should put it.
I wonder if it really needs to take "FixtureInteraction" as an argument. IMO, "InteractionAwareFixture" is sort of a lightweight "FixtureInteraction", thus it doesn't need to call FixtureInteraction to invoke the method.
Or we can keep the old interface of "methodInvoke" in the "FixtureInteraction" interface and explain that it is required only for "InteractionAwareFixture" compatibility ??

src/fitnesse/slim/SlimExecutionContext.java
```diff
@@ -53,7 +53,7 @@ public void create(String instanceName, String className, Object[] args)
 
     public void addPath(String path) {
         if (!paths.contains(path)) {
-            paths.add(path);
+            paths.add(0, path);
```
amolenaar: Why prepend instead of append?

src/fitnesse/slim/test/TestSlimInvocationAware.java
```diff
@@ -4,16 +4,17 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import fitnesse.slim.fixtureInteraction.DefaultInteraction;
```
amolenaar: This import is not used. I think it can be removed

amolenaar:
It seems okay to move the method resolution out of the "fixed" executor logic into the Interaction, so we can change the way methods are invoked.

I saw a few things on which I made a remark. Nothing special. I think it would be good to keep the old interface around for a while (with a `@Deprecated` annotation), so old code will not break instantly. We do not have a common base class, do we?

I miss some tests for the `CachedInteraction` class.

@fhoeben Can you have a look at this?

src/fitnesse/slim/SlimExecutionContext.java
LudovicVa: Previously, we append the path at the end of the list. However, each time we looked through the paths imported to find a class, the list was reversed. I removed the piece of code with was a little CPU consuming for nothing if we simply add at the head of the list.

LudovicVa:
I agree that it would be a shame to directly remove the old interface. I don't know how you want to manage having both at the same time. Maybe we can have the old one "wrapped" into the new one, if declared in the config file or via a plugin ?

src/fitnesse/slim/fixtureInteraction/DefaultInteraction.java
```diff
@@ -3,88 +3,144 @@
...
+		Type[] argumentParameterTypes = method.getGenericParameterTypes();
+		return ConverterSupport.convertArgs(args, argumentParameterTypes);
+	}
+
+	protected Object callMethod(Object instance, Method method, Object[] convertedArgs) throws Throwable {
+		try {
+			Object result;
+			if (instance instanceof InteractionAwareFixture) {
+				// invoke via interaction, so it can also do its thing on the aroundMethod invocation
+				result = ((InteractionAwareFixture)instance).aroundSlimInvoke(this, method, convertedArgs);
```
fhoeben: `aroundSlimInvoke()` is no longer called via the interaction. Either remove the comment, or (my preference) restore the behavior to call `aroundSlimInvoke` via the interaction (this would for instance allow a logging interaction to log the fact that `aroundSlimInvoke` was called).

src/fitnesse/slim/fixtureInteraction/CachedInteraction.java
```diff
@@ -0,0 +1,85 @@
...
+			return nArgs * 31 + method.hashCode() + 31 * k.hashCode();
+		}
+
+		@Override
+		public boolean equals(Object o) {
+			if(!(o instanceof MethodKey)) return false;
+			MethodKey m = (MethodKey) o;
+			if(m.k != k) return false;
+			if(m.nArgs != nArgs) return false;
+			if(!m.method.equals(method)) return false;
```
fhoeben: `return m.method.equals(method);`, instead of this line and next?

src/fitnesse/slim/fixtureInteraction/CachedInteraction.java
```diff
@@ -0,0 +1,85 @@
...
+		if(k == null) {
+			classCache.put(className, NotExisting.class);
+		} else {
+			classCache.put(className, k);
+		}
+		return k;
+	}
+
+	
+	private static class MethodKey {
```
fhoeben: make all fields final?

src/fitnesse/slim/fixtureInteraction/CachedInteraction.java
```diff
@@ -0,0 +1,85 @@
+package fitnesse.slim.fixtureInteraction;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
+
+public class CachedInteraction extends DefaultInteraction {
```
fhoeben: I believe all (static) fields could be made final, correct?

src/fitnesse/slim/fixtureInteraction/DefaultInteraction.java
```diff
@@ -3,88 +3,144 @@
...
+	@Override
+	public MethodExecutionResult findAndInvoke(String methodName, Object instance, Object... args) throws Throwable {
+		Method method = findMatchingMethod(methodName, instance.getClass(), args.length);
+		if (method != null) {
+			return this.invokeMethod(instance, method, args);
+		}
+		return MethodExecutionResult.noMethod(methodName, instance.getClass(), args.length);
+	}
+	
+	protected Method findMatchingMethod(String methodName, Class<?> k, int nArgs) {
```
fhoeben: Since the signature is being changed, why not supply all args (or at least their types) to this method? 
If an implementation chooses to use only the number of arguments: fine, but that would allow implementing subclasses the option to actually look more closely.

fhoeben:
Just a bit of nitpicking: why are not all `if`'s followed by a single space? 
I also saw some `else` clauses missing a space after `}`, or before, a `{`...

fhoeben:
Some more remarks/questions?
- The 'old interface' referred to above, is that `InteractionAwareFixture`? I hope that will not be deprecated I use this a lot to have my fixtures do aspect orient behavior specific to them (which I believe is different to the general concept of Slim interacting with all fixtures).
- Should the CachedInteraction become the new 'default' interaction used by Slim? How much performance benefit does it offer, for instance on FitNesse's own tests?

src/fitnesse/slim/fixtureInteraction/DefaultInteraction.java
LudovicVa: Indeed, that makes sense. However, findMatchingMethod (if you speak about this method) is not part of the interface, so other implementation can look more closely to method arguments.

src/fitnesse/slim/fixtureInteraction/CachedInteraction.java
LudovicVa: Yes, indeed, there have to stay constant

src/fitnesse/slim/fixtureInteraction/CachedInteraction.java
LudovicVa: Yes, we can

src/fitnesse/slim/fixtureInteraction/CachedInteraction.java
LudovicVa: Yep, definitely

src/fitnesse/slim/fixtureInteraction/DefaultInteraction.java
LudovicVa: I am not sure understand your point. This piece of code belongs to the fixture interaction implementation, doesn't it ?

LudovicVa:
Regarding the formatting, it's because I'm a lazy coder ^^ And as I switch a lot between languages these days, it's hard to set up my mind to a particular coding style.

The old interface is (in my mind, maybe I misunderstood), the old FixtureInteraction. I don't believe it is desirable to remove the _InteractionAwareFixture_ interface.

In Fitnesse own test, the performance diff is negligible. As it is an optimization, it don't think it is good to use it by default (at least for now), as it really depends on how your tests/fixture are written. In my particular case at work, I know (using measuring tool) that reflection is quite heavy. But I haven't found the time to measure the real benefits of caching.

src/fitnesse/slim/fixtureInteraction/DefaultInteraction.java
fhoeben: Previously this code explicitly called the method using the interaction (and not directly as a method call, and that was what the comment refers to, so now it no longer makes any sense: aroundSlimInvoke is NOT called via the interaction, but normally/directly). 
Previously the code making this call was not part of the interaction, so there was a bit more delegation going on. The code making the call could not rely on what the interaction would do exactly. Now the call is done from inside the interaction, so that uncertainty no longer exists. So maybe just removing the comment is a fine solution.

But now I wonder whether there will be an infinite loop with `InteractionAwareFixture.aroundSlimInvoke()`. Previously it would not do the actual method call itself, but delegate that to the interaction again (that's why it's passed as first argument, and that's also the recommended approach in `InteractionAwareFixture.aroundSlimInvoke`'s documentation). Now in the new setup the interaction calls `aroundSlimInvoke` so the fixture calling the interaction seems like a recipe for an infinite loop.

I'll dig a bit deeper what the current status is and what can be done about this.

LudovicVa: Indeed, infinite loop can be easily generated...
Maybe, the cleanest is to call `aroundSlimInvoke` outside the _FixtureInteraction_. However, this would mean having the `InteractionAwareFixture` interface to change as well to be completely consistent, doesn't it ?

fhoeben: Actually the infinite loop does not occur. The fixture is supposed to delegate to `interaction.methodInvoke()`, and that does not do the interactionAwareFixture handling. 
So my fears were unfounded: that problem does not exist.

LudovicVa: Indeed, but the interface is in my opinion, still a little by confusing (as an "good" implementation should call `methodInvoke` in its implementation of `findAndInvoke`, nothing preventing it from doing differently. Maybe we actually should split into two methods `findMethod` and `invokeMethod`?
Thus, the link between the two function is "hard coded" (maybe losing some possibilities in the process ??), avoiding all this headache ?

fhoeben: In the current release the fixtureInteraction only does the invokeMethod, with 'find' living in the `MethodExecutor`. I believe that's what you wanted to address  (and I think is a good idea) :-)

I believe the current setup will work, and although less clear than I would ideally want it's not too bad (and adding an extra layer to separate but keep together the find and invoke parts creates new complexity also). And this gives people the flexibility to do what they need.

I would still prefer it the calling of 'aroundSlimInvoke()' would use `methodInvoke()`, as it did before: to ensure that any overriding of `methodInvoke` also sees/captures the call to `aroundSlimInvoke`.

LudovicVa: > I would still prefer it the calling of 'aroundSlimInvoke()' would use methodInvoke(), as it did before: to ensure that any overriding of methodInvoke also sees/captures the call to aroundSlimInvoke.

I'm not quite sure of what you meant. So suggested implementation is OK for you ?

fhoeben: Not quite, I would like to see (like previously in `MethodExecutor.callMethod()`:
```
			if (instance instanceof InteractionAwareFixture) {
				// invoke via interaction, so it can also do its thing on the aroundMethod invocation
          Object[] args = { this, method, convertedArgs };
          result = methodInvoke(AROUND_METHOD, instance, args);
			} else {
				result = methodInvoke(method, instance, convertedArgs);
			}
```

LudovicVa: Ok got it ! It makes way more sense that way. Fixing it right now :)

Issue #910, opened 2016-04-25
Change build system to Gradle


amolenaar:
I'd like to change the build system to Gradle. Ant served us well over the years, but Gradle provides some advantages over Ant nowadays.

By using Gradle we can:

* Create a multi-module build, so Fit, Slim and the web server can be built separately.
* Slim can be built against Android libs, to ensure it works.
* Upload to Bintray is easier. Bintray is easier to upload to than Maven Central, this can ease the release process somewhat. Maybe move away from Cloudbees/Jenkins.
* The Gradle wrapper script will prevent the need for separate installation of a build tool.
* The code used to create the updateList and doNotCopyOverList can be moved to a plugin.
* Incremental builds are faster.
* Better integration with IDE's. IntelliJ has proper support for Gradle, and I think Eclipse has the same, alternatively project files can be generated from Gradle.

Am I missing out on something? Would it be better to stick with Ant?

six42:
Hi Arjan,

Good idea!

My experience with Gradle was positive.

amolenaar:
Fixed by #916.

PR #916, opened 2016-05-15
Chaneg build system to Gradle

amolenaar:
See #910.

This PR does a 1:1 replacement of build.xml for build.gradle.

Deployments are done to Bintray, this should make the release process simpler.

We should change the build job on Jenkins or use Travis CI instead (I prefer the second).


amolenaar:
@woodybrood @unclebob I'd love to change the build system to Gradle for reasons mentioned in #910. I suppose this will require some change in the Jenkins build (which I can not modify).

I'm also fine with moving to Travis-CI + Bintray, which works very well :).

amolenaar:
Let's add the PITest stuff first.

Issue #834, opened 2015-11-21
Output stream closed unexpectedly (Stop button pressed?) [Software caused connection abort: socket write error]


nikoudel:
Getting lots of those messages when navigating between pages (eg. simply clicking http://localhost/FitNesse and back). Version v20150814, fresh install on Windows 7.

```
PS C:\temp> java -jar .\fitnesse-standalone.jar -v
marraskuuta 21, 2015 2:15:32 IP. fitnesse.ConfigurationParameter loadProperties
INFO: No configuration file found (C:\temp\plugins.properties)
Configured verbose logging
Bootstrapping FitNesse, the fully integrated standalone wiki and acceptance testing framework.
root page: fitnesse.wiki.fs.FileSystemPage at .\FitNesseRoot#latest
logger: none
authenticator: fitnesse.authentication.PromiscuousAuthenticator
page factory: fitnesse.html.template.PageFactory
page theme: bootstrap
Starting FitNesse on port: 80
Creating plain socket on port: 80
Socket class: class java.net.Socket
   Remote address = /0:0:0:0:0:0:0:1:52871
   Local socket address = /0:0:0:0:0:0:0:1:80
   Closed = false
   Connected = true
   Bound = true
   isInputShutdown = false
   isOutputShutdown = false
Output stream closed unexpectedly (Stop button pressed?) [Software caused connection abort: socket write error]
Output stream closed unexpectedly (Stop button pressed?) [Software caused connection abort: socket write error]
Output stream closed unexpectedly (Stop button pressed?) [Software caused connection abort: socket write error]
Output stream closed unexpectedly (Stop button pressed?) [Software caused connection abort: socket write error]
Output stream closed unexpectedly (Stop button pressed?) [Software caused connection abort: socket write error]
Output stream closed unexpectedly (Stop button pressed?) [Software caused connection abort: socket write error]
Output stream closed unexpectedly (Stop button pressed?) [Software caused connection abort: socket write error]
```

MaximeArnstamm:
I have the same problem on a fresh install, i tried two versions v20150814 and v20140201
I run fitnesse locally with fitnesse-launcher-maven-plugin

In my console i have these messages, always by 4 : 
déc. 21, 2015 11:07:13 AM fitnesse.FitNesseExpediter send
INFOS: Output stream closed unexpectedly (Stop button pressed?)
java.net.SocketException: Software caused connection abort: socket write error
	at java.net.SocketOutputStream.socketWrite0(Native Method)
[...]

PR #923, opened 2016-05-19
Propagate exceptions

amolenaar:
In the code base exceptions are logged (and not propagated) on several occasions or RuntimeExceptions are thrown

This PR cleans up the exceptions, propagate them where possible.

This results in many IOExceptions to be propagated. Is this the right approach?


fhoeben:
Instead of propagating the IOExceptions as-is, would it be possible to wrap them in a (custom) runtime exception. There's little point in having the exceptions explicit all over your codebase if your just going to propagate them anyway. 
RuntimeExceptions allow only the code that actually deals with the exception to have to worry about them. The rest of the code base does not need to be 'polluted' by throws clauses.

amolenaar:
Well, quite a lot of the methods that throw IOException now are actually dealing with I/O. I ran into trouble with WikiPage, which should not be tied to IOException. I considered creating a custom exception. I try to keep the exceptions as explicit as possible (not hiding stuff with runtime exceptions, see also http://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html).

I suppose we need a custom exception hierarchy then.

fhoeben:
Like the page you refer to states there are different views on this topic. But I fully agree with the last paragraph:
> Here's the bottom line guideline: If a client can reasonably be expected to recover from an exception, make it a checked exception. If a client cannot do anything to recover from the exception, make it an unchecked exception.

I believe in many places in the current FitNesse codebase there is nothing to be done about an IOException, they will just propagate. So an unchecked (i.e. Runtime) exception is the way to go according to me.

src/fitnesse/FitNesseExpediter.java
```diff
@@ -69,13 +69,9 @@ public void run() {
   }
 
   @Override
-  public void send(byte[] bytes) {
-    try {
-      output.write(bytes);
-      output.flush();
-    } catch (IOException e) {
-      LOG.log(Level.FINE, format("Could not send data for URL %s: %s (Stop button pressed?)", (request != null ? request.getResource() : ""), e.toString()));
```
fhoeben: I see this message in my console quite often. I believe not only when I pressed stop, or navigated away from the page, but I have not been able to determine when it happens. 
Stopping a test when this occurs seems like a 'risky' change...

amolenaar: There is already an issue for this: #834. I expect it has something to do with browsers (most notably Chrome) prefetching data.

src/fitnesse/reporting/history/ExecutionReport.java
```diff
@@ -66,24 +69,35 @@ public int hashCode() {
...
   }
 
-  public static ExecutionReport makeReport(String xmlString) throws Exception {
-    Document xmlDocument = XmlUtil.newDocument(xmlString);
+  public static ExecutionReport makeReport(String xmlString) throws InvalidReportException {
+    Document xmlDocument = null;
+    try {
+      xmlDocument = XmlUtil.newDocument(xmlString);
+    } catch (IOException | SAXException e) {
+      throw new InvalidReportException("%s is not a valid execution report", e);
```
fhoeben: Should the %s be replaced with something (the xmlString for instance)?

src/fitnesse/reporting/history/ExecutionReport.java
```diff
@@ -66,24 +69,35 @@ public int hashCode() {
...
+  protected void unpackCommonFields(Element documentElement) throws InvalidReportException {
     version = XmlUtil.getTextValue(documentElement, "FitNesseVersion");
     rootPath = XmlUtil.getTextValue(documentElement, "rootPath");
     String dateString = XmlUtil.getTextValue(documentElement, "date");
     if (dateString != null)
-      date = DateTimeUtil.getDateFromString(dateString);
+      try {
+        date = DateTimeUtil.getDateFromString(dateString);
+      } catch (ParseException e) {
+        throw new InvalidReportException(String.format("'%s' is not a valid date.", dateString), e);
```
fhoeben: If you will be replacing %s more often in the message of exceptions of this type, maybe the String.format() can be part of (an extra) constructor for the exception?

src/fitnesse/reporting/history/PageHistoryReader.java
```diff
@@ -29,19 +29,24 @@ private boolean fileIsNotADirectoryAndIsValid(File file) {
...
-  private TestResultRecord buildTestResultRecord(File file) throws ParseException {
+
+  private TestResultRecord buildTestResultRecord(File file) {
     String[] parts = file.getName().split("_|\\.");
-    Date date = dateFormat.parse(parts[0]);
+    Date date;
+    try {
+      date = dateFormat.parse(parts[0]);
+    } catch (ParseException e) {
+      throw new IllegalStateException("File format should have been verified", e);
```
fhoeben: "File format should have been verified" does not seem like the most descriptive message when date parsing fails....

src/fitnesse/socketservice/SocketService.java
```diff
@@ -71,7 +71,7 @@ private void serviceThread() {
       } catch (SocketException sox) {
         running = false; // do nothing
       } catch (IOException e) {
-        throw new RuntimeException(e);
+        LOG.log(Level.SEVERE, "I/O exception in service thread", e);
```
fhoeben: This should not be propagated? This seems contrary to the purpose of the pull request?
Maybe a comment why this may be logged but not propagated is in place.

src/fitnesse/testsystems/fit/FitTestSystem.java
```diff
@@ -51,18 +51,20 @@ public void runTests(TestPage pageToTest) throws IOException, InterruptedExcepti
         client.send(html);
     } catch (InterruptedException e) {
       exceptionOccurred(e);
-      throw e;
+      throw new TestExecutionException("Testing has been interrupted", e);
```
fhoeben: No Thread.currentThread.interrupt()?

src/fitnesse/testsystems/fit/FitTestSystem.java
```diff
@@ -51,18 +51,20 @@ public void runTests(TestPage pageToTest) throws IOException, InterruptedExcepti
...
   }
 
   @Override
-  public void bye() throws IOException, InterruptedException {
+  public void bye() throws UnableToStopException {
     try {
       client.done();
       client.join();
+    } catch (InterruptedException | IOException e) {
+      throw new UnableToStopException("Unable to stop Fit client", e);
```
fhoeben: No Thread.currentThread.interrupt() on InterruptedException?

src/fitnesse/testsystems/slim/tables/SlimTable.java
```diff
@@ -122,9 +123,9 @@ protected SlimAssertion constructFixture(String fixtureName) {
   public void setFixtureName(String name){
 	  fixtureName = name;
   }
-  
+
   protected String getFixtureName() {
-	if (fixtureName == null){  
+	if (fixtureName == null){
```
fhoeben: If you're fixing whitespace: space before '{'?

src/fitnesse/testsystems/slim/tables/SlimTable.java
```diff
@@ -430,18 +431,18 @@ public TestResult evaluateExpectation(Object returnValue) {
         String value = getSymbol(this.symbolName);
 	      return super.evaluateExpectation(value);
 	    }
-	    
+
 	    @Override
 	    protected SlimTestResult createEvaluationMessage(String actual, String expected) {
-	      if (assignToName != null){	
+	      if (assignToName != null){
```
fhoeben: Again, space before '{'?

src/fitnesse/socketservice/SocketService.java
amolenaar: This is the top level in the thread. Propagating at this level would cause the server to stop.

PR #928, opened 2016-05-27
Fix assignation of output parameters of scenarios with graceful names in decision tables

fleger:
The name of the output parameter of a scenario should be disgraced when assigning the parameter in a decision table.


amolenaar:
Thanks for the fix!

PR #929, opened 2016-06-01
Add ability to specify the locale used to format evaluated expressions

fleger:
This patch allows the user to specify which locale to use when formatting the result of a ${= =} expression.
The locale can be specified by setting the FORMAT_LOCALE variable, i. e. !define FORMAT_LOCALE {fr} will make the formatting use the French locale regardless of the JVM default locale.
If the FORMAT_LOCALE variable is not defined, the default JVM locale is used.


amolenaar:
I like this patch. I'm wondering if there are more places where localization might be desired. 

Is the locale setting also desired when parsing expressions? In that case the variable can be named LOCALE.

amolenaar:
I think  it's fine to merge now and extend localization later on.

PR #939, opened 2016-06-19
Slim: run TearDown after StopText exception and SuiteTeardown after StopSuite

### LINKED ISSUE #684: Execute TearDown on StopException
Currently the TearDown for a Slim tests that throws a StopTest exception is not executed. This is not very intuitive (since different from TearDown handling in other (e.g. unit) test environments), which also show from the following quote from  FitNesse.UserGuide.WritingAcceptanceTests.SliM.ExceptionHandling: 
> Be warned, that if you throw a "StopTest" exception then the tear down for your test will not run. A fix for this is to run the tear down procedure before you throw the exception.

The fix suggested in the quote does not work for me as the test fixture throwing the exception is not aware of the content of the TearDown page.

- I believe the content of the TearDown page should be executed irrespective of whether the test was stopped or not. 
- On a SuiteStop exception not executing the TearDown is ok, but I would expect the SuitTearDown to be executed.

mgurov:
closes #684 for slim testing system. 

Assumes fit system doesn't need this change as stop test/suite exceptions are slim-specific. 

For TearDown pages we marked them as such on inclusion during the wiki markup parsing. This flag is applied to the tables built from such page and carried all the way to the execution point at `fitnesse.testsystems.slim.SlimTestSystem#processTable`. 

SuiteTearDown is executed as a separate page by slim, so we can recognize it by the name, following the practices of `fitnesse.testrunner.WikiTestPage#isSuiteSetUpOrTearDownPage` and alike.

### DIFF
--- a/src/fitnesse/util/HtmlParserTools.java
+++ b/src/fitnesse/util/HtmlParserTools.java
@@ -102,7 +102,7 @@ public final class HtmlParserTools {
     return newAttributes;
   }
 
-  public static boolean nodeHasClass(String classToCheck, Node node) {
+  public static boolean nodeHasClass(Node node, String classToCheck) {
     if (!(node instanceof TagNode)) {
       return false;
     }
--- a/src/fitnesse/util/HtmlParserTools.java
+++ b/src/fitnesse/util/HtmlParserTools.java
@@ -5,6 +5,7 @@ import java.util.Vector;
 import org.htmlparser.Attribute;
 import org.htmlparser.Node;
 import org.htmlparser.Tag;
+import org.htmlparser.nodes.TagNode;
 import org.htmlparser.util.NodeList;
 
 /**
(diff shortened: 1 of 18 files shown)


mgurov:
Dear maintainers, could you please hint on whether the change proposed is sensible in principle? Thank you!

src/fitnesse/util/HtmlParserTools.java
```diff
@@ -101,4 +102,20 @@ private static Vector cloneAttributes(Vector<Attribute> attributes) {
     return newAttributes;
   }
 
+  public static boolean nodeHasClass(String classToCheck, Node node) {
```
amolenaar: I would put Node as first argument, and the class second.

amolenaar: I would put Node as first argument, and the class second.

amolenaar:
I missed this PR. It looks really nice (one minor comment). The solution you propose is surprisingly simple.

I think the changes to `SlimTestSystem` need some unit tests as well.

mgurov:
@amolenaar thank you for the feedback. The minor comment has been addressed.

Unit-testing the `SlimTestSystem` is a bit more complicated story. I will try to come up with something reasonable without much changes to the class under test. Meanwhile the tests at `FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests`  (`StopSuite`/`StopTest`) have full coverage of the changes, albeit with a somewhat slower feedback loop.

mgurov:
@amolenaar added some unit test coverage to the affected part of the `SlimTestSystem`.

mgurov:
Dear maintainers, please let me know if there is anything I could further improve in this pull request and facilitate its acceptance.

amolenaar:
Hi,

I need to take some time to merge it. It's a nice PR.

PR #940, opened 2016-06-20
Added Slim Hybrid Decision Table

six42:
The Hybrid Decision Table is a combination of the Decision Table and the Dynamic Decision Table.
A decision table requires for each table column the definition of a method.
This is impractical when you deal with structured data that has a varying size like: hash maps, XML, properties, database tables, etc.
The number of methods you need to write could be infinite.

A dynamic decision table has just one method which is called for all columns and your fixture has to do the dispatching.
This solves the above problem but dispatching should be done by the test system and not by your fixture.

Sometimes you want both features in one table and that is when you use the Hybrid Decision table.

The Hybrid Decision Table allows the methods called for each column to be
redefined and method parameters can be extracted from the column names.

This PR also cleans up the code as the Dynamic Decision Table is now just a special cases of the Decision Table

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/HybridDecisionTable/ImplementDynamicDecisionTableWithHybridTable/content.txt
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/HybridDecisionTable/ImplementDynamicDecisionTableWithHybridTable/content.txt
@@ -1,0 +1,34 @@
+|import            |
+|fitnesse.slim.test|
+
+!3 The Dynamic Decision Table is just a special case of the Hybrid Decision table.
+With the following rules for getter and setter the behavior of Dynamic Decision Tables is achieved.
+!define SLIM_DT_GETTER (!-
+{
+"FormatVersion":"1.0",
+"MethodExtractorRules":[
+  {
+    "Scope":".+",
+    "TargetName":"get",
+    "Parameters":"$0"
+  }
+ ]
+}
+-!)
+!define SLIM_DT_SETTER (!-
+{
+"FormatVersion":"1.0",
+"MethodExtractorRules":[
+  {
+    "Scope":".+",
+    "TargetName":"set",
+    "Parameters":"$0"
+  }
+ ]
+}
+-!)
+|dt: add up change                                                                                                          |
+|# description                           |1c         |5c|10c|25c|50c|$1|total cents?|$ total?                               |
+|some simple addition                    |2          |2 |4  |0  |0  |0 |52          |0.52                                   |
+|save the total cents in a symbol        |56         |0 |0  |0  |1  |20|$totalCents=|21.06                                  |
+|now use the total cents that were stored|$totalCents|0 |0  |0  |0  |10|3106        |~=31.1|An example for Value Comparisons|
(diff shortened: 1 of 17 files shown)


amolenaar:
This is a tough one. I like most of the refactoring you did and the method extractor code. It's definately an improvement separating the method mapping logic from the table itself.

However, it seems to me the hybrid table approach, with the `SLIM_DT_[SG]ETTER` variables, make things too complicated for my taste.

six42:
Hi Arjan,

I had the following constraints when I decided this design:

The method mapping is a technical implementation detail and should not be mixed with the requirements in the decision table itself.

The hybrid table should be supported in all slim servers (java, c#, c, ...) without a need to change them.

It should be possible to use it with libraries for which the source code is not available.

If fixture code is written then ideally the mapping should be defined in the fixture code. 

Also the last item is not yet implemented it decided the choice of JSON as the configuration format.
In a next version the BeginTable or Table function or a newly defined  function of a fixture could return this JSON object to steer the method mapping.

To configure the mapping in a test page I had the option to create a new table type like the 'define alias' table or something simpler.

I decided for the simplest solution to start.
First it is in the light of being agile. Let's see first if there is interest in the hybrid table before creating a fancy user interface.

Second I expect the format might further evolve. It will be easier to support for backward compatibility multiple definitions in JSON format than in a table format.

Creating the JSON is easy as shown in the test page MethodExtractorRuleTest. So nobody has to type JSON by hand.

Last but not least the usage of variables is the standard to configure the test system so it follows the current design.

To make a long story short: I don't think this is the final user interface design but a good start.

amolenaar:
> The method mapping is a technical implementation detail and should not be mixed with the requirements in the decision table itself.

The good of this PR.

> Also the last item is not yet implemented it decided the choice of JSON as the configuration format.
In a next version the BeginTable or Table function or a newly defined  function of a fixture could return this JSON object to steer the method mapping.

I'm not sure if we should go down that route. Then rather use the table function and parse it yourself.

> To configure the mapping in a test page I had the option to create a new table type like the 'define alias' table or something simpler.

I wonder if you want that information in a wiki page: it feels like programming to me.

> Last but not least the usage of variables is the standard to configure the test system so it follows the current design.
> 
> To make a long story short: I don't think this is the final user interface design but a good start.

Agreed. We should make clear from the documentation that the use of the `SLIM_DT_[SG]ETTER` variables is experimental and subject to change.

six42:
Hi @amolenaar 
I added a warning. 
Anything else required to merge?

amolenaar:
This should be enough, right?

amolenaar:
Somehow one test seems to fail (with me at least):

```
FitNesse.SuiteAcceptanceTests.SuiteSlimTests.HybridDecisionTable.MethodExtractorRuleTest

java.lang.Exception
	at fitnesse.junit.JUnitRunNotifierResultsListener.firstFailure(JUnitRunNotifierResultsListener.java:133)
	at fitnesse.junit.JUnitRunNotifierResultsListener.testExceptionOccurred(JUnitRunNotifierResultsListener.java:81)
	at fitnesse.testsystems.CompositeTestSystemListener.testExceptionOccurred(CompositeTestSystemListener.java:61)
	at fitnesse.testrunner.MultipleTestsRunner$InternalTestSystemListener.testExceptionOccurred(MultipleTestsRunner.java:223)

```

six42:
Strange.

Worked without fail under Linux with openJdk but failed under Windows with Oracle JDK.
It looked like the wrong constructor was called.

After removing an unused constructor the test works now.
But your error message is slightly different than what I have seen under Windows.

@amolenaar,  Can you test once more.

amolenaar:
Worked :)

PR #944, opened 2016-06-29
[Docker] add Dockerfile to run Fitnesse

sjeandeaux:


amolenaar:
Can you explain a little what should be the purpose of this addition.

I think it should bundle fitnesse.jat (or standalone jar) and launch that instead of building and running it from the container.

sjeandeaux:
I propose this solution to help anyone to run fitnesse without Java and without git project.
We could publish this image on https://hub.docker.com/.
Everyone will can run this image with `docker run -p 8001:8001 -d fitnesse`.

amolenaar:
I'm definitely in favor of providing a docker image, but not  one that builds stuff from scratch. And the build should be handled by Gradle.

woodybrood:
I think also that in the case of FitNesse running in a docker container, we should be providing some guidance on how to manage the tests themselves in a container. I'm still new to docker, but it seems like the lifecycle of containers is different than those of other environments. Often containers are pointing at databases that exist on a different machine, and therefore are not lost when a container is relaunched. FitNesse tests could be lost if someone doesn't also hook into git or similar and ensure changes get committed and pushed.

README.md
```diff
@@ -103,3 +103,17 @@ There are a few things to keep in mind when working from an IDE:
 3. Select the file `build.gradle` in the fitnesse folder.
 4. Follow the wizard. Use Java 7 or newer. It should find source and test folders.
 5. Open the Gradle Build tool, select the task `copyRuntimeLibs` and (right-click) mark it as _Execute After Make_.
+
+#### Dockerfile
+
+```
+#To create your image
+$ docker build --tag fitnesse .
```
mgurov: At this form you could simply extend commands below to `docker run -d -p 8001:8001 -v ~/.gradle:/root/.gradle -v $(pwd):/fitnesse -w /fitnesse java:7 ./gradlew run` and achieve the same result. Minus the overhead of building a dedicated docker image.

PR #948, opened 2016-07-08
Expose wiki page to jUnit listeners

fhoeben:
When FitNesseRunner executes a suite it would sometimes be convenient for jUnit listeners connected to the run to be able get access to the page context (for instance to access the tags associated with a test). 
Test listeners don't need to know anything about this information, but if they want to access it they can.
An example of when this is nice when creating a specialized listener to connect FitNesse test runs to a dashboard like [Allure](http://allure.qatools.ru). 

jUnit provides a mechanism to attach additional information to a test's context via annotations contained in the test Description.
In this pull request I ensure that the page context available when creating the jUnit Description is stored in an annotation inside the generated Description.

I recently got [a pull request](https://github.com/fhoeben/hsac-fitnesse-fixtures/pull/70) to change the workings of my customized FitNesseRunner to support connections to Allure. The only need for a code change was the need to have page information available for jUnit listeners. With this pull request any runner can make use of this information, without further custom coding.

### DIFF
--- a/src/fitnesse/junit/DescriptionFactory.java
+++ b/src/fitnesse/junit/DescriptionFactory.java
@@ -1,5 +1,7 @@
 package fitnesse.junit;
 
+import java.lang.annotation.Annotation;
+
 import fitnesse.testsystems.TestPage;
 import fitnesse.wiki.WikiPage;
 import org.junit.runner.Description;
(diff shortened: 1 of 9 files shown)


amolenaar:
Thanks! This is a nice addition.

Issue #931, opened 2016-06-02
Single file page format


amolenaar:
FitNesse is using a directory (the page name) and two files: `content.txt` and `properties.xml` to store a wiki page to disk.

While this works fine if you're only using FitNesse from the wiki, this structure has some downsides when used from IDE's. Version control is also somewhat more laborious with those many files.

I'd like to propose a single file format: `<page-name>.wiki` for a single wiki page.
## Nesting of pages

~~I think here the folder should have the page name and it should contain a default file: `content.wiki`.~~

Do as Gregor says.
## Page properties

Page properties (if it's a test page, symlinks, etc.) will be stored in a "front-matter", just like Jekyll does it (https://jekyllrb.com/docs/frontmatter/). So a wiki page, when opened in a text editor, may look like this:

```

---
Edit
Test

---
Page content comes here
```

To do list:
- [x] Use old and new style pages interchangeably
- [x] Parse Front mapper with current wiki parser
- [x] Create page properties from front matter
- [x] Create page data
- [x] Remove pages and sub pages
- [x] Pages should be listed in the contents section
- [x] Store new file versions
- [x] Create (zip) backup files -> support `VersionsController` interface
- [x] Use new page format as default for new pages

ggramlich:
In general, I think this is a good idea.
Do you think of a single migration step or should it be able to read pages in a backwards compatible way and do an ad hoc conversion?

Concerning the **Nesting of pages** part: As you don't know in advance, whether a page might have subpages, the format you are proposing could cause frequent changes from page-file to folder. What happens, when the last subpage is deleted?
I think, a more consistent solution might be:
Normal <page-name>.wiki file, but as soon as a subpage exists, you add a folder <page-name> and leave the <page-name>.wiki file untouched.

```
|- PageA.wiki
|- PageA (folder)
|  |- PageAA.wiki
|  |- PageAA (folder)
|     |- PageAAA.wiki
|  |- PageAB.wiki
|-PageB.wiki
```

amolenaar:
Thanks for your suggestion on the nested pages. Indeed that makes things a lot simpler.

Regarding migration: I thought of making a new WikiPage type (rename the old one to `OldFileSystemPage` or something), and create new pages using this new structure. One reason to change the extension is to make a clear distinction between the old format (`content.txt`) so that both files can coexist next to each other. Just like we do with `ExternalSuitePage`.

I think that doing migration is tougher since we have to mix the old `FileSystemPage` with the new implementation.

[...]

PR #952, opened 2016-07-27
New style wiki pages

amolenaar:
See #931 for rationale and discussion.

This PR introduces a new style for wiki page content.

Of course the old style pages can still be edited, they are not converted or something like that.

Pages can be renamed and moved. In both cases the pages are rewritten in the new file format.

If you really need the old file format, you can set the property `wiki.page.old.style` to `true`.
This will result in all pages being created as a directory with a `content.txt` and `properties.xml`.


woodybrood:
So this is the big one? Is the recommendation that interested parties try it out before you merge, or will you merge and then get comments?

jediwhale:
Nice!

amolenaar:
Please try before merging, or merge it yourself if you're okay with this change. :)

fhoeben:
I would like to have the option to 'bulk rewrite' an entire (sub-)wiki. That would allow me to ensure the entire test set is in one format. 

I would run this once (maybe by passing a special startup argument, e.g. `java -jar finesse-standalone.jar -convertPagesToNewWikiStyle`), and commit it as such in to version management. And know from then one there is only one page format for the entire test set. 

Would this be really hard to add (as you can already do a format change on move/rename)?

amolenaar:
While on the topic, I do now know how (if at all) the user guide can be moved to the new format.

We do not want to change page types by default (since other tooling may depend on this format).

Would it be okay it if were possible from an IDE only?

fhoeben:
IDE only (IntelliJ?) would be ok. But it would be nicer if it were just a call to make.

kollaharsha:
Is there a way to convert all the existing pages into the new wiki page format from IDE (IntelliJ)?

amolenaar:
I have not implemented that. For now the most convenient way is to move your test suite though the wiki.

Johnny2Shoes:
For anyone else just stumbling upon this looking for why you can't get the new style wiki format to work on windows... it's because it requires Linux Style line endings (i.e. lf not crlf).

Add a .gitattributes file with `*.wiki text eol=lf ` to force git to sort this for you (or change your editor settings)

I was seeing this behaviour because of a slightly strange setup we are using but it may help someone else.

Cheers!

PR #956, opened 2016-08-04
make "last change date" visible in the wiki

six42:
As a Wiki User
To understand the flow of changes
I need to see the "last changed date" of a wiki page when comparing versions or when looking at the properties of a single page

currently only the last author can be seen


six42:
Thanks

PR #960, opened 2016-08-15
Fixed one typo in each of README.md and CONTRIBUTING.md

FelixRiemann:
Two typos have been fixed.

I also noticed that CONTRIBUTING.md still talks about ant targets, that may need an update.

Moreover, the section "Configure SLIM" on the wiki page describing SliM (i.e. http://fitnesse.org/FitNesse.UserGuide.WritingAcceptanceTests.SliM or the file FitNesseRoot/FitNesse/UserGuide/WritingAcceptanceTests/SliM/content.txt) does not match completely with what can be found in `fitnesse.slim.SlimService`:
- According to the wiki page, the default value for `slim.port` is 8085, while in the SlimService class it defaults to 8099.
- On the wiki page, the command line argument pattern for the slim service misses the `port` option. In the class source the port parameter is present. Also, as a default value is specified, shouldn't the port parameter simply be made optional?

I'd be happy to correct this as well.


amolenaar:
Nice! Thanks!

Issue #859, opened 2016-01-07
Graceful names not work in some cases


salandur:
I recently upgraded from 20140901 to 20151230 and noticed a problem with the Graceful names. This used to work in the older version.

For Class names, if they are a single word, the capitalization does not work. I.E.
|table:rest|
will give an error: "Could not invoke constructor for rest"

|script|slimRest|
Does not capitalize the first S for a classname.

For methods calls, it also goes wrong if the first letter is capitalizes
|set service|value|
|Set service|value|
The first one works, the second one doesn't (generates the SetService method name)

It is not clear from the [Gracefull Names](http://fitnesse.org/FitNesse.FullReferenceGuide.UserGuide.WritingAcceptanceTests.SliM.GracefulNames) description what the correct one should be.

amolenaar:
I did some checking. To me it seems like the graceful name handling has not changed: Only multi-word names are dealt with, and case is preserved.

```
public class DisgracerTest {

  @Test
  public void disgraceClassName() {
    assertEquals("classname", SlimTable.Disgracer.disgraceClassName("classname"));
    assertEquals("className", SlimTable.Disgracer.disgraceClassName("className"));
    assertEquals("ClassName", SlimTable.Disgracer.disgraceClassName("class name"));
  }

  @Test
  public void disgraceMethodName() {
    assertEquals("methodName", SlimTable.Disgracer.disgraceMethodName("method name"));
    assertEquals("methodname", SlimTable.Disgracer.disgraceMethodName("methodname"));
    assertEquals("MethodName", SlimTable.Disgracer.disgraceMethodName("Method name"));
  }
}
```
(this tests is applicable for 20140901)

salandur:
It is actually an odd case, where the name is a single word. This means that it cannot be graceful, although in case of a classname it doesn't start with a capital. It is a minor issue, but we can add some logic for this to capitalize the first letter in the correct way for both class names and method names

amolenaar:
I am aware of this behavior. I'm cautious about changing this behavior since I have no clue on what the impact will be on existing test suites. I'm not sure though how it will work out with tools like FitSharp, where method names also start with an uppercase character. @jediwhale Any thoughts on this?

jediwhale:
FitSharp does a case-insensitive match so it will not be impacted by any changes in case for class or method names.

amolenaar:
Related to #655, me thinks.

benhamidene:
[...]

PR #964, opened 2016-08-23
Allow Slim to use a custom NameTranslator

FelixRiemann:
- NameTranslator interface now has two methods (class/method specific)
- JavaSlimFactory now needs a NameTranslator for construction.
- Added -nt parameter to SlimService, to specify a custom class name of a NameTranslator for SlimFactory. If none is given, NameTranslatorIdentity is used by default.
- Reworked the way instructions are created by InstructionFactory: The InstructionFactory now needs an instance of a NameTranslator. It translates the method/class names before creating the instructions, so the instructions themselves are now unaware of NameTranslators. The factory methods are not static anymore.
- Updated description of slim.flags on the SliM wiki page

I noticed that different default ports for Slim are specified in fitnesse.testsystems.slim.SlimClientBuilder and fitnesse.slim.SlimService.

See also: #859

### DIFF
--- a/FitNesseRoot/FitNesse/UserGuide/WritingAcceptanceTests/SliM/content.txt
+++ b/FitNesseRoot/FitNesse/UserGuide/WritingAcceptanceTests/SliM/content.txt
@@ -47,7 +47,7 @@ The Slim test system can be configured using the following properties:
 | slim.pool.size | 10 | The size of the pool of ports to cycle through. By default the ports 8085 up to 8095 are used (''slim.port'' + ''slim.pool.size''). |
 | slim.host | localhost | The host the SLIM server will be running on. This is mostly useful if you're running a remote SLIM server. |
 | slim.flags | | Extra flags to provide to the SLIM server. Arguments supported by the slim service are: !-
--![-v] [-i interactionClass] [-s statementTimeout] [-d] [-ssl parameterClass] |
+-![-v] [-i interactionClass] [-nt nameTranslatorClass] [-s statementTimeout] [-d] [-ssl parameterClass] |
 | slim.timeout | 10 seconds | Connection timeout starting and finishing a test run. |
 | slim.debug.timeout | ''slim.timeout'' | Same as ''slim.timeout'', used when the debug property is set (falls back to ''slim.timeout''). |
 | manually.start.test.runner.on.debug | false | Do not launch a SLIM server if the test is ran in debug mode. |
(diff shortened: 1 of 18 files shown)


six42:
Hi Felix,

Thanks for the PR.
Can you tell more about your use case.
What kind of Name Translations you are planning to do in what kind of situations?
You mentiond #859 But I understand some of these issues have been fixed already. 

In addition:
Arbitrary Class Name Translation can be done with: http://fitnesse.org/FitNesse.FullReferenceGuide.UserGuide.WritingAcceptanceTests.SliM.DefineAlias

Arbitrary Method Name Translation can be done with: FitNesse.SuiteAcceptanceTests.SuiteSlimTests.HybridDecisionTable

What needs to be added to thease features to make them work for you?

I know the Java Slim Client had already a dummy Name Translation class.
But in my opinion this class should be removed. The translation should be in the test system where the graceful naming is happening.
This would have the advantage that it would work for all Slim Clients implementations like .Net, C++, ...
So a smart solution would be to make the existing Disgracer configurable which is partially done with the HybridDecisionTable.

If you must do it in the Java Slim Client i suspect a new implementation of the interface "FixtureInteraction"  should do the job as well. See how it is use in the class: fitnesse.slim.fixtureInteraction.LoggingInteraction
and  look at http://fitnesse.org/FitNesse.FullReferenceGuide.UserGuide.WritingAcceptanceTests.SliM.InteractionAwareFixture

If you really need the name translator
please add a wiki test page which demonstrates and tests the usage of an non standard NameTranslator.
The test page should contain a decision table to show that they are working. (decision tables call internally functions like 'reset', 'execute', 'beginTable', ... which must not be translated.)
Further the test page should demonstrate that Symbols which reference classes are still working. FitNesse.SuiteAcceptanceTests.SuiteSlimTests.SlimSymbols.SlimSymbolCanHoldInstanceUsedAsParameter

FelixRiemann:
Hello six42,

thanks for your detailed reply.

The translations we are thinking of are replacements of special (non-alphanumeric) characters, e.g. German Umlauts. So character 'ä' would be replaced by 'ae' and so forth. This would allow using these characters on FitNesse Wiki pages to make the tests more readable, while still using ascii in the fixture code (see lalvincz' post on #859).
Doing this with aliases or dynamic/hybrid decision tables seems impractical.

The Disgracer also appears to be a good candidate, though at the moment it is used via its two static methods, which makes it harder to replace it dynamically.

However, your suggestion to reimplement FixtureInteraction also makes replacing characters possible, and slim can already be configured to use a particular implementation.

Thanks for the heads up on the special method names.

PR #965, opened 2016-08-25
Add SLIM Baseline decision table to make tests easier to read

six42:
The aim of the baseline decision table is to make tests easier to read.

Background
The idea of a decision table is to demonstrate how different combinations of input parameters generate different results.
To make the impact of each parameter to the output results transparent it is good practice to modify just one parameter from row to row.
For big tables with many input parameters it is often not immediately visible to the reader which parameter changes from row to row.

Syntax
As a user of the baseline decision table you specify in the first row below the header in each column the values for all input and output parameters.
This defines your base scenario and so far this is 100% identical to a normal decision table.
In all following lines you just specify the values which differ from the base scenario all other columns are left empty.

This makes your table more readable. It is also easier to maintain in case a value must be changed.
Don't repeat yourself is a good rule even when you write test cases. :)

Fixture Code
When the baseline decision table is tested all empty columns are filled by the Slim test system with the corresponding values from the base scenario.
In all other aspects the bahviour is identical to the decision table.
Your fixture code is identical to a decision table. The fixture will not be able to identif any difference between a baseline decision table and a normal decision table.

Example
Let's look at an example which you know already from the Decision Table

Given I have at least one milk remaining
Then I should NOT go to the store

|baseline: should I buy milk                                    |

| cash in wallet | credit card | pints of milk remaining | go to store? |
| --- | --- | --- | --- |
| 0 | no | 1 | no |
|  |  | 2 |  |
|  |  | 7 |  |
| 10 |  |  |  |
|  | yes |  |  |
| 10 | yes |  |  |
| 1 |  | 1 |  |

for more details see: FitNesse.SuiteAcceptanceTests.SuiteSlimTests.BaseLineDecisionTable

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/BaseLineDecisionTable/ComparatorsAreSupported/content.txt
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/BaseLineDecisionTable/ComparatorsAreSupported/content.txt
@@ -1,0 +1,16 @@
+|baseline: eg.Division                                           |
+|numerator|denominator|quotient?|# Explication                   |
+|10       |2          |5.0      |S1 Base scenario                |
+|20       |4          |         |S2 Same result as base - correct|
+|         |5          |2.0      |S4 Same numerator as base       |
+|16       |           |8.0      |S5 Same denominator as base     |
+|22       |7          |~=3.14   |S6 Special comparator           |
+|11       |2          |4<_<6    |S8 Special comparator           |
+|         |           |         |S9 Correct but useless duplicate|
+
+
+|baseline: eg.Division                                           |
+|numerator|denominator|quotient?|# Explication                   |
+|10       |2          |4<_<6    |S1 Base scenario                |
+|20       |4          |         |S2 Same result as base - correct|
+|11       |2          |         |S8 Special comparator           |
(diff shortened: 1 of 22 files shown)


amolenaar:
Can't you just do this with a normal decision table where you do a null/empty string check on the input? Instead of using primitives you can use their object counterparts.

six42:
A fixture could implement an identical behaviour for inputs. But not for outputs, except the implementor switches to a Slim TableTable type and you do everything in your fixture code.

But the goal is to decouple fixture implementation and test case format. The implementation stays as easy as with a normal decision table. And the test case writer gets two ways to express his requirements and can choose whatever he thinks is best suited in the current situation.

amolenaar:
One question: the changes are relative to the top row (the base scenario), right?

fhoeben:
I've never had a need for such a decision table, but there are more table types I've never used. One question though: how hard would it be to just make the standard codebase extensible enough so that this table type can be delivered by a plugin? Maybe that will create some extra flexibility (other) plugin developers may also benefit from, and you don't have to convince @amolenaar that this is a great extra table type :-)

six42:
Yes

six42:
A different design option would be to make it relative to the previous row.
But i think this becomes to difficult to keep the overview.
In the 10th row you have to trace back all previous 9 rows to understand which values are used.

amolenaar:
I just gave it a swing. It does read in a clean way. To me it's more obvious to relate the alterations to the top row.
 
I think it would be nice if the blank fields were filled in after a test execution, so it's really clear which values were tested with. Just like blank fields are filled with output in a normal decision table (tagged as "ignored", so they appear on a blue background).

Found one issue in the code. The second parameter of DecisionTableCaller is a boolean. This usually denotes there are actually 2 different things going on, so they should be separated in a different way (composition, subclassing, whatever). I thing the baseline table should use its own DecisionTableCaller implementation. This would also add to extensibility of this table type.

PR #967, opened 2016-09-01
Make autocomplete in plain text editor configureable

six42:
- Add any words for auto complete to the following page: .WikiAutoComplete
  Start Typing and press [CTRL][SPACE] to get proposals:
  autocomplete2
- Also all child page names are loaded and can be accessed by typing '>'
  followed by [CTRL][SPACE]
  autocomplete

Possible next step would be to identify all scenario names which are defined. The ?packet responder is a good starting point for this.
Or to identify all java classes and methods which are in the imports defined. 

I am currently blocked with other things. 
Maybe somebody feels inspired and can pick this up ...

### DIFF
--- a/FitNesseRoot/WikiAutoComplete/content.txt
+++ b/FitNesseRoot/WikiAutoComplete/content.txt
@@ -1,0 +1,21 @@
+!define
+!today
+!help
+!path
+!fixture
+!lastmodified
+!img -l
+!contents
+!contents -R2 -g -p -f -h
+!include
+!include -setup
+!include -seamless
+!include -teardown
+!define TEST_SYSTEM (SLIM)
+!define TEST_SYSTEM (FIT)
+!define MANUALLY_START_TEST_RUNNER_ON_DEBUG (true)
+!define MANUALLY_START_TEST_RUNNER_ON_DEBUG (false)
+!define COMMAND_PATTERN (java -cp %p %m)
+!define TEST_RUNNER ()
+!anchor
+!style_()
--- a/FitNesseRoot/WikiAutoComplete/properties.xml
+++ b/FitNesseRoot/WikiAutoComplete/properties.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<properties>
+<Edit/>
+<Files/>
+<Help>This file contains text samples which will be used to auto complete words in the wiki plain editor once you press [ctrl][space]. Each line contains one textblock.</Help>
+<Properties/>
+<RecentChanges/>
+<Refactor/>
+<Search/>
+<Suites>beta, codemirror, wiki</Suites>
+<Versions/>
+<WhereUsed/>
+</properties>
--- a/src/fitnesse/resources/codemirror/addon/hint/fitnesse_anyword-hint.js
+++ b/src/fitnesse/resources/codemirror/addon/hint/fitnesse_anyword-hint.js
@@ -12,7 +12,38 @@
 })(function(CodeMirror) {
   "use strict";
 
-  var WORD = /([@\!\$\w][\w]*)([^|]*\|)?/, RANGE = 500;
+  var WORD = /([@>!$\w]\w*)([^|]*\|)?/, RANGE = 500;
+  var autonames;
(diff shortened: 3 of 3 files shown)


amolenaar:
Why did you create an auto-complete page? I think it's mixing abstraction layers that way.

It may be better to fetch it from a file located in the resources package. That one can still be overwritten in the files section.

One other thing: the autocompletes data is loaded asynchronous, which means that if fetching data is slow (e.g. when working on a remote server) the autocomplete statements are not visible at all.

six42:
Hi @amolenaar,
just saw your question today.

I see your point about making it a page.
I expect every FitNesse user has his domain specific language which he should add to autocomplete.
So I wanted to make it as simple as possible for the user to do this. 
A page is much more user friendly than a file upload. 

Asynchronous loading has a big advantage with slow connections: You can start editing immediately. Auto complete will not be available in the beginning. Once the data has been loaded auto complete can be used. 
If it would be synchronous all users have to wait for the load to finish even if they don’t need auto complete.

amolenaar:
> I see your point about making it a page.
> I expect every FitNesse user has his domain specific language which he should add to autocomplete.
> So I wanted to make it as simple as possible for the user to do this. 
> A page is much more user friendly than a file upload.

I still have some doubts. Having autocomplete information in a wiki pages makes it easier to modify indeed. But then, it's another special page that we'll introduce, and we have quite a few already. I'd rather keep the number of special pages to a minimum. I can foresee that people want to have auto-complete pages on every level in the wiki tree (like we have with ScenarioLibrary and TemplateLibrary).

I also feel for making it a separate responder (e.g. `?suggest`), so it can be extended with information from (the relevant part) of the wiki tree.

six42:
I agree this is not the final concept for auto-complete.
It is a proof of concept that the FitNesse Wiki technology 
* supports auto-complete 
* with asynchronous AJAX 
* and adjusted based on the current page. 
and is adds already value for the users. Therefore it is worse to have it. 

I am not sure how the final concept will look like. 
Likely it will be a dedicated responder.
Ideally there is no need to add manual items for the auto-complete all is derived automatically.
But this is still a long way:
* Scenarios must be derived from the wiki page in scope. For this you need to know if it is Slim or Fit page.
* classes, methods and functions must be extracted from the code in scope. 
 * Plugins likely want to add their functionality to the auto-complete library.

I can tag the auto-complete wiki page as BETA with a comment that this is just experimental and likely to change and even might disappear in the future.

Would this give you enough comfort to merge?

fhoeben:
Does the page reference completion only work for child pages, or for any (relative or absolute) reference?
If it were to work for any that would also be great for moving pages...

tcnh:
I did a first stab at a responder that returns json containing all methods (by class) and scenario's available in a given context. See https://github.com/tcnh/FitNesseAutocompleteResponder

This is a first version that needs polishing, but should get us close to a usable autocomplete in the wiki editor if combined with @six42 's effort in this PR.

tcnh:
I also updated this PR's javascript file as a proof-of concept. Commit is at: https://github.com/tcnh/fitnesse/commit/5a1c8787438dda11491cd495d8600c0dcdb2965a

Usage: add the responder for ?autoComplete and edit pages. All imported fixture methods and available scenario's are now in the autocomplete list. When editing a scenarioLibrary, do so from a test page. In that case the test case context will be used, so that relevant fixture methods are also available.

six42:
Hi @fhoeben and @tcnh,
I just see your comments today. Strange that I was never informed via mail. 

Fried 
I used the standard "names" responder. (/fitnesse/src/fitnesse/responders/NameWikiPageResponder.java) 
And to not impact performance I limited it to the direct child pages.
If you add the following to your root page you get a list of all pages in your FitNesse instance:
?names&Recursive
Having this integrated for the move page function would be very good.

Tom, 
thanks for extending the autocomplete further.  🥇 👍 
Would be great if your extension could be come part of the standard FitNesse package.
I also like your theme and think it would add value to everybody and should become the standard theme.

I will close this PR! 
Any additions should be added at: https://github.com/tcnh/fitnesse-bootstrap-plus-theme and https://github.com/tcnh/FitNesseAutocompleteResponder

Issue #972, opened 2016-09-14
CachedInteraction cannot be used


fhoeben:
When I try to use fitnesse.slim.fixtureInteraction.CachedInteraction as Slim interaction I am unable to do so.

It turns out the class cannot be instantiated....

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: 0
	at fitnesse.slim.fixtureInteraction.CachedInteraction.<clinit>(CachedInteraction.java:10)
```

I will try to send a pull request to address this shortly

amolenaar:
Will close alongside #973.

PR #973, opened 2016-09-14
CachedInteraction fix

fhoeben:
Fix for #972, and some initial tests for CachedInteraction


fhoeben:
@LudovicVa Did I miss something regarding CachedInteraction, or am I indeed fixing it?

LudovicVa:
I think you did fix it !! My bad...

PR #974, opened 2016-09-17
All scenario resolution in test context

fhoeben:
This pull request moves all scenario resolution logic to the slim test context.
Furthermore it only creates the pattern to match a single cell scenario call with parameters once, instead of each time the scenario is evaluated to determine whether it matches.

### DIFF
--- a/src/fitnesse/testsystems/slim/SlimTestContext.java
+++ b/src/fitnesse/testsystems/slim/SlimTestContext.java
@@ -18,6 +18,8 @@ public interface SlimTestContext {
 
   ScenarioTable getScenario(String scenarioName);
 
+  ScenarioTable getScenarioByPatternMatching(String invokingString, CustomComparatorRegistry customComparatorRegistry);
+
   Collection<ScenarioTable> getScenarios();
 
   void incrementPassedTestsCount();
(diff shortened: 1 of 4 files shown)


src/fitnesse/testsystems/slim/SlimTestContext.java
```diff
@@ -18,6 +18,8 @@
 
   ScenarioTable getScenario(String scenarioName);
 
+  ScenarioTable getScenarioByPatternMatching(String invokingString, CustomComparatorRegistry customComparatorRegistry);
```
amolenaar: I think `getScenarioByPattern` is a more appropriate name.

src/fitnesse/testsystems/slim/tables/ScenarioTable.java
```diff
@@ -330,6 +334,11 @@ public ScenarioTable getScenario(String scenarioName) {
     }
 
     @Override
+    public ScenarioTable getScenarioByPatternMatching(String invokingString, CustomComparatorRegistry customComparatorRegistry) {
```
amolenaar: Same method rename here.

fhoeben:
Renamed as suggested, and added small optimization: no need to sort the scenarios for each lookup

PR #976, opened 2016-09-19
Fix failing junit test under Windows

six42:
junit test "WikiFilePageTest.java" failed under Windows.
The below make it work and I assume it will still work under Linux and Mac


amolenaar:
Ah :)

PR #977, opened 2016-09-27
Adding a socket free SlimService.  Remove all hassles with port conflicts and network setup

six42:
Port number 1 has now a special meaning: Don’t open a socket. Instead communicate via StdIn and StdOut. Nevertheless all output from the SUT is captured, tunneled via StdErr and added to the Execution Log.

This implementation is now the default behavior (= port number 1 is the default) 
The change is fully backward compatible by specifying any other port number.

The Slim protocol number has been increased to 0.5

As there is no need to open a socket it now also easier to implement a Slim Server in other programming languages.
 
The socket functionality is still useful for distributed testing and should be preserved also in the future. Last but not least it is required for backward compatibility.


amolenaar:
I like it that you can use FitNesse without sockets. This will rule out a whole class of errors. Kudos!

From the PR it was not clear to me how log messages on stdout and stderr mix with the Slim output.

src/fitnesse/slim/SlimSlaveSocket.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public SlimSlaveSocket() throws IOException {
+
+    // preserve original streams
+    this.stdout = System.out;
+    this.stderr = System.err;
+    this.stdin = System.in;
+
+    // bind System.stdout/System.stderr to original stderr
+    System.setOut(new PrintStream(new LoggingOutputStream(this.stderr, "SOUT"),
```
six42: @amolenaar  
This redirects all messages send to stdout to stderr. So stdout is free. SLIM can send results without the SUT mixing in any log messages.
In line 58 in the "accept" method stdout is bound as communication channel for SLIM

amolenaar:
I am trying to use the SocketFactory to also create the pipe-like socket. Expect a PR soon ;)

six42:
Thanks. All merged.

amolenaar:
Awesome. I merged the PR.

PR #978, opened 2016-09-29
Add urlpathvariable overrides to FitNesseRunner.

ctrl-space:
First incomplete iteration of adding urlpathvariables to FitNesseRunner.
Feedback welcome.

Thanks,
Claes


amolenaar:
Hi, I think this is starting to look quite okay. You might want to use a value array, instead of concatenating all properties with `&'. Then the format is closer to Java.

Something like this:

    @FitNesseRunner.UrlPathVariables({ "ID=100", "gizmo=true"})

You might want to figure out to what extend the current parser logic (e.g. the `Request` class) can be used. Some change might be required, but it will keep the code DRY.

(and the variables need to be used somewhere, of course)

ctrl-space:
Updated annotation and using Request to parse query string. Added urlpathvariables to context.

PR #992, opened 2016-11-29
Issue 859 gracefull names

### LINKED ISSUE #655: Capital column names in tables require incorrect Java method name convention
If you have a capital column name like this:

```
|Do Something?|
```

the correspending method must be capitalized:

```
DoSomething()
```

This is incorrect for Java, and should instead be:

```
doSomething()
```

After searching the Internet for some time, the only reference I found to this issue was on the Yahoo group:

https://groups.yahoo.com/neo/groups/fitnesse/conversations/topics/20509

The problem seems to be this:

https://github.com/unclebob/fitnesse/blob/master/src/fitnesse/slim/MethodExecutor.java#L20

Is this by design? This seems pretty short-sighted, and I'm honestly surprised at the lack of discussion or complaint I've found online concerning this issue. The only conclusion I've come to is that everyone is just expected to make their column names lower case. My two issues with this are 1.) some people may find this less appealing or readable, and 2.) I haven't found anywhere in the documentation that says "make your column names lower case". (If this does exist in the documentation, please point me to it.)

### LINKED ISSUE #783: Using Table table with single word class doesn't work if class defined in lower case
Hi.

I updated my project from 20150226 to 20150814 (2 versions)
And now if I'm using [Table table](http://fitnesse.org/FitNesse.FullReferenceGuide.UserGuide.WritingAcceptanceTests.SliM.TableTable) with lower case class.
In example it'll be ```Table: bowling```.
 it couldn't find a constructor for the class.

I've found a workaround with aliases:
```
| Define alias      |
| bowling | Bowling |
```
But it's still a workaround...

Is it a solvable problem? And is it a bug or a feature? :)

### LINKED ISSUE #859: Graceful names not work in some cases

### LINKED ISSUE #863: Not possible to pass an array in fixture
I tried with a fixture with below sample script table but it does not recognize the method as a list as input value. Here is my code
First line  code : 
public List<String> **ReturnAList()** {
        List<String> names = new ArrayList<String>();
        names.add("a");
        names.add("b");
        return names;
    }
Second line code: 
public boolean **IsSorted**(List< String > values) {
            return Ordering.natural().isOrdered(values);
    }

image

salandur:
This fixes #859, fixes #863, fixes #655, fixes #783

The best way to fix the loading of classnames and matching methods was not in the Disgracer, which works fine.
The fix is a bit deeper in the system where the actual class or method is found, so we have a good fallback option.

### DIFF
--- a/src/fitnesse/slim/fixtureInteraction/DefaultInteraction.java
+++ b/src/fitnesse/slim/fixtureInteraction/DefaultInteraction.java
@@ -65,8 +65,8 @@ public class DefaultInteraction implements FixtureInteraction {
     throw new SlimError(String.format("message:<<%s %s>>", SlimServer.NO_CLASS, className));
   }
 
-  private String swapCaseOfFirstLetter(String className) {
-    return StringUtils.swapCase(className.substring(0, 1)) + className.substring((1));
+  private String swapCaseOfFirstLetter(String classOrMethodName) {
+    return StringUtils.swapCase(classOrMethodName.substring(0, 1)) + classOrMethodName.substring((1));
   }
 
   protected Class<?> getClass(String className) {
@@ -110,13 +110,19 @@ public class DefaultInteraction implements FixtureInteraction {
 
     int nArgs = args.length;
     for (Method method : methods) {
-      boolean hasMatchingName = method.getName().equals(methodName);
-      boolean hasMatchingArguments = method.getParameterTypes().length == nArgs;
-      if (hasMatchingName && hasMatchingArguments) {
+      boolean isMatchingMethod = findMatchingMethod(methodName, nArgs, method);
+      isMatchingMethod = isMatchingMethod || findMatchingMethod(swapCaseOfFirstLetter(methodName), nArgs, method);
+      if (isMatchingMethod) {
         return method;
       }
     }
     return null;
+  }
+
+  private boolean findMatchingMethod(String methodName, int nArgs, Method method) {
+    boolean hasMatchingName = method.getName().equals(methodName);
+    boolean hasMatchingArguments = method.getParameterTypes().length == nArgs;
+    return hasMatchingName && hasMatchingArguments;
   }
 
   protected MethodExecutionResult invokeMethod(Object instance, Method method, Object[] args) throws Throwable {
--- a/src/fitnesse/slim/fixtureInteraction/DefaultInteraction.java
+++ b/src/fitnesse/slim/fixtureInteraction/DefaultInteraction.java
@@ -10,6 +10,7 @@ import fitnesse.slim.ConverterSupport;
 import fitnesse.slim.MethodExecutionResult;
 import fitnesse.slim.SlimError;
 import fitnesse.slim.SlimServer;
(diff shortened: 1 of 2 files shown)


src/fitnesse/slim/fixtureInteraction/DefaultInteraction.java
```diff
@@ -22,138 +23,153 @@
...
+      isMatchingMethod = isMatchingMethod || findMatchingMethod(swapCaseOfFirstLetter(methodName), nArgs, method);
+      if (isMatchingMethod) {
+        return method;
+      }
+    }
+    return null;
+  }
+
+  private boolean findMatchingMethod(String methodName, int nArgs, Method method) {
+    boolean hasMatchingName = method.getName().equals(methodName);
```
six42: Same here:

equalsIgnoringCaseOfFirstLetter

six42:
Hi Salandur,

A little  sad that this can't be fixed in the disgracer. Would be a better place.

Can you update the Wiki documentation as well about the new logic and that it only applies to Java.

salandur:
Indeed fixing it in the Disgracer would have been nicer, but that means the Disgracer needs to give back more than 1 option. So deferring to the location where the actual class is loaded or the method is matched is then better.

I will restructure the code a bit so that the new behavior is only triggered after the old method, not mixed like now.

I surely will update the documentation too.

amolenaar:
This looks like the right solution. Like FitSharp, the matching is now more language (convention) dependent.

salandur:
I was thinking of changing [Disgracer:77](https://github.com/unclebob/fitnesse/blob/e38a89a32bebb0ec099219d58d7aab5e7b1b54e6/src/fitnesse/testsystems/slim/tables/Disgracer.java#L77) to `disgracedName.append(capitalizeNextWord ? toUpperCase(c) : toLowerCase(c));`, but I don't know if that might have unwanted side effects

six42:
Could break some cases:

disgrace ClassName -> disgraceClassname instead of currently disgraceClassName

SQL command -> sqlCommand instead of currently SQLCommand

Also would not be enough as single word names are not covered in the current code:

"Empty" as method name is never disgraced in the current code. 

You could be strict and always lowerCase the first letter for Methods and upperCase the first for Classes in the Disgracer and use the existing rule for the remaining chars
This would be consistent.
But might also break existing tests which must be fixed with the code from your current patch.
But not sure how this would impact .net or other languages.

Issue #962, opened 2016-08-18
There is an error in the text for the definition of WikiWord


siegfried1e:
in fitnesse/FitNesseRoot/FitNesse/UserGuide/FitNesseWiki/WikiWord/content.txt
a phrase as a redundant "more" in it
... a wiki word is a string of two **more more** capital letters with ...

amolenaar:
Thanks for pointing out.

PR #994, opened 2016-12-01
removed double more in WikiWord documentation

salandur:
fixes issue #962


amolenaar:
Hi Jeroen, good to see you're still into FitNesse :)

salandur:
:-)

Issue #989, opened 2016-11-18
Slim process dies without any error in log with FitNesse 20161106


fhoeben:
I have a test, using Slim, which works with the previous version of 20160618. It also works with version 20161106 if I run it via jUnit using `FitNesseRunner` with Slim in-process. It does not work when started from the wiki (separate Java process for Slim), on Windows 7.

I just get a message that the communication with the Slim process was lost, and in the execution log I see no indication of what went wrong. The Slim process just seems to die, without any error output.
A part of the test is executed, the SuiteSetUp and SetUp, but Slim dies on the real test page. 
I've placed logging (to standard error) in my fixture code and I see that in the output of my SetUp pages, but the constructor of the fixture of the first (script) table never seems to be called. So I assume something happens in the code between the wiki and the Slim process before my fixture is even created.

I have no clue what the cause is, but processes stopping without any indication of what went wrong are very annoying....

fhoeben:
It turns out there was 'something' in the execution log. I found that the slim process died with exit code 98. Which I tracked in the source to mean that an exception occurred. Unfortunately the logging of the exception did not work (as I did not see that in the execution log).

When I configure the connection to the slim process to be done on a random free port (i.e. I added `!define slim.port {0}` to the root of my suite, where I also configure the test system to be slim). The test works again.

My conclusion is that something is going wrong with the 'pipe based' connection between FitNesse and Slim. And logging on exception at the level does not end up in the execution log.

amolenaar:
Can we replay this scenario?

fhoeben:
I haven't been able to make a short sample that shows the problem. I have encountered the problem in multiple test sets in my current projects. All tests only show the problem on Windows 7 when started from the wiki (i.e. not when run on a build server using the jUnit runner on Linux).
The problem goes away on downgrade of FitNesse or when I set the slim.port variable.

fhoeben:
I just checked #995 is the same problem as this issue. Adding the `!define slim.port {0}` also fixed the problem with the euro symbol

keetron:
Gonna do that right away on Saturday night so I wont have to remember until Monday...

fhoeben:
I believe the problem comes from the new 'default usage of console input/output' to communicate between wiki and Slim. On Windows the console does not support Unicode, at least not by default (see for instance http://stackoverflow.com/questions/20386335/printing-out-unicode-from-java-code-issue-in-windows-console). 
[...]

PR #996, opened 2016-12-04
Don't use pipe via console on windows

fhoeben:
Prevent problems with multi byte characters when running slim tests on windows, issue #989

### DIFF
--- a/src/fitnesse/testsystems/slim/SlimClientBuilder.java
+++ b/src/fitnesse/testsystems/slim/SlimClientBuilder.java
@@ -14,6 +14,7 @@ import fitnesse.socketservice.PlainClientSocketFactory;
 import fitnesse.socketservice.PlainServerSocketFactory;
 import fitnesse.socketservice.SslClientSocketFactory;
 import fitnesse.testsystems.*;
+import org.apache.commons.lang.SystemUtils;
 
 public class SlimClientBuilder extends ClientBuilder<SlimCommandRunningClient> {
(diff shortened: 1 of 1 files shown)


fhoeben:
Fix prevents stack trace on Windows 10, but original problem did not occur there so not sure yet that this is full fix for #989

fhoeben:
Checked this patch on Windows 7, issue #989 indeed no longer occurs.

amolenaar:
If #997 works fine, I'd prefer that patch.

six42:
I tested with WIndows 10 and Java 8 and have no problems: 
No Stack Trace 
and test with € is parsing.

Preventing pipes for Windows  is to radical.
Anybody who has problems with pipes  can just add to the root page
!define SLIM_PORT (8891) 
This will give back the default behaviour of the previous FitNesse version.
You can also define in the properties file 
slim.port =  8891
to get the same effect.

fhoeben:
I believe the 'old behavior' was a random port, which was my workaround, using port '0' which selects a random port. Which is also the solution I use in this pull request.

One thing that is not clear to me: what is the advantage of not using a port?

amolenaar:
One advantage is that there is no (re)cycling of ports, so you can run as many FitNesse instances as you like. This is a nice property of you're running your tests on CI for example.

fhoeben:
I still see no problem using a random port in the ephemeral port range, but since defining `SLIM_PORT {0}` gets the same result I'll close this PR

Issue #995, opened 2016-12-01
When using the Euro-symbol in a decision table, FitNesse errors out "cannot connect to SUT"


keetron:
The title says it all, when running a test with the Euro-symbol €, the test fails to connect to the SUT. Removing the symbol resolves this, but the test then fails of course. Is this known behavior? I tried putting !- -! around the expression but it made little difference.

Console gives this error:
Test system terminated with exception [fitnesse.testsystems.TestExecutionException: fitnesse.testsystems.slim.SlimCommunicationException: Could not send/receive data with SUT]

six42:
Hi Keetron,

€ works nicely for me.

can you try if below test works as it is first and second after you add a €:
http://localhost:8001/FitNesse.SuiteAcceptanceTests.SuiteSlimTests.MultiByteCharsInSlim

Which version of FitNesse, which OS and which programming language are you using? 

![euro](https://cloud.githubusercontent.com/assets/5906592/20815525/802f779c-b81e-11e6-8718-1f1af6aee445.JPG)

keetron:
We are actually using quite the Christmas tree of addons to do what we want to do.

First, we use the latest fitnesse on Windows 7 with the HSAC plugin.
Second, the value from the decision table with the € is used in a table template that puts it in a map so we can use it to modify XML stubs as well as use/check the value in the screen and later to validate the output XMLs as well.
Finally this all runs in a VM.

I blaim hsac to be honest, right now I moved the testcase to the comment table "framework has issues with €" and I'll focus on the dynamic stubbing more as that has much wider use and impact to get working right.

fhoeben:
On Windows 7, IBM JVM 7 64bit http://localhost:8001/FitNesse.SuiteAcceptanceTests.SuiteSlimTests.MultiByteCharsInSlim causes the same problem as mentioned by @keetron.
Adding `!define slim.port {0}` to .FitNesse.SuiteAcceptanceTests.SuiteSlimTests fixes the problem and makes the test green.

keetron:
Yes, that fixed it. Checked when running the test individually or in a suite.

amolenaar:
I tried with the Win7+IE10 box from http://modern.ie, and can not reproduce the error.

<img width="206" alt="screen shot 2017-01-14 at 21 53 32" src="https://cloud.githubusercontent.com/assets/96249/21958002/eaf85a1a-daa3-11e6-89c4-1d52e12d7236.png">

[...]

PR #997, opened 2016-12-04
Fix encoding for pipe based socket

### LINKED ISSUE #989: Slim process dies without any error in log with FitNesse 20161106

amolenaar:
Should fix #989 and #995.

### DIFF
--- a/src/fitnesse/slim/SlimPipeSocket.java
+++ b/src/fitnesse/slim/SlimPipeSocket.java
@@ -42,12 +42,10 @@ public class SlimPipeSocket extends ServerSocket {
     this.stdin = System.in;
 
     // bind System.stdout/System.stderr to original stderr
-    System.setOut(new PrintStream(new LoggingOutputStream(this.stderr,
-        STDOUT_PREFIX),
-        true));
-    System.setErr(new PrintStream(new LoggingOutputStream(this.stderr,
-        STDERR_PREFIX),
-        true));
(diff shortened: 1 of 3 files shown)


amolenaar:
I have tried testing it on my Windows XP VM, but I can not reproduce the errors stated in #989.

I noticed, however, that the output streams on the Slim side do not apply the proper encoding (=UTF-8).

This PR should fix that. I hope someone can verify that (@fhoeben, @keetron)?

fhoeben:
Unfortunately that does not fix the issue #989 on Windows 7.
It does fix the stack trace I encountered in the console on Windows 10, running Slim's multi-byte tests.

amolenaar:
:(

amolenaar:
Figured this change wouldn't hurt anyway.

PR #1000, opened 2016-12-09
Added wiki.page.old.style = true to the ConfigurationFile page

keetron:
Just a one liner again, some day there will be two.


keetron:
I changed my mind, we should migrate the entirety of the application to new style files.

Issue #966, opened 2016-08-31
SliM execute fixture constructor based on parameters types


george-toma:
Hi folks,

While using FitNesse SliM protocol, I encountered a "nice to have feature", the fixture constructor call should be based on the parameters types.
E.g. : 
- |Query:my business fixture|Michael|5|
  -- should execute the constructor "trigger X "
- |Query:my business fixture|20120210|Adela|
  -- should execute the constructor "trigger Y"
- |Query:my business fixture|20120210|10|
  -- should execute the constructor "trigger Z"

And the Java class : 
class MyBusinessFixture{

public MyBusinessFixture(String s, String p){
...trigger X....
}
public MyBusinessFixture(Date date, String p){
...trigger Y....
}

public MyBusinessFixture(Date date, Integer i ){
...trigger Z....
}
...my methods...
}

In the current code, an iteration is done over all class constructors, and the first match on the number of parameters is picked up.

**I would like to implement this fix** by reusing the current code, mostly ConverterRegistry.java, and the fix should always pick up the constructor based on the parameters types ( e.g. 1 can be mapped to a String and to an Integer, but would seems logical to pick up the constructor with an Integer parameter).

Can someone provide me a feedback regarding the fix I want to do ?

Regards,
George

six42:
Hi @george-toma ,
I think this would be a nice and useful feature.
I am not familiar with this part of the code base so I can't comment on your approch.
Go ahead.

Six42

PR #1016, opened 2017-01-16
SliM execute fixture constructor based on parameters types #966

george-toma:
The scope of this fix is that when a fixture is executed, the fixture class constructor should be selected by taking in consideration the constructor arguments types.


george-toma:
Another fix is needed + an extra test for a specific case.

Will create another pull request

PR #1025, opened 2017-01-23
Implemented table of content of all page headings

jdufner:
* Inserted a id attribute to headings
* Implemented a inner classes to
  * extract headings
  * create ordered list
  * build links to headings
* Implemented some unit tests

Please do a review and give me some hints for documentation and example.

### DIFF
--- a/src/fitnesse/wikitext/parser/Headings.java
+++ b/src/fitnesse/wikitext/parser/Headings.java
@@ -1,0 +1,238 @@
+package fitnesse.wikitext.parser;
+
+import fitnesse.html.HtmlElement;
+import fitnesse.html.HtmlTag;
+import fitnesse.html.HtmlUtil;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Stack;
+
+/**
+ * Generates a ordered list of all headers from within the current wiki page.
+ */
+public class Headings extends SymbolType implements Rule, Translation {
+
+  public static final Headings symbolType = new Headings();
+
+  private static final String STYLE = "STYLE";
+  private static final String[] OPTION_KEYS = new String[]{STYLE};
+
+  public Headings() {
+    super("Headings");
+    wikiMatcher(new Matcher().startLineOrCell().string("!headings"));
+    wikiRule(this);
+    htmlTranslation(this);
+  }
+
+  @Override
+  public Maybe<Symbol> parse(Symbol current, Parser parser) {
+    final Symbol body = parser.parseToEnd(SymbolType.Newline);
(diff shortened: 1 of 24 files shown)


src/fitnesse/wikitext/parser/Headings.java
```diff
@@ -0,0 +1,202 @@
...
+          headerLines.add(symbol);
+        }
+      }
+    }
+
+    private void extractSyntaxTreeFromTranslator() {
+      HtmlTranslator htmlTranslator = (HtmlTranslator) this.translator;
+      ParsingPage parsingPage = htmlTranslator.getParsingPage();
+      WikiSourcePage wikiSourcePage = (WikiSourcePage) parsingPage.getPage();
+      BaseWikitextPage baseWikitextPage = (BaseWikitextPage) wikiSourcePage.getPage();
```
amolenaar: There are a couple of problems with this approach. Foremost it's breaking the abstraction between parser and wiki page (the `WikiSourcePage` class). I think it would be better to "ask" the source page for symbols of a particular type, instead of traversing the object graph down to the syntax tree.

On the coding side it would be better to return the SyntaxTree, instead of putting it in an instance variable.

src/fitnesse/wikitext/parser/WikiSourcePage.java
```diff
@@ -127,4 +127,9 @@ private boolean isParentOf(WikiPage possibleParent) {
     public int compareTo(SourcePage other) {
         return getName().compareTo(other.getName());
     }
+
+    public WikiPage getPage() {
```
amolenaar: This method creates a leak in the abstraction. From the parser point of view, it should not be aware of the `WikiPage` interface.

amolenaar:
This is a nice PR. It's missing some documentation, as you already mentioned.

I think it needs a mention in the Quick reference guide at least (http://fitnesse.org/FitNesse.FullReferenceGuide.UserGuide.QuickReferenceGuide). Also a test case describing the intent (a subpage of http://fitnesse.org/FitNesse.SuiteAcceptanceTests.SuiteWidgetTests) would be a good addition.

jdufner:
Thanks for your review. I'll rework during next week.

2017-10-15 16:24 GMT+02:00 Arjan Molenaar <notifications@github.com>:

jdufner:
I reworked this pull request. From my point of view everything look well.

One point to discuss: The HTML / CSS style of this table of contents differs from that of the wiki word !contents. Do you think !contents and !heading should produce the same HTML / CSS structure?

PR #1035, opened 2017-02-13
All base installation wiki files now in .wiki format

keetron:
Migrated all base installation wiki files, edited build.gradle to match the new files, some changes in gradle-wrapper were made by intellij.


keetron:
Today when using this I found that the FrontPage *has* to exist in FrontPage/context.txt.
Is that hardcoded?

amolenaar:
@keetron The frontpage should also work as a `.wiki` page.

I'm wondering though. If you have an installed version of fitnesse, with the pages in content/properties files, would the old files still exist? I mean: the `FitNesse` folder should be cleaned up too.

keetron:
Not sure, it is a long time ago. Let me grab some old installation and do some testing.

keetron:
Just did some checks.
I use the 2016110 version of FitNesse. All files are in .wiki format, other than the FrontPage, which is content.txt in the FrontPage folder. There is nothing special in the properties.
If I remove the FrontPage folder and run the command `mvn compile exec:exec` then the folder stays gone. If I remove the files but not the folder, the content.txt file returns, the FrontPage.wiki file will be ignored.
Resolution is removing the folder and not just the files. I will make some changes to this PR after I updated the branch and recommit.

keetron:
Ok, asking the wiki for the FrontPage will re-generate a default FrontPage folder with the content.txt etc.
It is hardcoded and the FrontPage.wiki file is ignored.
Next up: me searching the java code where this is done and nuke it from orbit.

keetron:
completely forgot this was open.

PR #1037, opened 2017-02-21
Bootstrap theme improvement for tabbed browsing

tcnh:
Minor bootstrap theme improvement:
Show different favicons for tests, suites and editing.
Improved page title display for users who use browser tabs.

Before this PR: 
image

After this PR (same pages are open):
image


tcnh:
Extracted this in a separate theme to be able to easily extend: https://github.com/tcnh/fitnesse-bootstrap-plus-theme

No need to do this in fitNesse's repository

PR #1046, opened 2017-04-29
Added two new keywords - 'check partial' and 'check partial not'

suratdas:
Many times we have to ensure that the text contains certain word to decide if the step is pass/fail. Keyword check is not sufficient as it checks the whole text. So, we have been writing extra code ( DoesContain(String,String) method in this example) to take care of partial validation like below.
image
With this pull request, this can be simplified in wiki and also we don't have to write native code to check text contain validation. The proposed way will be like this.
image
Similarly, we can use 'check partial not' keyword for negative validations in steps.

### DIFF
--- a/src/fitnesse/testsystems/slim/tables/ScriptTable.java
+++ b/src/fitnesse/testsystems/slim/tables/ScriptTable.java
@@ -2,10 +2,6 @@
 // Released under the terms of the CPL Common Public License version 1.0.
 package fitnesse.testsystems.slim.tables;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-
 import fitnesse.slim.converters.BooleanConverter;
 import fitnesse.slim.converters.VoidConverter;
 import fitnesse.slim.instructions.Instruction;
@@ -14,6 +10,10 @@ import fitnesse.testsystems.TestResult;
 import fitnesse.testsystems.slim.SlimTestContext;
 import fitnesse.testsystems.slim.Table;
 import fitnesse.testsystems.slim.results.SlimTestResult;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 
 public class ScriptTable extends SlimTable {
   private static final String SEQUENTIAL_ARGUMENT_PROCESSING_SUFFIX = ";";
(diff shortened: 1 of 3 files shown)


tcnh:
This does look nicer in the wiki, but are you aware that using regular expressions with check and check not is also supported?

suratdas:
Hi @tcnh . I checked the code in SlimTable.java and found that 'check' keyword does an exact equal check per below snippet. I could not figure out how I can use regular expression with 'check' for partial validations.
```
else if (actual.equals(replacedExpected))
        testResult = SlimTestResult.pass(announceBlank(replaceSymbolsWithFullExpansion(expected)));
```

tcnh:
Hi @sarutdas, Ik can't look into it right now but see http://www.fitnesse.org/FitNesse.UserGuide.WritingAcceptanceTests.SliM.ValueComparisons for some usage examples and syntaxis.

suratdas:
Thank you @tcnh for pointing out this. I have made use of regular expression and updated the pull request. In order for easier automated test authoring and analysis by non-developer. it will be very useful to have these two new keywords - 'check partial' and 'check partial not'. Also, it will read better if somebody decides to execute these steps manually for any reason.

ALuckyGuy:
Personally I think the word "partial" is a bit unusual. I think "Contains" would be a better keyword, especially because contains() is a common and familiar function supported by strings.

amolenaar:
I'm hesitant to add this feature. For one it can also be expressed easily in a fixture (`ensure something contains _xyz_`). The phrasing is a little different, but the message is the same. If you want to see the output, add a `show` case as well.

suratdas:
@ALuckyGuy - Yes, I agree. I can change it to that if approved.
@amolenaar  - Yes, this is how we have it organized now using ```ensure``` and ```show``` 
This wiki is often maintained by QA/BA without automation engineers involvement and they update the wiki themselves but they feel it is too cluttered and lengthy. That prompted me for this pull request. I am okay if we still think that's not worth the change.
image

keetron:
These are prima candidates for a scenario that would achieve the same thing. You can even name it "check partial".
Not sure what your fixture looks like but with a scenario you can bundle repeating wiki instructions within the wiki itself. I think think would resolve your issue without any additional fixture changes.

amolenaar:
I'll close this PR. Just reopen if you think more discussion is required.

PR #1057, opened 2017-06-20
Fixing issue 1050 (.wiki format not handling CRLF)

essenius:
Simplest fix that works: translate CRLF to LF on content read.

### DIFF
--- a/src/fitnesse/wiki/fs/WikiFilePage.java
+++ b/src/fitnesse/wiki/fs/WikiFilePage.java
@@ -144,7 +144,7 @@ public class WikiFilePage extends BaseWikitextPage implements FileBasedWikiPage 
 
     if (fileVersion != null) {
       try {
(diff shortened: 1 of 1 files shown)


src/fitnesse/wiki/fs/WikiFilePage.java
```diff
@@ -144,7 +144,7 @@ private PageData getDataVersion() throws IOException {
 
     if (fileVersion != null) {
       try {
-        String fileContent = loadContent(fileVersion);
+        String fileContent = loadContent(fileVersion).replace("\r\n", "\n");
```
keetron: May I suggest using `replaceAll("\\r\\n", "\n")` as to ensure the regex grabs the line endings properly as replace uses literal matches?

amolenaar:
I think #1058 is a nicer solution. Do you agree?

essenius:
Agreed.

PR #1073, opened 2017-08-09
Fix crash when Table Table returns extra columns

legoscia:
As per the documentation, a table table fixture should be able to
return rows that are wider than in the input.  However, an off-by-one
error was preventing that.


amolenaar:
Is this PR trying to fix the same issue as #1055?

legoscia:
Yes, that's the same issue. Didn't see that one when I opened my pull request...

amolenaar:
I merged #1055. I hope you're fine with that.

PR #1079, opened 2017-09-04
A SuiteQuery should not run test pages that are excluded

NicoleRauch:
Current situation: When a test page is skipped, and there is a SuiteQuery that matches this page, it will execute the page although it is skipped.

This pull request addresses this situation and makes the SuiteQuery ignore skipped tests.

Note: Although the skipping is meant to be recursive, the SuiteQuery does not yet ignore children of a skipped test. This is a bit more complicated, and I will try to work on it. But I think even this small improvement is worthwhile, therefore I'm sending it as an individual PR.


jdufner:
+1

amolenaar:
Thanks :)

PR #1087, opened 2017-09-25
Added publishing in-execution individual test results to JMS

suratdas:
Let's say, Fitnesse suite execution is triggered from standalone program/script/Jenkins. If the execution runs for hours (e.g. Selenium tests), there is no easy way to programmatically know the real time execution status before its completion. Real time status will be useful for : 
- Programmatic real time display of test progress. 
- Real time result update in test management system (JIRA/ALM/TFS etc.) using their Rest API. 
- Alert/email stakeholders for specific failure(s) without waiting for hours till suite execution completion.

This pull request uses JMS and ActiveMQ libraries to publish the results in real time. Below are the steps to achieve this goal:

1. In Fitnesse root, we have to add two variables 
``` 
!define MESSAGE_BROKER_ADDRESS {tcp://<server>:61616}
!define QUEUE_NAME {FitnesseExecution.TESTQ}
 ```
2. Ensure ActiveMQ is running in the address specified above as MESSAGE_BROKER_ADDRESS. ActiveMQ will capture test result in below format:
`{"ignores":0,"testPath":"<suitePath.testName>","right":1,"exceptions":0,"testName":"<testName>","wrong":0}`
3. Execute suite with additional query parameter (publish) like below
`http://<fitnesseServerUrl>/<suiteName>?suite&publish`
4. Consume the service from ActiveMQ as needed (display/update test management system/alert).

I have tested it and it does not impact the old functionality. New functionality can be achieved by using above steps. Since I am not very familiar with gradle, I used below command to successfully test this idea
`java -cp activemq-all-5.15.0.jar;javax.jms-api-2.0.1.jar;fitnesse-standalone.jar fitnesseMain.FitNesseMain -e 0 -a passwords.txt -p 9090`

Improvements (unit test, documentation update, program refactor etc.) can be made in future commits if this idea and approach is approved.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,10 +66,12 @@ dependencies {
   compile "org.htmlparser:htmlparser:2.1"
   compile "org.htmlparser:htmllexer:2.1"
   compile "org.apache.velocity:velocity:1.7"
+  compile "org.apache.activemq:activemq-all:5.15.0"
   compile "commons-lang:commons-lang:2.6"
   compile "commons-collections:commons-collections:3.2.2"
   compile "org.json:json:20151123"
   compile "com.googlecode.java-diff-utils:diffutils:1.3.0"
+  compile "javax.jms:javax.jms-api:2.0.1"
   optional "org.apache.ant:ant:1.9.6"
   optional "junit:junit:4.12"
 
--- a/src/fitnesse/components/MessagePublisher.java
+++ b/src/fitnesse/components/MessagePublisher.java
@@ -1,0 +1,48 @@
+package fitnesse.components;
+
+import fitnesse.testrunner.WikiPageIdentity;
+import fitnesse.testsystems.TestPage;
+import fitnesse.testsystems.TestSummary;
+import org.apache.activemq.ActiveMQConnectionFactory;
+import org.json.JSONObject;
+
+import javax.jms.*;
+
+public class MessagePublisher {
+  public static void publishStatusToJMSSend(TestPage sourcePage, TestSummary summary) {
+    try {
+      String textToPublish = getStringifiedMessageToPublish(sourcePage, summary);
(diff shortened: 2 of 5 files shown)


fhoeben:
It seems a bit overkill to add the JMS dependencies to the project just to be able to do this. I understand that this can be quite a feature if you are in a JMS-minded environment, but I believe many users of FitNesse will not be (and this will make all using projects bigger).

Couldn't the same be achieved when executing the test via the jUnit runner (`FitNesseRunner`) and just use a jUnit listener to publish the results (to JMS)?
I use a listener like that to: generate surefire XML reports for each page as it is completed, log suite progress to standard out and to update test results in TeamCity while the suite runs.

See for instance: https://github.com/fhoeben/hsac-fitnesse-fixtures/blob/master/src/main/java/nl/hsac/fitnesse/junit/ProgressLoggerListener.java and https://github.com/fhoeben/hsac-fitnesse-fixtures/blob/master/src/main/java/nl/hsac/fitnesse/junit/JUnitTeamcityReporter.java

suratdas:
I agree with the comment above. 
I created another pull request  ( https://github.com/unclebob/fitnesse/pull/1089 ) with RMI approach instead of JMS and it does not add any additional library. It solves my problem without breaking existing functionality.
Please review that pull request. Thank you for your time and valuable feedback.

amolenaar:
In addition to @fhoeben's comment: it's worthwhile to build FitNesse with as little external dependencies as possible. This avoid conflicts with any dependencies that are part of the SUT (the system under test). That's why it's using "old" technologies like commons-lang and Apache Velocity. Apart from this it's trying to depend on the JDK as much as possible.

I'll close this PR. Looking forward to JMX hooks, though :)

PR #1089, opened 2017-10-04
Publish in-execution test result via RMI

suratdas:
Let's say, Fitnesse suite execution is triggered from standalone program/script/Jenkins. If the execution runs for hours (e.g. Selenium tests), there is no easy way to know the real time execution status before its completion. Real time status will be useful for : 
- Programmatic display of test progress/dashboard.
- Update results in test management system (JIRA/ALM/TFS etc.) using their Rest API. 
- Alert/email stakeholders for specific failure(s) without waiting for hours till suite execution completion.

This pull request uses RMI to publish the results as soon as individual test executes without waiting for full suite execution. Below are the steps to achieve this goal:

1. In Fitnesse root, we have to define the publisher server address
``` 
!define INTERMEDIATE_RESULT_PUBLISHER_SERVER_ADDRESS {//localhost/MyServer}
 ```
2. Ensure RMI server is running in the address specified . The result will be published in below format:
`{"ignores":0,"testPath":"<suitePath.testName>","right":1,"exceptions":0,"testName":"<testName>","wrong":0}`
3. Execute suite with additional query parameter (publish) like below
`http://<fitnesseServerUrl>/<suiteName>?suite&publish`
4. Consume the result published as needed (display/update test management system/alert).

I have tested it and it does not impact the old functionality. New functionality can be achieved by using above steps. 

Improvements (unit test, documentation update, program refactor etc.) can be made in future commits if this idea and approach is approved.


amolenaar:
I think the case for RMI is to specialized. I like the idea of more detailed reporting, but also on the web interface (e.g. feedback per instruction).

PR #1096, opened 2017-11-05
Bugfix/#1095 tests depend on locale

jdufner:
Sets explicitly the locale to US in tests.


fhoeben:
Shouldn't you set the default locale back to previous value after the tests (e.g. in a tearDown)?

jdufner:
Tests set back back the locale to original locale after execution.

PR #1098, opened 2017-11-16
Ensure commons-collections 3.2.2 and commons-lang 2.6 are used

fhoeben:
Ensure commons-collections 3.2.1 is not included as transitive dependency (from velocity)

Since it was not explicitly excluded from velocity I found that my project that depended on fitnesse still got version 3.2.1. Probably because velocity is mentioned before commons-collections.
This change makes that explicit and removes the transitive dependency on 3.2.1, so we get the 3.2.2


fhoeben:
Also ensure commons-lang 2.4, which velocity references, is excluded

PR #1101, opened 2017-11-17
Add plugins page to jar

fhoeben:
The project contains wiki content for a Plugins page, but that page is not distributed.
This pull requests aims to remedy that.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -101,8 +101,8 @@ task createUpdateLists(type: WikiFileListBuilderTask) {
     "FitNesseRoot/FrontPage/properties.xml",
     "FitNesseRoot/PageHeader/content.txt",
     "FitNesseRoot/PageHeader/properties.xml",
-    "FitNesseRoot/Plugins/content.txt",
-    "FitNesseRoot/Plugins/properties.xml",
+    "FitNesseRoot/PlugIns/content.txt",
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -156,7 +158,11 @@ jar {
       include 'FitNesseRoot/FrontPage/**/properties.xml'
       include 'FitNesseRoot/PageFooter/**/content.txt'
       include 'FitNesseRoot/PageFooter/**/properties.xml'
+      include 'FitNesseRoot/PlugIns/**/properties.xml'
+      include 'FitNesseRoot/PlugIns/**/content.txt'
       include 'FitNesseRoot/PageHeader/**/content.txt'
+      include 'FitNesseRoot/Plugins/**/properties.xml'
+      include 'FitNesseRoot/Plugins/**/content.txt'
```
amolenaar: Why is FitNesseRoot/Plug[iI]ns added twice?

fhoeben: What exactly is your question? 
 * I first added it with the wrong casing (lowercase 'i'), and then I corrected that. 
 * Like all others (already present pages) I needed to add a line for both the properties and the content.

fhoeben:
Sorry I didn't read the 'duplicate comment' in proper context first. Of course I intended to remove the lowercase 'i' directory on adding the 'I' version.

PR #1105, opened 2017-11-29
Only fail test run if too few tests are executed

fhoeben:
Currently a jUnit test run is failed if the number of executed tests is different from what is expected.

This pull request ensures that only too FEW tests fail a run, too MANY is logged as a warning.
Furthermore it adds unit tests for this behaviour (the old behaviour was not covered in a test).

### DIFF
--- a/src/fitnesse/junit/JUnitRunNotifierResultsListener.java
+++ b/src/fitnesse/junit/JUnitRunNotifierResultsListener.java
@@ -1,6 +1,8 @@
 package fitnesse.junit;
 
 import java.io.Closeable;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import fitnesse.testrunner.TestsRunnerListener;
 import fitnesse.testsystems.Assertion;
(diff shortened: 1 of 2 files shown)


amolenaar:
What is the case where the number of test cases is not equal to the total number of tests calculated?

fhoeben:
There isn't really any 'normal' case this would happen, of course. BUT when I have a custom test system ('slimCoverage', which doesn't actually run tests but just determines whether scenarios are used, and where) which adds an extra test/page containing the analysis results. This change allows a run with this system to be used (in a CI environment).

I currently use a [workaround](https://github.com/fhoeben/hsac-fitnesse-plugin/blob/5dbbdfd476a3b0e1f53b843531a732b8f1e338d8/src/test/java/nl/hsac/fitnesse/slim/tables/HsacPlugiunFitNesseRunner.java#L65) to achieve the same result. But I figured the real goal of the current implementation is also to ensure missing tests are detected (for instance when a test system crashes, failing to execute all planned tests), so I made this pull request.

amolenaar:
Thanks for the clarification. I suppose you can never run enough tests :)

PR #1107, opened 2017-12-01
A SuiteQuery must neither run test pages that are excluded nor children of excluded pages

NicoleRauch:
As a follow-up to #1079 I explored this further and finally came up with a viable solution.

I refactored quite some code before I actually got to this solution. 
Technically speaking, I followed Separation of Concerns and extracted all but one implementations of the TraversalListener<WikiPage> interface into their own classes which makes the code much less convoluted in these areas. (I do not understand the leftover case - WikiImporter - well enough in order to tackle it right now.)

In hindsight I realised that this refactoring is not strictly necessary for the provided solution, and if you don't like the refactoring as such, I can rebase / reimplement the solution without the refactoring.


src/fitnesse/wiki/PageCrawler.java
```diff
@@ -4,30 +4,189 @@
 
 import fitnesse.components.TraversalListener;
 
-public interface PageCrawler {
-  WikiPage getPage(WikiPagePath path);
+public class PageCrawler {
```
amolenaar: I think this should be left as an interface.

src/fitnesse/wiki/WikiPage.java
```diff
@@ -79,6 +79,8 @@
 
   PageCrawler getPageCrawler();
 
+  PageCrawler getPageCrawler(PagePruningStrategy strategy);
```
amolenaar: I think the interface (which is already quite scattered functionality-wise) should not be extended with such specific methods I think. Wouldn't it be nicer to create an adapter class that implements the pruning strategy on top of the original PageCrawler?

src/fitnesse/wiki/PageCrawler.java
NicoleRauch: The reason why I removed it is that there were the interface and its only implementing class directly next to each other in the same package. Plus, the implementing class did not represent a proper concept of its own (It was simply called PageCrawlerImpl). So I tried to sharpen the concept here.

Can you please explain why you would prefer an interface plus an ...Impl class? (I don't understand this in general, so please bear with my question.)

src/fitnesse/wiki/WikiPage.java
NicoleRauch: I'm not sure I can follow you here.
Your point (not to extend the already broad interface with specific methods) is valid, and I'd like to follow through on this. But I don't really understand your suggestion; can you please elaborate a bit more?

Also, I'm wondering whether this is indeed a specific method -- isn't the ability to prune a search tree part of the "core business" of the crawler? Following this thought further, maybe it would be a better solution to remove the getPageCrawler() method without arguments and implement a NoPruning strategy, so that these two ways of handling the crawling can be unified here, in the WikiPageFinder and in the PageCrawler itself?
I'd appreciate your thoughts on this.

amolenaar: > ...can you please elaborate a bit more?

My hunch is to do something like

    crawler = new PruningCrawler(wikiPage.getPageCrawler())

`PruningCrawler` would act as a filter, e.i. it would remove all pruned pages.

> Also, I'm wondering whether this is indeed a specific method -- isn't the ability to prune a search tree part of the "core business" of the crawler?

I suppose it is :). How about if the PageCrawler would have a method `traverse` to which you can provide a strategy? At least that would move the concern of traversing/pruning down into the page crawler and the wiki page itself is not bothered with it?

 Adding an additional argument to `WikiPage.getPageCrawler()` would only make the interface more complicated, so I would not be in favour of this.

NicoleRauch: > PruningCrawler would act as a filter, e.i. it would remove all pruned pages.

The problem with this approach is that it is not efficient. When looking at any given page in isolation, how would I be able to figure out whether it needs to be removed? If the page itself is pruned, that's easy; but if it isn't, I have to follow the parent pages up to the root page to decide whether one of the parent pages is pruned. This gives us an exponential algorithm (in the worst case), I believe, and I'm not a big fan of this. That's why I tried to follow the other approach: not traverse the skipped page tree parts in the first place.

I will look into moving the strategy parameter from the getPageCrawler() method to the crawler's traversal. Makes more sense altogether, I guess. Thanks for the suggestion!

NicoleRauch:
@amolenaar I've committed the changes, and I like this much better now. Thanks for the suggestion! Are you ok with this change?

One last question: There are now two `search()` methods in the `WikiPageFinder` class (one with strategy and one without), and one could argue that this is an unnecessary widening of the interface as well. OTOH, unifying the two `search()` methods would make the pruning strategy much more visible (it would e.g. appear in all search methods, most of which would simply use the `NoPruningStrategy`. What are your thoughts on this?

PR #1112, opened 2017-12-14
Improve copying content of standard 'FitNesseRoot content'

### LINKED ISSUE #1111: Plugins and 'root' pages not extracted on startup
When starting fitnesse (from standalone.jar) it checks version and extracts 'standard pages', if they are missing.

But the jar's Resources/FitNesseRoot directory contains more files than are extracted (specifically `context.txt` and `properties.xml`, and a `Plugins` folder). These files are not extracted, and are therefore not available in the wiki. I believe they should be.

This is probably because these files are not listed in `Resources/updateList` or `Resources/updateDoNotCopyOverList`...

fhoeben:
Ensure that all pages below the project's FitNesseRoot are included in the distribution and are also extracted on first startup of the wiki. 
Fixes #1111.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -156,6 +156,7 @@ jar {
   dependsOn createUpdateLists
   into('Resources') {
     from('.') {
+      setIncludeEmptyDirs(false)
       include 'FitNesseRoot/**/content.txt'
       include 'FitNesseRoot/**/properties.xml'
       include 'FitNesseRoot/**/*.wiki'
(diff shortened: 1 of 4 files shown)


amolenaar:
What I really want is just add the files that are under version control (e.i. `git ls-files`). With the PR merged, when I build a release from my working directory, it will also include the `ErrorLogs` and my `RecentChanges`.

fhoeben:
I don't quite know how to get the list of files under version control. But we could quite easily ignore `ErrorLogs` and `RecentChanges`. 

Based on the `.gitignore` I believe we should exclude:
```
FitNesseRoot/ErrorLogs/*
FitNesseRoot/files/testResults/*
FitNesseRoot/RecentChanges*
```

Would skipping those make it ok?

PR #1113, opened 2017-12-21
Add ability to directly reference values in a hashtable like slim symbols

tcnh:
Needs some polishing, would like some input on how to proceed.

What this PR achieves is that we can reference values from a hashtable directly in a slim table, using a dot notation on slim symbols.

Say we define a hashtable !{EN:!{flag:flag, cup:cup}, NL:!{flag:vlag, cup:beker}} and store it in a symbol $myMap
With this PR, we can directly reference values in the hashtable as if they were slim symbols:
```
$myMap.EN.flag returns 'flag'
$myMap.NL.cup returns 'beker'
```

At this point with this PR this functionality is usable, but the wiki doesn't show the actual value in the usual form: $myMap.EN.flag->[flag]. Instead it just display: $myMap.EN.flag

I've also added an example in an acceptance test. This renders the hashmap and then appends .result.

Any input is appreciated.

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/HashTableTests/ReferenceHashTableValueFromSlimSymbol.wiki
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/HashTableTests/ReferenceHashTableValueFromSlimSymbol.wiki
@@ -1,0 +1,7 @@
+---
+Test
+---
+|script     |echo fixture                             |
+|$myHashMap=|echo|!{key1:value1,key2:!{subkey:subval}}|
+|check      |echo|$myHashMap.key1           |value1   |
+|check      |echo|$myHashMap.key2.subkey    |subval   |
--- a/src/fitnesse/slim/SlimSymbol.java
+++ b/src/fitnesse/slim/SlimSymbol.java
@@ -5,11 +5,11 @@ import java.util.regex.Pattern;
 
 public abstract class SlimSymbol {
   public static final Pattern SYMBOL_PATTERN = Pattern
-      .compile("\\$([A-Za-z\\p{L}][\\w\\p{L}]*)");
+      .compile("\\$([A-Za-z\\p{L}][.\\w\\p{L}]*)");
 // This would be a better pattern as it allows to define the end of a symbol name with another $ sign
-// but this could break existing tests. See discussion in #790  
+// but this could break existing tests. See discussion in #790
 //  public static final Pattern SYMBOL_PATTERN = Pattern
-//      .compile("\\$([A-Za-z\\p{L}][\\w\\p{L}]*)\\$?");  
+//      .compile("\\$([A-Za-z\\p{L}][\\w\\p{L}]*)\\$?");
   public static final Pattern SYMBOL_ASSIGNMENT_PATTERN = Pattern
       .compile("\\A\\s*\\$([A-Za-z\\p{L}][\\w\\p{L}]*)\\s*=\\s*\\Z");
 
@@ -26,14 +26,14 @@ public abstract class SlimSymbol {
 
 
 
-  
+
   public String replace(String s) {
     if(null == s) return null;
 
     // Don't replace assignments, return as is
     if (isSymbolAssignment(s) != null)
       return s;
-    
+
     replacedString = s;
     symbolMatcher = SYMBOL_PATTERN.matcher(s);
     replaceAllSymbols();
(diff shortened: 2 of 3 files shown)


amolenaar:
This sounds like a reasonable addition.

The "problem" (trick) with slim symbols is that you'll have to do some double bookkeeping: the values need to live both on the slim server and inside the wiki server.

Also, how should it behave in case of an invalid key? My guess would be that it will just leave that bit of text as is, it may be something like a URL.

* `$myMap.NL.chair` would result in `{flag:vlag, cup:beker}.chair`? (how are maps displayed anyway?)
*  `$myMap.NL.cup.example.com` would result in `beker.example.com`

ggramlich:
Another point to remember: this is just the Java specfic implementation.
Every other slim executor will have to be adapted or just behave differently.

jediwhale:
This will be an increment to the Slim protocol version number.

tcnh:
> The "problem" (trick) with slim symbols is that you'll have to do some double bookkeeping: the values need to live both on the slim server and inside the wiki server.

I was looking into this. If I understand it right, the wiki server only knows the String representation of symbols and has no means to get values from the slim server side, right?

In my view, it now functions as I'd expect on the slim server side now, so I should:
- Create some way to parse the hashtable pattern to hashmap on the wiki server side
- Replace the slim symbol if the wiki knows it (Fixtures implementing statementexecutionerinterface may also create map symbols on the slim server side, that the wiki doesn't know about?) and find a consistent way to display references that don't exist

right?

fhoeben:
I understand the desire to get values from a hashtable but I don't really know whether special handling for a hashtable in a slim symbol is what is needed. What I think you've run into is the desire to evaluate expressions on a slim symbol's value. For now you want to get a (nested) key from a table, but why not an element in a list/array or add 1 to an integer? 

This also raises an issue of syntax, which I think is related to the 'missing key' issue that @amolenaar raised. Slim symbols are currently already not very nice, as there is no way to explicitly end the symbol's name when using it (FitNesse just uses the longest possible match, which might be ambiguous if you want to concatenate some text to the symbol's value).
I would suggest that when you support this more powerful handling of slim symbols you also introduce more explicit syntax that defines where your expression (e.g. index/key in the hashtable) ends and what text should be added to it. This will allow an error to be raised, or null/empty string(?), to be returned for an unknown key, but also support concatenating some more text to the returned value. 

Evaluating such expressions might be done in the wiki (which currently only knows the string value so the symbol's value should be parsed first) or in the slim server side, which probably requires a protocol update.

The double bookkeeping of symbols poses some additional complexity when you want to address this on the wiki side. The string representation known to the wiki might be 'out of date' (i.e. no longer be in sync with the value stored in the slim server). For instance when the hashtable gets additional keys or values are updated via fixture code without a new string representation being returned to the wiki. Or even worse: the wiki is completely unaware of a symbol's value if it was assigned on a different page (e.g. a SuiteSetUp) but the slim executor can still use it. 

Evaluating on the server side (like you did in this pull request) allows the expressions to work on all possible symbol value objects directly, without the need of the wiki to first parse them. This seems especially convenient when using another language than Java or with custom classes being used. But it does mean the expression's result can not be displayed directly in the wiki (which I don't consider a very big issue). And indeed it has to be implemented for each slim servers/executor (possibly leading to different interpretations of the expression) which is not very nice but seems unavoidable.   

This seems like quite a can of worms to open. But I like the options it offers....

For Java, a lot might actually already be possible via a custom `FixtureInteraction` subclass and/or fixtures implementing `StatementExecutorConsumer` and/or implementing `InteractionAwareFixture`.

fhoeben:
I thought about it a bit more and I realised that providing a value back to the wiki **is** important so that the value in a hashtable can also be used as an expected value. So your test should be extended to:
```
|script     |echo fixture                                      |
|$myHashMap=|echo|!{key1:value1,key2:!{subkey:subval}}         |
|check      |echo|$myHashMap.key1       |value1                |
|check      |echo|$myHashMap.key2.subkey|subval                |
|check      |echo|value1                |$myHashMap.key1       |
|check      |echo|subval                |$myHashMap.key2.subkey|
```

Furthermore I looked a bit closer at the code and I think you added the handling of the 'dot notation' at a strange place in the `VariableStore`. You updated `VariableStore.getStoreSymbolValue()` to handle the nested periods. This method is only used from `replaceSymbolsInString()` which is a bit strange given that this is a direct variable reference, and not a string value containing a variable. 
I would suggest you update `containsValueFor()` and `getStored()` so that these methods also resolve the values inside the hashtable.

tcnh:
I'll close this PR. @fhoeben has come up with a better idea. Will create a new PR when that's fully working

PR #1117, opened 2017-12-29
Update FitNesse for Java 9

amolenaar:
- [X] Support the `plugins/` folder for Java 9
- [X] Fix XML serialization for Java 9
- [X] Fix test cases (where dealing with sets and maps)
- [X] Plugin jars are no longer added to the `java.class.path` system property


src/fitnesse/wiki/fs/PageXmlizer.java
```diff
@@ -43,7 +43,7 @@ public void deXmlizeSkippingRootLevel(Document document, WikiPage context, Xmliz
   public Document xmlize(PageData data) {
     Document document = XmlUtil.newDocument();
     Element dataElement = document.createElement("data");
-    XmlUtil.addCdataNode(dataElement, "content", data.getContent());
+    XmlUtil.addTextNode(dataElement, "content", data.getContent());
```
fhoeben: Why is this no longer a 'cdata' block? This probably means the content will be xml escaped, which it was not before. Will that not make it harder to edit the content as 'plain text'? Can it not remain cdatan for maximum backward compatibility?

test/fitnesse/components/PluginsClassLoaderTest.java
```diff
@@ -16,66 +16,29 @@
...
-    pluginsClassLoader.addPluginsToClassLoader();
+    new PluginsClassLoader().loadPlugins("nonExistingRootDirectory");
+
     assertTrue("didn't cause exception", true);
   }
 
   @Test
   public void addPluginsToClassLoader() throws Exception {
     String[] dynamicClasses = new String[]{"fitnesse.testing.PluginX", "fitnesse.testing.PluginY"};
     //todo This fails because some other test probably loads plugin path    assertLoadingClassCausesException(dynamicClasses);
```
fhoeben: This todo line can be removed, as the `assertLoadingClassCausesException` method no longer exists

src/fitnesse/components/PluginsClassLoader.java
```diff
@@ -2,45 +2,54 @@
...
+          urls.addAll(toUrls(plugin.getCanonicalPath()));
+
+    return urls.toArray(new URL[urls.size()]);
   }
 
-  public static void addItemsToClasspath(String classpathItems) throws Exception {
+  private static List<URL> toUrls(String classpathItems) throws Exception {
     final String separator = File.pathSeparator;
     String currentClassPath = System.getProperty("java.class.path");
     System.setProperty("java.class.path", currentClassPath + separator + classpathItems);
```
fhoeben: Shouldn't this reading/parsing/updating of system property be in a different method? (Called after generating the full list of urls.)
Actually updating the property should probably only be done in once the class loader is created and made active. Should it be moved to `ClassUtils.setClassLoader()`, if it is called with a URLClassLoader it could call `getURLs()` and update the property with that result?

src/fitnesse/components/PluginsClassLoader.java
```diff
@@ -2,45 +2,54 @@
...
-  public void addPluginsToClassLoader() throws Exception {
-    if (pluginsDirectory.exists())
+  private static URL[] urlsForPlugins(File pluginsDirectory) throws Exception {
+    List<URL> urls = new ArrayList<>();
+
+    if (pluginsDirectory.exists() && pluginsDirectory.isDirectory())
       for (File plugin : pluginsDirectory.listFiles())
         if (plugin.getName().endsWith("jar"))
-          addItemsToClasspath(plugin.getCanonicalPath());
+          urls.addAll(toUrls(plugin.getCanonicalPath()));
```
fhoeben: Each plugin file is a single element, right? So why not call `toUrl()`, and add that one url the list. The updating of the system property does not have to be done for each plugin found, but can be postponed to after full url list is assembled (and has items).

src/fitnesse/components/PluginsClassLoader.java
```diff
@@ -2,45 +2,54 @@
...
+import java.util.List;
 
 
+/**
+ * Update the current thread class path with jars foundin a "plugins" directory.
+ */
 public class PluginsClassLoader {
 
-  private File pluginsDirectory;
+  public static ClassLoader loadPlugins(String rootPath) throws Exception {
```
fhoeben: This method should probably be renamed to reflect the fact that it doesn't actually load anything, but gets a class loader.

src/fitnesse/components/ComponentFactory.java
```diff
@@ -76,12 +82,26 @@ public ComponentFactory(Properties properties) {
...
+    try {
+      clazz = lookupComponentClass(className);
+    } catch (ClassNotFoundException e) {
+      throw new ComponentInstantiationException("Unable to load class named " + className, e);
+    }
+    return createComponent(clazz);
+  }
+
+  public <T> Class<T> lookupComponentClass(String className) throws ClassNotFoundException {
+    return (Class<T>) classLoader.loadClass(className);
```
fhoeben: Should this not be delegated to `ClassUtil` also? It seems a bit strange that the class loader is maintained both in that util singleton and this class. It would be nicer if all class loader access is done in a uniform way.

src/fitnesse/components/ComponentFactory.java
```diff
@@ -24,10 +24,16 @@
 public class ComponentFactory {
 
   private final Properties properties;
+  private final ClassLoader classLoader;
   private Map<String, Object> components;
 
   public ComponentFactory(Properties properties) {
+    this(properties, ClassLoader.getSystemClassLoader());
```
fhoeben: Shouldn't this get its default class loader from `ClassUtils`?
Actually, if this class needs/has its own class loader reference it should probably not have a constructor that does not supply a class loader.

src/fitnesse/wiki/fs/PageXmlizer.java
amolenaar: In Java 9, the default XML transformation will put a CDATA section on a new line. When parsing the content, the parsed text does include extra newlines and spacing. This is not what we want.
Just using text avoids this.

This part is only used for the SerializationResponder (the proxy endpoint). Nowhere else. The page history is recorded as normal escaped text already and that's been working for years already.

src/fitnesse/components/PluginsClassLoader.java
amolenaar: Yes. I actually thought of removing this line altogether (no more system properties updating).

src/fitnesse/components/ComponentFactory.java
amolenaar: There are a few places deep in the FitNesse code that do Class.forName magic.

Preferable I want to get rid of the ClassUtils class, since it's introducing a static variable. For now I want to leave it like this, because working the ClassUtils.forName out of those pieces of code is a huge endeavour. I should probably put some documentation describing my intent on the `ClassUtils` class.

src/fitnesse/components/ComponentFactory.java
amolenaar: As said above, I'd like to get rid of `ClassUtils` altogether.

fhoeben: Maybe the static method can be moved to `ComponentFactory`, removing `ClassUtils`? Because it seems that already has the responsibility to load stuff.

src/fitnesse/ContextConfigurator.java
```diff
@@ -66,6 +67,7 @@ public static ContextConfigurator empty() {
   public static ContextConfigurator systemDefaults() {
     return empty()
       .withRootPath(DEFAULT_PATH)
+      .withClassLoader(ClassLoader.getSystemClassLoader())
```
fhoeben: I just figured out that this is the class loader used when created for jUnit test. This means the class loader can no longer be changed for jUnit tests by a runner (which is what `FitNesseRunnerExtensionTest.createContext()` is trying to demonstrate).

Can we either add a method `ClassUtils.getClassLoader()` and use that here instead of directly using the system one, or provide some what to supply a class loader from the jUnit runner which subclasses could update, so the class loader created by `PluginsClassLoader.loadPlugins()` is used

src/fitnesse/components/ComponentFactory.java
amolenaar: That would not solve the problem of having *mutable* static variables. The fact that we need those points to a design flaw.

src/fitnesse/components/PluginsClassLoader.java
amolenaar: I'm fine with this method. It loads them into a ClassLoader.

src/fitnesse/components/PluginsClassLoader.java
amolenaar: Fixed.

src/fitnesse/components/PluginsClassLoader.java
amolenaar: Fixed (removed this method)

test/fitnesse/components/PluginsClassLoaderTest.java
amolenaar: Fixed.

test/fitnesse/junit/FitNesseRunnerExtensionTest.java
```diff
@@ -41,9 +43,12 @@ protected String getSuiteName(Class<?> klass) throws InitializationError {
 
     @Override
     protected FitNesseContext createContext(Class<?> suiteClass) throws Exception {
-      new PluginsClassLoader(getFitNesseRoot(suiteClass)).addPluginsToClassLoader();
+      ClassLoader classLoader = new PluginsClassLoaderFactory().getClassLoader(getFitNesseRoot(suiteClass));
+      ClassUtils.setClassLoader(classLoader);
```
fhoeben: This line is not needed right? `makeFitNesseContext()` called below already ensures ClassUtils is properly set.

fhoeben:
In understand that passing/setting the class loader for in process runners is good for backward compatibility. But is it actually desirable?
When running in a different process the class loader is not changed (i.e. the plugins are not on the classpath). Why should they be when running in-process? Now that there is a special class loader it seems better to make the in- and out-process runners behave more alike (i.e. work without the plugins on the class path).

amolenaar:
Indeed it's when stuff is run in-process. It's no different than running tests through the JUnit runner (which takes plugins into account).

PR #1123, opened 2018-01-30
Add ability to evaluate expressions in/on slim symbols

tcnh:
This PR adds the ability to evaluate expressions using Slim in a slim-symbol like syntax ($`expression`).
Based on the use case where it would be desireable to be able to directly reference values from a hashmap (https://github.com/unclebob/fitnesse/pull/1113), but extended with the use of a javascript expression evaluator by @fhoeben 

We can now for example evaluate expressions like:
```
|$myHashMap=|echo|!{key1:value2,key2:!{subkey:subval1}}|
|check      |echo|$`myHashMap.key1`  |value2           |
|check      |echo|$`myHashMap.size()`|3                |
```

or:

```
|$myVar=|echo|this is a text                       |
|check  |echo|$`myVar.toUpperCase()`|THIS IS A TEXT|
```
Implementation is done on the runner and on the wiki side of things, so we can both use the expressions' value as input and as expected outcome.


six42:
Hi Tom, 
this is an awesome extension. Thanks for this.
The design of Slim is to store symbols in the slim server and in the slim client.
Your implementation of expressions follows this design. 
As a result every expression  must be evaluated once in the server and once in the client.
In my opinion this unnecessarily limits the usability of your implementation of expressions:
 * It will not work for any Slim Client not based on the JVM. (.Net, python, c++)
 * Even for slim clients using the JVM it reaches quickly limitations. See an example below.

How about changing the implementation to execute the expressions only in the Slim server? 
 * This would bring the feature directly to all Slim clients.
 * It should serve the purpose of most uses cases. 

The other option to execute it in the Slim Client I see as less important as fixture code can be written to achieve the same. 

What do you think?

------------------------------------
## Example where evaluation fails  in JVM 

|import            |
|fitnesse.slim.test|

|script  |Test Slim With Converter                      |
|$myList=|getList                                       |
|start   |echo fixture                                  |
|check   |echo|$`myList.size()`    |3                   |
|check   |echo|$`myList.getClass()`|$`myList.getClass()`|

grafik

fhoeben:
@six42 the whole point of this PR was that evaluation was done on the server side, as that allows all methods to be executed on the actual objects, instead of on the 'String representation' which is the only thing that exists on the client side.
The problem you're experiencing has to do with the Java implementation of the Slim protocol, which has special handling for java.util.List (which is what `getList()` returns). If you were to call a method which returned java.util.ArrayList you would see that it would work as expected.

tcnh:
I do think you found an omission. No conversion was done on the string representation lists. I added that in the last commit and added your example to the acceptance test page.

All evaluation is done by the new SlimExpressionEvaluator class. Wiki representations of objects are converted if necessary (Hashtable, and now List) before an expression is evaluated.

I have no experience with non-Java slim clients/implementations, so I can't estimate the amount of work it would take to create/port similar behaviour in these clients.

six42:
Hi Tom,

thanks for taking a second look at my comment. We are coming closer to  a common understanding.
What you identified as an omission is in my view a design limitation of Slim which your PR inherits.
You worked around it by adding a list parser to the Slim Server. Like you added a parser for hash maps in your initial PR.

How about the following example:


!define SLIM_FLAGS {-v}

|import            |
|fitnesse.slim.test|

|script|  Test Slim                     |42|
|$ts=|getFixture                                       |
|start   |echo fixture                                  |
|check   |echo|$`ts.returnConstructorArg()`    |42                   |

Again it will display an error message in the server and works nicely in the system under test ( returning 42).
grafik

Fixing the display problem by building another parser in the Slim Server is not a solution. For a general pruepose solution you would need to build a parser for any potential class used in the system under test.
Therefore my initial question: what is your main use case?
1 Running arbitrary commands on arbitrary objects in the system under test
2 or having support to evaluate functions on a limited number of object classes (hash maps, list and strings) in the Slim Server?

Both have sex appeal. My feeling is the second option is more useful. 
The first option you can achieve with standard fixture code in most cases.

if you opt for option 2 then a potential implementation should evaluate all $`...` occurrences somewhere in the makeAssertion function or in this area. 
The slim client would never see a $`...`. And this has the additional advantage that it would work out of the box for all slim client implementations.

You are likely aware that you see in the execution log what has been send to the client when the SLIM_FLAGS (-v) has been defined

```
------
[scriptTable_43_3, call, scriptTableActor, echo, $`ts.returnConstructorArg()`]

[scriptTable_43_3, 42]
------
```

tcnh:
Hehe, option 2. But also on the client side.

One of the main use cases would be to define testdata in a hashmap and then reference it using expressions. That saves at least half the wiki lines compared to letting a fixture do it in a scenario.

ybermejo:
Hi @tcnh,

Does this new feature can compute exponential expression? Thanks

fhoeben:
@ybermejo this feature could do any computation you can express in Javascript, as I understand it. What exactly are you looking for/thinking of?

six42:
Hi Fried,

I agree this PR adds value.
The only trouble I see is that one day if an implementation without the mentionend limitations would be done 
then the nice and common semantic 
$`...` is already used and another syntax must be invented. Luckily there are many special characters on a keyboard :)


Please go ahead with the merge.
Thanks again to Tom for getting this coded and all the other great contributions you made to FitNesse.

Issue #1124, opened 2018-02-13
Custom theme based on bootstrap no longer found using 20180127


tcnh:
I have a custom theme that extends on bootstrap. Since 20180127, the theme files are no longer found.

Before 20180127 I could have a jar with the theme files in plugins and just define the theme in plugins.properties. This no longer works.

Is there a new way to make this type of themes available in the latest version?
Custom responder plug-ins  for example continued to work just fine when placed in the plugins folder.

tcnh:
I believe this happens because plugins now have a dedicated classloader, but velocity only knows the system classloader (that no longer holds any reference to plugins/themes). Should velocity get a hold of the plugins classloader?

fhoeben:
Is this now fixed with the 20190127 release, which includes the classloader and javascript expression PRs you made?

tcnh:
Yes. fixed in 20190127

PR #1127, opened 2018-02-20
Fix velocity templates from jar not loading

tcnh:
Fix that solves the problem that since the new PluginsClassLoader was implemented, velocity templates from jar files were no longer usable.
This PR fixes that by setting the current thread's context classloader to Fitnesse's extended classloader containing all of the plugin Jars and making the fileresponder use that classloader instead of getting its classloader from the current class.

Related issue: https://github.com/unclebob/fitnesse/issues/1124

### DIFF
--- a/src/fitnesse/responders/files/FileResponder.java
+++ b/src/fitnesse/responders/files/FileResponder.java
@@ -74,8 +74,18 @@ public class FileResponder implements SecureResponder {
       return createNotModifiedResponse();
 
     String classpathResource = "/fitnesse/resources/" + resource.substring("files/fitnesse/".length());
+    InputStream input;
 
-    InputStream input = getClass().getResourceAsStream(classpathResource);
+    input = Thread.currentThread().getContextClassLoader().getResourceAsStream(classpathResource);
+
+    if (input == null) {
+      //remove leading slash so path will work with resources inside a JAR file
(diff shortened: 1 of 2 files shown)


src/fitnesse/responders/files/FileResponder.java
```diff
@@ -74,8 +74,18 @@ private Response makeClasspathResponse(FitNesseContext context, Request request)
...
     String classpathResource = "/fitnesse/resources/" + resource.substring("files/fitnesse/".length());
+    InputStream input;
 
-    InputStream input = getClass().getResourceAsStream(classpathResource);
+    input = Thread.currentThread().getContextClassLoader().getResourceAsStream(classpathResource);
+
+    if (input == null) {
+      //remove leading slash so path will work with resources inside a JAR file
+      while (classpathResource.startsWith("/"))
+      {
```
fhoeben: This brace should really be on the previous line...

fhoeben:
@tcnh any chance you can add a unit test to ensure templates can be loaded from jars in the plugins directory? Its one of those things that are easily to miss when making a change.

tcnh:
> any chance you can add a unit test to ensure templates can be loaded from jars in the plugins directory?

I tried, but I can't seem to figure out how to load a resource for testing... :(

Issue #1152, opened 2018-08-31
FitNesseRunner lacks an option to override variables


micw:
When running with fitnesse-standalone, it is possible to override variables by adding url params. Internally this is done by parsing the params and creating an UrlPathVariableSource from it. Later during variable resolution, UserVariable source checkis if the page variable source is of type UrlPathVariableSource, and if so, it is used with higher priority than the other variable sources. This is an implementation that is full of implicit assumptions - maybe this should be re-written some day.

When using FitNesseRunner, there's no option to override variables this way. The reason is that FitNesseRunner does not call context.getRootPage() with a map of values.
Since all those calls are in private methods, there's also no way to create a custom FitNesseRunner that calls the method with Map arguments.

I propose to
a) refactor FitNesseRunner to allow overriding this behaviour
b) define a proper way to set custom parameters (e.g. provide a class that does it by using a annotation)

fhoeben:
As a suggestion to 'workaround' this issue you could consider using my plugin (https://github.com/fhoeben/hsac-fitnesse-plugin, sample usage https://raw.githubusercontent.com/fhoeben/hsac-fitnesse-fixtures/master/wiki/FitNesseRoot/HsacExamples/SymbolSamples.wiki). 

It defines a `!defineDefault` extension to the wiki language that allows you to define a variable that may be overridden by system properties. Alternatively it defines `!defineDefaultFromProperties` that allows the variables to be overridden using a .properties file.

I use this to override variables when running tests (with the jUnit FitNesseRunner) on build server.
This way the jUnit class also does not have to contain the values to use during the test run, but they can be specified externally. Specifically I use it to point my tests to the correct server.

woodybrood:
The hsac-fitnesse-plugin option also should allow for -D vm arguments at the command line. I'm going to close this with a viable workaround.

PR #1153, opened 2018-08-31
Feature/1152

micw:
Annotation based configuration of a provider class for custom variables. Solves https://github.com/unclebob/fitnesse/issues/1152

### DIFF
--- a/src/fitnesse/junit/CustomVariableProvider.java
+++ b/src/fitnesse/junit/CustomVariableProvider.java
@@ -1,0 +1,12 @@
+package fitnesse.junit;
+
+import java.util.Map;
+
+/**
+ * Provide custom variables to override page variables in a similar way as done with URL-Parameters
+ * when using the fitnesse-stabdalone runner
+ * @author mwyraz
+ */
+public interface CustomVariableProvider {
+	public Map<String,String> getCustomVariables();
+}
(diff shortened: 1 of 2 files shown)


six42:
Hi Michael,
thanks for contributing.
Your pull request looks similar to #978 

Can you update the junit overview page to mention this new feature:
http://localhost:8001/FitNesse.UserGuide.WritingAcceptanceTests.RunningFromJunit
and add a junit test which uses it and serves as documentation and test at the same time.
Ideally write a test which defines at least two variables.

src/fitnesse/junit/FitNesseRunner.java
```diff
@@ -416,14 +444,21 @@ protected void addExecutionLogListener(RunNotifier notifier, MultipleTestsRunner
     testRunner.addExecutionLogListener(new ConsoleExecutionLogListener());
   }
 
+  protected WikiPage getRootPage() {
+	  if (this.customVariables==null) {
```
fhoeben:
```suggestion
	  if (this.customVariables == null) {
```

src/fitnesse/junit/CustomVariableProvider.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.Map;
+
+/**
+ * Provide custom variables to override page variables in a similar way as done with URL-Parameters
+ * when using the fitnesse-stabdalone runner
+ * @author mwyraz
+ */
+public interface CustomVariableProvider {
+	public Map<String,String> getCustomVariables();
```
fhoeben: Should this method also receive the suiteClass?

PR #1178, opened 2019-01-17
Fix handling of decision table with unknown output parameter

### LINKED ISSUE #1177: Unknown output column in decision table is ignored
When a decision table based on a scenario has an unknown input column an exception is thrown and the test result is error. When an output column (name ends with '?') is unknown this column is ignored. And the test may still pass. This is undesirable as tests that are expected to be executed are skipped without this being clear in the overal test outcome. 

Example (the scenario 'my division' has no output column 'missing' but the test passes):
```
|scenario      |my division       |numerator||denominator||my result2?|
|setNumerator  |@numerator                                            |
|setDenominator|@denominator                                          |
|$myResult2=   |quotient                                              |

Get the Division implementation from the eg package
|Library    |
|eg.Division|

|my division                               |
|numerator|denominator|my result2?|missing?|
|10       |2          |5.0        |a       |
```
screenshot 2019-01-17 at 07 19 24

fhoeben:
Fixes #1177 

screenshot 2019-01-17 at 07 12 49

### DIFF
--- a/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
+++ b/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
@@ -1,5 +1,6 @@
 !2 ${FITNESSE_VERSION}
  * Issues fixed:
+   * Raise error if decisions table (based on scenario) has output column which is not an output of the scenario ([[1178][https://github.com/unclebob/fitnesse/pull/1178]])
    * fitnesse.jar was too big (bigger than fitnesse-standalone.jar) because of duplicated files
    * Build upgraded to Gradle 5.1
 
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/ScenarioTestSuite/StopTestExceptionInDescisionTable.wiki
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/ScenarioTestSuite/StopTestExceptionInDescisionTable.wiki
@@ -1,0 +1,37 @@
+---
+Test
+---
+A !-StopTestException-! should cause the rest of the table to be skipped.
+
+!| script                       |
+|given page|MySuite|with content|${SUT_PATH} !-
+!define TEST_SYSTEM {slim}
+-! |
+|given page|MySuite.TestPageA|with content|!-
+|import                                   |
+|fitnesse.slim.test                       |
+
+|scenario      |my division _     |numerator,denominator,my result3?|
+|setNumerator  |@numerator                                          |
+|setDenominator|@denominator                                        |
+|start         |constructor throws|stop test                        |
+|$myResult3=   |quotient                                            |
+
+Get the Division implementation from the eg package
+|Library    |
+|eg.Division|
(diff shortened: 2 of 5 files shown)


tcnh:
There seems to be a side effect (possibly intentional?) to this change.
Prior to this change, we could define a scenario as follows:

```
| scenario     | my scenario    |
| action                        |
| $returnValue=| something else |
```
And call it as a DT like:
```
| my scenario  |
| returnValue? |
| expectation  |

```
This now throws the new exception. We now need to specify the return variable in the scenario signature. So to continue using our DT's we now need to refactor the scenario to a more verbose:
```
| scenario     | my scenario    | | returnValue? |
| action                                         |
| $returnValue=| something else                  |
```
Not a big issue, but might break some tests (for example BrowserTest.ScenarioTest in the HsacExamples).

fhoeben:
This is not intentional, but an expected side effect (as you may have noticed I did update some tests). Scenario's already required explicit input parameters and supported explicit output parameters. The fact that specifying the output was optional was unexpected and could lead to strange situations, as #1177.
The current check for whether an output parameter is present could possibly be improved, but erring on the side of expected behaviour and caution seemed wise in this case. Any suggestion (PR?) is more than welcome.

When using hsac-fitnesse-plugin: `table template` does not require explicit in- or output parameters. so in your case just replacing 'scenario' with 'table template' might be an easier fix than adding the explicit output parameter to the scenario's definition.

maikelgithub:
We are experiencing serious issues since this change. All of our tests are now failing...

How does this work when having a scenario in a table template?

The goal of our tests is to fill a variable, so it's not part of any input.

All our tests now fail with the output:

> Bad table! The argument _yourvariable_ is not an output of the scenario.

fhoeben:
@maikelgithub can you give a sample wiki page with the scenario/table template combination that shows the issue you are experiencing?

maikelgithub:
Hi @fhoeben, 

thank you for your fast response.

We used to have similar code to this:

```
|scenario      |my division|numerator|and|denominator|
|setNumerator  |@numerator                           |
|setDenominator|@denominator                         |
|$myResult2=   |quotient                             |

|Library    |
|eg.Division|

|table template|test    |
|my division   |10|and|2|

|test       |
|my result2?|
|           |
```

This was still working on version 20190224.

With the new update, this is what I'm trying:
```
|scenario      |my division|numerator|and|denominator|makes|myResult2?|
|setNumerator  |@numerator                                            |
|setDenominator|@denominator                                          |
|$myResult2=   |quotient                                              |

|Library    |
|eg.Division|

|table template|test                     |
|my division   |10|and|2|makes|myResult2?|

|test      |
|myResult2?|
|          |
```

We're considering to keep working on version 20190224 for now. We have 100's of tests to adjust if we want to use the update of 20190409, while we never encountered issues with our scripts prior to the February update.

fhoeben:
I have not tried this combination. It works if the table template is converted to a normal scenario.
Can you file an issue on https://github.com/fhoeben/hsac-fitnesse-plugin, I believe the bug is in table template's code

Issue #1141, opened 2018-06-11
How to make FitNesse only listen to localhost? And how to use the Shutdown java class


maikelgithub:
Hi, 

for my projects I'm looking to use the localhost binding for security reasons. However, I can't seem to get the right command to start FitNesse with the -h command, I know it's not supported in the Reference Guide. It is defined in the Shutdown.java file located [here](https://github.com/unclebob/fitnesse/blob/master/src/fitnesse/Shutdown.java) at the bottom of the page.

The command to start FitNesse:
java -jar fitnesse-standalone.jar

The command I'm looking for:
java -jar fitnesse-standalone.jar -h localhost

A similar patch has been proposed some time ago, see [this link](https://sourceforge.net/p/fitnesse/patches/1/).

PR #1183, opened 2019-02-08
Added -lh flag / LocalhostOnly property to just bind to the local loopback interface

tcnh:
Added -lh flag / LocalhostOnly property to just bind to the local loopback interface.
Effectively preventing wiki-access from other machines in the newtork.

Should address https://github.com/unclebob/fitnesse/issues/1141

Note: I am unable to test this change with SSL enabled

### DIFF
--- a/FitNesseRoot/FitNesse/UserGuide/AdministeringFitNesse/CommandLineArguments/content.txt
+++ b/FitNesseRoot/FitNesse/UserGuide/AdministeringFitNesse/CommandLineArguments/content.txt
@@ -15,7 +15,7 @@ A subset of configuration options can be provided through the command line. Comm
 | {{{-c <command>}}} | Command | ''none'' | Run a Rest Command and then exit. You can find more information on the -c options at RestfulServices. |
 | {{{-b <filename>}}} | !-RedirectOutput-! | ''stdout'' | redirect command output. This is most useful in conjunction with the ''-c'' option |
 | {{{-f <config file>}}} | n.a. | plugins.properties | Load an alternative configuration file. The file format adhere's to the java standard of property files. If this option is not provided, the default file ''plugins.properties'' will be loaded if it exists. |
-
+| {{{-lh}}} | !-LocalhostOnly-! | ''false'' | Bind only to the local loopback interface, preventing access to thewiki from other computers in the network |
 
 !1 Java system properties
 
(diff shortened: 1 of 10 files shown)


FitNesseRoot/FitNesse/UserGuide/AdministeringFitNesse/CommandLineArguments/content.txt
```diff
@@ -15,7 +15,7 @@ A subset of configuration options can be provided through the command line. Comm
 | {{{-c <command>}}} | Command | ''none'' | Run a Rest Command and then exit. You can find more information on the -c options at RestfulServices. |
 | {{{-b <filename>}}} | !-RedirectOutput-! | ''stdout'' | redirect command output. This is most useful in conjunction with the ''-c'' option |
 | {{{-f <config file>}}} | n.a. | plugins.properties | Load an alternative configuration file. The file format adhere's to the java standard of property files. If this option is not provided, the default file ''plugins.properties'' will be loaded if it exists. |
-
+| {{{-lh}}} | !-LocalhostOnly-! | ''false'' | Bind only to the local loopback interface, preventing access to thewiki from other computers in the network |
```
fhoeben:
```suggestion
| {{{-lh}}} | !-LocalhostOnly-! | ''false'' | Bind only to the local loopback interface, preventing access to the wiki from other computers in the network |
```

fhoeben:
Do you think you will be able to add a unit test showing a connection from localhost when the flag is set?

tcnh:
Yes, I created one yesterday. I hope to find the time to polish and push it today

test/fitnesseMain/FitNesseMainTest.java
```diff
@@ -112,13 +114,25 @@ public void runningCommandWithNonExistentAddressResultsInError() throws Exceptio
...
-        throw e;
+      Integer exitCode = new FitNesseMain().launchFitNesse(arguments);
+    } catch (Exception e) {
+      assertEquals("error loading page: 404", e.getMessage());
+      throw e;
     }
   }
 
+  @Test
+  public void localhostOnlyFlagResultsInConnectableFitnesseOnLocalHost() throws Exception {
```
fhoeben: Locally, a `./gradlew test` now fails on many tests. Does it work for you?

I believe this test keeps a server running. Is that indeed the case? If so, please ensure it is cleaned up after the test completes...

tcnh: I have no failing tests... May be OS dependent.
Let me see if I can force close the Fitnesse process

fhoeben: The `Shutdown` class will do it.
If I add a call to that (in a finally block) all tests pass...

fhoeben: Since I actually have the code available, let me just push it.

tcnh: Changed the test to call the shutdown responder. Could you verify if that prevents the other tests from failing?

fhoeben: You were faster than I expected. Testing now....

PR #1189, opened 2019-02-18
Possible fix for https://github.com/unclebob/fitnesse/issues/1187

tcnh:
Clean version now.
Removed closeTrailer() call from SuiteResponder to prevent socket connection to be dropped before close() is called in some circumstances.

See https://github.com/unclebob/fitnesse/issues/1187#issuecomment-464702014 for rationale.


fhoeben:
I'm not sure whether this would address the issue reported. BUT given that `close()` calls `closeTrailer()` I see no problem merging this.

Issue #1190, opened 2019-02-19
Having a symbol that contains % breaks the evaluation of slim expressions


tcnh:
When the slim context contains a var that contains a percentage (%) charavter, the mapConverter throws a StringIndexOutOfBounds exception when filling the evaluator's context. This seems to happen because htmlparser interprets any string containing % as valid html.

Reproduction:
```
|script|echo fixture|
|$var1=|echo |10%   |
|show  |echo |$`1+1`|
```

PR #1191, opened 2019-02-19
Fix for https://github.com/unclebob/fitnesse/issues/1190

tcnh:
wrapped hash/list conversion when building evaluator context in try/catch block.

### DIFF
--- a/src/fitnesse/slim/SlimExpressionEvaluator.java
+++ b/src/fitnesse/slim/SlimExpressionEvaluator.java
@@ -37,12 +37,8 @@ public class SlimExpressionEvaluator {
     for (Map.Entry<String, MethodExecutionResult> entry : variables.entrySet()) {
       String key = entry.getKey();
       Object value = entry.getValue().getObject();
-      try {
         value = convertWikiHashes(mapCnv, value);
         value = convertWikiLists(listCnv, value);
-      } catch (Exception e) {
-        //in case conversion throws an exception, ignore it and use the current value
-      }
       engine.put(key, value);
     }
   }
--- a/src/fitnesse/slim/SlimExpressionEvaluator.java
+++ b/src/fitnesse/slim/SlimExpressionEvaluator.java
@@ -37,8 +37,12 @@ public class SlimExpressionEvaluator {
     for (Map.Entry<String, MethodExecutionResult> entry : variables.entrySet()) {
       String key = entry.getKey();
       Object value = entry.getValue().getObject();
-      value = convertWikiHashes(mapCnv, value);
-      value = convertWikiLists(listCnv, value);
+      try {
+        value = convertWikiHashes(mapCnv, value);
+        value = convertWikiLists(listCnv, value);
+      } catch (Exception e) {
+        //in case conversion throws an exception, ignore it and use the current value
+      }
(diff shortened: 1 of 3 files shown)


fhoeben:
Can you add a unit test showing the issue reported being handled?

It would, of course, be nicer if no exception would be thrown handing values with a `%` in them...

tcnh:
I agree, but the problem is in org.htmlparser.htmlparser. The parser throws the IndexOutOfBoundsException where I believe it should have just thrown a ParseException.

I don't think we can fix that right now, but I did move the exception handling to where the ParseExceptions are handled and handle the unwanted exception in the same way. I think that makes more sense.

Also added a unit test that feeds the map converter a percentage instead of decent html.

src/fitnesse/slim/SlimExpressionEvaluator.java
```diff
@@ -37,8 +37,8 @@ public void setContext(Map<String, MethodExecutionResult> variables) {
     for (Map.Entry<String, MethodExecutionResult> entry : variables.entrySet()) {
       String key = entry.getKey();
       Object value = entry.getValue().getObject();
-      value = convertWikiHashes(mapCnv, value);
-      value = convertWikiLists(listCnv, value);
+        value = convertWikiHashes(mapCnv, value);
+        value = convertWikiLists(listCnv, value);
```
fhoeben:
```suggestion
      value = convertWikiLists(listCnv, value);
```

src/fitnesse/slim/SlimExpressionEvaluator.java
```diff
@@ -37,8 +37,8 @@ public void setContext(Map<String, MethodExecutionResult> variables) {
     for (Map.Entry<String, MethodExecutionResult> entry : variables.entrySet()) {
       String key = entry.getKey();
       Object value = entry.getValue().getObject();
-      value = convertWikiHashes(mapCnv, value);
-      value = convertWikiLists(listCnv, value);
+        value = convertWikiHashes(mapCnv, value);
```
fhoeben:
```suggestion
      value = convertWikiHashes(mapCnv, value);
```

test/fitnesse/slim/converters/MapConverterTest.java
```diff
@@ -134,6 +134,13 @@ public void shouldRenderTableAsHtml() {
         "\t<tr class=\"hash_row\">", "\t\t<td class=\"hash_key\">c</td>", "\t\t<td class=\"hash_value\">d</td>", "\t</tr>", "</table>"), HtmlTag.endl), editor.toString(aMap()));
   }
 
+  @Test
+  public void nonHtmlStringWithPercentageReturnsEmptyMap() {
+    //See https://github.com/unclebob/fitnesse/issues/1190
+    MapConverter cnv = new MapConverter();
+    assertEquals(new HashMap<>(), cnv.fromString("1%"));
```
fhoeben: Why don't you use `mapMap()` as done in other tests?

tcnh: Agree that's nicer. Didn't pay enough attention to how the other tests are implemented. Current revision is more consistent

PR #1214, opened 2019-04-19
Ignore bin folder (Eclipse Buildship default output)

jan-z:
According to
https://discuss.gradle.org/t/eclipse-plugin-did-not-configure-default-output-directory-to-build/7586
it is suggested to leave the output folder as "bin", so that Gradle and
Eclipse don't interfere.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+bin
 build
 out
 lib/*


fhoeben:
I see no problem ignoring this directory.

But may I suggest you look at IntelliJ (community edition)? @amolenaar wrote a nice plugin that helps to write wiki pages (including jumping to scenario definitions) from it and run/debug tests (although you only get text and not html output).

jan-z:
Thanks for merging. I've tried IntelliJ multiple times as my main IDE and I never liked it. I only use it to develop Groovy code, because it's got the best plugin for that.

PR #1215, opened 2019-04-19
Fix locale-dependent test failure in HistoryComparerTest

jan-z:
Otherwise it fails with "java.lang.AssertionError: substring '91.67' not
found in string '     91,67'." using LANG=de_DE.UTF-8


fhoeben:
This is a change to the production code, not a test.
What is desirable in the history: I would assume using the user's locale to format a percent value is a good idea? 

Can we ensure that all tests are run using a en_US locale, also for #1217.

jan-z:
I don't really know what that code is doing. I've checked that the test is the only caller of that method. And I usually see the test as the specification. If it should return the locale specific version, the test should specify it.

Should I change the test to use a localized assert?

I don't think always running the tests using a specified locale is a good idea, as that is hard to do when you run single tests from an IDE. And for general robustness it's good to care about locales.

fhoeben:
I also do not know why this method is present, but my preference would be to change the test to expect the localised string and not the actual implementation

jan-z:
Ok, I've only changed the test to a locale specific assert now.

jan-z:
Thanks for merging

PR #1216, opened 2019-04-19
Fix MockitoException on ClosableTestSystemListener

jan-z:
When doing a "clean test" build with gradlew the error was:

org.mockito.exceptions.base.MockitoException:
Mockito cannot mock this class: interface
fitnesse.testrunner.MultipleTestsRunnerTest$ClosableTestSystemListener.

Mockito can only mock non-private & non-final classes.
If you're not sure why you're getting this error, please report to the
mailing list.

[...]

Underlying exception : org.mockito.exceptions.base.MockitoException:
Cannot create mock for interface
fitnesse.testrunner.MultipleTestsRunnerTest$ClosableTestSystemListener

The type is not public and its mock class is loaded by a different class
loader.
This can have multiple reasons:
 - You are mocking a class with additional interfaces of another class
loader
 - Mockito is loaded by a different class loader than the mocked type
(e.g. with OSGi)
 - The thread's context class loader is different than the mock's class
loader
	at fitnesse.testrunner.MultipleTestsRunnerTest.shouldCallCloseOnClosableTestSystemListener(MultipleTestsRunnerTest.java:55)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)

[...]

Caused by: org.mockito.exceptions.base.MockitoException:
Cannot create mock for interface
fitnesse.testrunner.MultipleTestsRunnerTest$ClosableTestSystemListener

The type is not public and its mock class is loaded by a different class
loader.
This can have multiple reasons:
 - You are mocking a class with additional interfaces of another class
loader
 - Mockito is loaded by a different class loader than the mocked type
(e.g. with OSGi)
 - The thread's context class loader is different than the mock's class
loader
	at net.bytebuddy.TypeCache.findOrInsert(TypeCache.java:152)
	at net.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:365)
	at net.bytebuddy.TypeCache.findOrInsert(TypeCache.java:174)
	at net.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:376)
	... 47 more

### DIFF
--- a/test/fitnesse/testrunner/MultipleTestsRunnerTest.java
+++ b/test/fitnesse/testrunner/MultipleTestsRunnerTest.java
@@ -88,7 +88,7 @@ public class MultipleTestsRunnerTest {
     }
   }
 
-  static interface ClosableTestSystemListener extends TestSystemListener, Closeable {
+  public interface ClosableTestSystemListener extends TestSystemListener, Closeable {
   }
 
 }


fhoeben:
Surprising. I run the tests all the time and do not recall getting this exception.

fhoeben:
For me a `./gradlew clean test` works just fine.
Which OS and JVM are you using? Are you running from the command line on through an IDE?

I see no reason in the test code why mockito cannot mock it.

jan-z:
```
jan@thalia:~/devel/git/fitnesse$ ./gradlew -Dorg.gradle.java.home=/usr/lib/jvm/java-8-oracle/ -version

------------------------------------------------------------
Gradle 5.3.1
------------------------------------------------------------

Build time:   2019-03-28 09:09:23 UTC
Revision:     f2fae6ba563cfb772c8bc35d31e43c59a5b620c3

Kotlin:       1.3.21
Groovy:       2.5.4
Ant:          Apache Ant(TM) version 1.9.13 compiled on July 10 2018
JVM:          1.8.0_191 (Oracle Corporation 25.191-b12)
OS:           Linux 5.0.0-13-generic amd64

jan@thalia:~/devel/git/fitnesse$ cat /etc/issue
Ubuntu 19.04 \n \l
```

It only fails with "clean test":

```
jan@thalia:~/devel/git/fitnesse$ ./gradlew -Dorg.gradle.java.home=/usr/lib/jvm/java-8-oracle/ clean test
[...]
> Task :test

fitnesse.testrunner.MultipleTestsRunnerTest > shouldCallCloseOnClosableTestSystemListener FAILED
    org.mockito.exceptions.base.MockitoException at MultipleTestsRunnerTest.java:55
        Caused by: org.mockito.exceptions.base.MockitoException at MultipleTestsRunnerTest.java:55
```

Without clean it works:

```
jan@thalia:~/devel/git/fitnesse$ ./gradlew -Dorg.gradle.java.home=/usr/lib/jvm/java-8-oracle/ test

BUILD SUCCESSFUL in 54s
7 actionable tasks: 1 executed, 6 up-to-date
```

jan-z:
It seems to be because Gradle is using different classloaders for the mocked type. I don't use Gradle myself (just Maven), so I can't really help there.

fhoeben:
It's really strange: the problem doesn't appear for me (or any other contributor as far as I'm aware) or on travis-ci.

test/fitnesse/testrunner/MultipleTestsRunnerTest.java
```diff
@@ -88,7 +88,7 @@ public boolean matches(Descriptor descriptor) {
     }
   }
 
-  static interface ClosableTestSystemListener extends TestSystemListener, Closeable {
+  public static interface ClosableTestSystemListener extends TestSystemListener, Closeable {
```
fhoeben: The `static` keyword is redundant here. Does removing that impact the error you get in any way?

I don't mind changing this to `public`, but at least let's remove the extra static as well

jan-z:
The `static` does not make a difference, it fails either way. I've removed it and replaced it with `public`.

test/fitnesse/testrunner/MultipleTestsRunnerTest.java
jan-z: The `static` does not make a difference, it fails either way. I've removed it and replaced it with `public`.

fhoeben:
I still want to understand why you run into this problem. I see 2 possible reasons:

1. You specify `org.gradle.java.home` on the command line. According to the documentation:
> Specifies the Java home for the Gradle build process. The value can be set to either a jdk or jre location, however, depending on what your build does, using a JDK is safer. A reasonable default is derived from your environment (JAVA_HOME or the path to java) if the setting is unspecified. This does not affect the version of Java used to launch the Gradle client VM
Could it be that the problem originates from the two different VMs (one for the gradle client and one for the build process)? Do you get a same problem if you run without that `-D` option (and change your path so that java is the default used on the command line)?

2. Do you have a `bin` directory below the directory where you execute the gradlew command? If so: Does the problem occur if you remove that before running `clean test`?

jan-z:
1. It also happens when I change my default VM:

```
jan@thalia:~/devel/git/fitnesse$ java -version
java version "1.8.0_191"
Java(TM) SE Runtime Environment (build 1.8.0_191-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.191-b12, mixed mode)

jan@thalia:~/devel/git/fitnesse$ ./gradlew clean && ./gradlew test

[...]

> Task :test

fitnesse.testrunner.MultipleTestsRunnerTest > shouldCallCloseOnClosableTestSystemListener FAILED
    org.mockito.exceptions.base.MockitoException at MultipleTestsRunnerTest.java:55
        Caused by: org.mockito.exceptions.base.MockitoException at MultipleTestsRunnerTest.java:55
```
Btw: It works on the second test run.


2. No difference:

```

jan@thalia:~/devel/git/fitnesse$ rm -r bin

jan@thalia:~/devel/git/fitnesse$ LANG=C ls -l bin
ls: cannot access 'bin': No such file or directory

jan@thalia:~/devel/git/fitnesse$ ./gradlew clean && ./gradlew test

[...]

> Task :test

fitnesse.testrunner.MultipleTestsRunnerTest > shouldCallCloseOnClosableTestSystemListener FAILED
    org.mockito.exceptions.base.MockitoException at MultipleTestsRunnerTest.java:55
        Caused by: org.mockito.exceptions.base.MockitoException at MultipleTestsRunnerTest.java:55
```

jan-z:
Thanks for merging this workaround. I guess the root cause is a Gradle or Mockito issue.

PR #1217, opened 2019-04-19
Fix locale-dependent failure in ErrorSituationsDuringSlimServerStartup

jan-z:
The test fails because the exception message is localized. For example
when running with LANG=de_DE.UTF-8 the message is:

java.net.BindException: Die Adresse wird bereits verwendet (Bind failed)
        at java.net.PlainSocketImpl.socketBind(Native Method)

(which is German for "Address already in use")

It seems that there is no way to get the non-localized message:
https://stackoverflow.com/questions/17159357/is-there-a-way-to-force-exception-message-to-be-english-for-java-1-7

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/TestSystemSlimSuite/ErrorSituationsDuringSlimServerStartup/A_TimeoutAsSlimHeaderNotReceived/content.txt
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/TestSystemSlimSuite/ErrorSituationsDuringSlimServerStartup/A_TimeoutAsSlimHeaderNotReceived/content.txt
@@ -15,9 +15,9 @@ After slim.timeout seconds we should abort and stop the test.
 !define SLIM_PORT (-!${FITNESSE_PORT}!-)
 !define SLIM_FLAGS (-v)
 -! |
-|check         |request page    |TestPage?test                   |200          |
-|ensure        |content contains|Timeout while reading slim header from client.|
-|show collapsed|content                                                        |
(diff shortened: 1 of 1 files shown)


jan-z:
Should we change something about this?

FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteSlimTests/TestSystemSlimSuite/ErrorSituationsDuringSlimServerStartup/A_TimeoutAsSlimHeaderNotReceived/content.txt
```diff
@@ -19,5 +19,5 @@ After slim.timeout seconds we should abort and stop the test.
 |ensure        |content contains|Timeout while reading slim header from client.|
 |show collapsed|content                                                        |
 |check         |request page    |TestPage?executionLog           |200          |
-|ensure        |content contains|Address already in use                        |
+|ensure        |content contains|java.net.BindException                        |
 |show collapsed|content                                                        |
```
fhoeben: Can you also check for `(Bind failed)`, which I gather from your report is also present in German

jan-z: I changed it to match for both versions, so that the original exception message is still visible in the test. That way it's easier to understand what it's doing.

fhoeben:
The regex makes the test a bit harder to read, but correct and local independent. Thanks

Issue #925, opened 2016-05-24
WhereUsed for pages under a Symbolic Link


cedclement:
Hello, the last release (20160515) seems to have greatly improved the symbolic links management, however I think there is still an issue with the WhereUsed responder : pages located under a Symbolic Link are not beeing found.

For instance, FitNesse.UserGuide.FitNesseWiki.SymbolicLinks?whereUsed shows 2 results, whereas the same page under a Symbolic Link FitNesse.FullReferenceGuide.UserGuide.FitNesseWiki.SymbolicLinks?whereUsed the references are not found.

amolenaar:
I think this commit solves it. At least to the extend that you'll find the pages the symlink is related to.

PR #1218, opened 2019-04-20
Fix suites consisting of symbolic pages

jan-z:
If using a Suite-page that uses symlinks to link to different external
roots, starting with commit 16bb43e only the contents of the first
symlinked page was displayed in the output of "!contents".

This regression was introduced by fixing #925.


jan-z:
Thanks for merging. This bug has been keeping us from upgrading for some time.

PR #1220, opened 2019-04-25
Optimize expression evaluator

tcnh:
Add only symbols referenced in the expression to the evaluator's context. This results in up to 40% less consumed time evaluating expressions on pages that have a large amount of slim symbols in their scope.

Also, catch any exception from ParseUtil, as it can trip over localized characters encoded with %'s causing it to throw runtime exceptions that don't need specific handling (if we can't parse the string as html, we treat it as a string)

### DIFF
--- a/src/fitnesse/slim/SlimExpressionEvaluator.java
+++ b/src/fitnesse/slim/SlimExpressionEvaluator.java
@@ -36,7 +36,7 @@ public class SlimExpressionEvaluator {
 
     for (Map.Entry<String, MethodExecutionResult> entry : variables.entrySet()) {
       String key = entry.getKey();
-      if(expr.contains(key)) {
+      if (expr.contains(key)) {
         Object value = entry.getValue().getObject();
         value = convertWikiHashes(mapCnv, value);
         value = convertWikiLists(listCnv, value);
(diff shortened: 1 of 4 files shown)


src/fitnesse/slim/SlimExpressionEvaluator.java
```diff
@@ -30,16 +30,18 @@ public SlimExpressionEvaluator(ScriptEngine engine) {
...
     Converter<Map> mapCnv = ConverterRegistry.getConverterForClass(Map.class);
     Converter<List> listCnv = ConverterRegistry.getConverterForClass(List.class);
 
     for (Map.Entry<String, MethodExecutionResult> entry : variables.entrySet()) {
       String key = entry.getKey();
-      Object value = entry.getValue().getObject();
-      value = convertWikiHashes(mapCnv, value);
-      value = convertWikiLists(listCnv, value);
-      engine.put(key, value);
+      if(expr.contains(key)) {
```
fhoeben:
```suggestion
      if (expr.contains(key)) {
```

tcnh:
Note: There's a hidden/undocumented change in this PR, causing the Initial suiteName to be extracted from the Description object instead of the wiki page path. This ensures proper naming of tests that were initiated from the "new" ReRunLastFailures, allowing history, trend graphs and flakiness reporting in Allure.

Issue #1221, opened 2019-04-25
Fitnesse slim fails with puzzling IndexOutOfBoundsException


abailly:
We are using Fitnesse/Slim with (slightly customised) pySlim server. When running `Suite` for a set of pages I got the following error: 
<img width="538" alt="Screenshot 2019-04-25 at 17 36 36" src="https://user-images.githubusercontent.com/40221/56748916-fd9e5e00-6780-11e9-89e2-03141c31c8e5.png">

I tried to debug the server but to no avail, probably because my debug-fu on Java is too rusty to be helpful. I also tried to upgrade to latest fitnesse (20190421), and to pass `-v` for more verbose debugging but to no avail. Is there way to know where/why this exception is thrown?

tcnh:
Can you share the wiki content to reproduce this? I'm pretty sure I've seen this before with a table that was not properly formatted

abailly:
Unfortunately not, but I came to a similar conclusion. However it's pretty much unclear _which_  page is wrongly formatted...

abailly:
I found the culprit: It's a script table line containing a single item. `ScriptTable` fails to extract the action's name and throws this error. Would you consider a PR to fix that? What would be a more helpful error to throw?

tcnh:
A PR with an improved exception message is always welcome!

PR #1222, opened 2019-04-26
script table throws SyntaxError on malformed assignment

abailly:
The `SyntaxError` exception is propagated up the call chain which requires more changes than expected. This should translate to `Bad Table` message when rendered in FitNesse. 
See #1221 for context of this small PR.

### DIFF
--- a/src/fitnesse/testsystems/slim/tables/ScriptTable.java
+++ b/src/fitnesse/testsystems/slim/tables/ScriptTable.java
@@ -21,6 +21,7 @@ public class ScriptTable extends SlimTable {
   public ScriptTable(Table table, String tableId, SlimTestContext context) {
     super(table, tableId, context);
   }
+
   /**
    * Template method to provide the keyword that identifies the table type.
    *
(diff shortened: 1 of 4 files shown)


src/fitnesse/testsystems/slim/tables/ScriptTable.java
```diff
@@ -297,15 +298,19 @@ protected boolean invokesSequentialArgumentProcessing(String cellContents) {
...
     }
 
-    public static String getActionNameStartingAt(Table table, int startingCol, int endingCol, int row) {
+    public static String getActionNameStartingAt(Table table, int startingCol, int endingCol, int row) throws SyntaxError {
       StringBuilder actionName = new StringBuilder();
-      actionName.append(table.getCellContents(startingCol, row));
+      try {
+        actionName.append(table.getCellContents(startingCol, row));
+      } catch (IndexOutOfBoundsException e) {
+        throw new SyntaxError("badly formatted table: Expected an action name to be found at (" + row + "," + startingCol + ")");
```
fhoeben: I believe the message could be a bit clearer. The [script table documentation](http://fitnesse.org/FitNesse.UserGuide.WritingAcceptanceTests.SliM.ScriptTable) does not mention 'action name' it uses the term 'function call'. 
Furthermore the 'catch block' of the `SyntaxException` already adds 'Bad table!' so that does not have to be repeated in the message.

What would you think of the following message: 
`"Too few columns in row " + (row + 1) + ". Expected a  function in column " + (startingCol + 1) + "."`

abailly: I think it's better :)
Updating the PR now

abailly: BTW, do you know where those error messages rendering happen in the code? I think it would make sense to add a test there, perhaps?

fhoeben: `fitnesse.testsystems.slim.HtmlSlimTestSystem` line 57.

I don't believe adding a test there is needed, as it is very generic handling of SyntaxErrors coming from processing any Slim table. So having a unit test at ScriptTable level seems sufficient to me.

PR #1225, opened 2019-05-18
Refactor Call(AndAssign)Instruction in order to remove duplicate code

kdridi:
The main idea of this PR is just about merging the functionalities of CallAndAssignInstruction and CallInstruction inside an abstract class named CallAndOptionalAssignInstruction by using the Optional JDK class.

This PR remove a lots of duplicate lines in the two original classes.

### DIFF
--- a/src/fitnesse/slim/instructions/CallAndOptionalAssignInstruction.java
+++ b/src/fitnesse/slim/instructions/CallAndOptionalAssignInstruction.java
@@ -1,0 +1,74 @@
+package fitnesse.slim.instructions;
+
+import fitnesse.slim.NameTranslator;
+import fitnesse.slim.SlimException;
+
+import java.util.Arrays;
+import java.util.Optional;
+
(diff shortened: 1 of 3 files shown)


src/fitnesse/slim/instructions/CallAndOptionalAssignInstruction.java
```diff
@@ -0,0 +1,74 @@
...
+    this.symbolName = symbolName;
+    this.instanceName = instanceName;
+    this.methodName = methodName;
+    this.args = args;
+  }
+
+  @Override
+  protected InstructionResult executeInternal(InstructionExecutor executor) throws SlimException {
+    if (symbolName.isPresent()) {
+      Object result = executor.callAndAssign(symbolName.get(), instanceName, methodName, args);
```
fhoeben: why not have a single `return` and use an if/else to assign `result`?

PR #1227, opened 2019-06-05
Allow plugins to override default wiki theme

fhoeben:
By default FitNesse uses the 'bootstrap' theme, but this can be overridden using a `Theme` property in either: `plugins.properties`,  an environment variable  or system property.

In this PR a .jar plugin placed in the `plugins` directory can define an alternative theme, instead of bootstrap, to be used when no explicit theme is configured. This can be achieved by implementing the new method `getDefaultTheme()` in its `PluginFeatureFactory`.


tcnh:
lgtm!

Issue #1229, opened 2019-06-15
Order of suites executed should not depend on presence of SuiteSetup


fhoeben:
The order that tests is executed shouldn't matter, but for reporting purposes it can be convenient if ordering remains the same between runs.
At the moment the ordering of tests by FitNesse is stable, also when working with nested suites (e.g. Suite1 contains SubSuite1 and SubSuite2). BUT the ordering of execution of nested suites changes based on whether they have their own SuiteSetUp or SuiteTearDown pages.

So initially (without SuiteSetUp inside SubSuite1) SubSuite1 is executed before SubSuite2, but when a SuiteSetUp is added inside SubSuite1 it could happen that SubSuite2 is suddenly executed before SubSuite1. This is not always the case, but can happen. 
For test outcome it shouldn't matter, but it can be surprising and make test report more complex to read/explain. (See for instance https://stackoverflow.com/questions/56597945/alternative-for-nested-suitesetups-that-maintains-ordering)

In the past there has been an issue about test ordering (#131), which was marked as "won't fix". The discussion in that issue seemed to go in the direction that test ordering should not matter. While that is true in general. Test ordering by FitNesse is normally quite predictable and it seems odd that adding a SuiteSetUp page changes ordering, sometimes.

ALuckyGuy:
I agree with you that the order should be consistent, and it *can* happen that the order matters. 

For me it's common to have sub-suites where when executed as part of a larger suite, the order *is* important. I name my suites and test explicitly so I know the order they execute. For instance I might have one suite of tests that hires employees, and another suites of tests that verify I can perform certain job actions on newly hired employees. Of course I need the first sub-suite to execute before I can run the next suite... the order is very important in this case and I don't think it's an uncommon or unreasonable expectation.

fhoeben:
Another thing I noticed when looking at the code determining suite ordering and -setup and -teardown is that the same SuiteSetUp or SuiteTearDown may be executed more than once. 

FitNesse creates groups of tests that have the same SuiteSetUp and SuiteTearDown and runs these groups one after the other. For each group it runs the appropriate setup and teardown. But it does not consider whether another group uses the same setup or teardown. So if two groups share the same setup (but a different teardown) the setup is executed twice.

This is also not what I would expect. I would expect the SuiteSetUp to be executed once, before the first test in the suite it belongs to, and the SuiteTearDown to be executed once, after the last test in its suite.

JessefSpecialisterren:
[...]

PR #1230, opened 2019-06-15
Make test ordering more logical

fhoeben:
This PR addresses #1229 and ensure each SuiteSetUp and -TearDown is only executed once per run.

Test ordering no longer depends on the presence (or absence) of SuiteSetUp pages. The test ordering is fixed (alphabetically on full page path) and each SuiteSetUp and SuiteTearDown is added into the mix at the appropriate point. (setups just before the first test in their suite and teardown just after the last in their suite)

### DIFF
--- a/src/fitnesse/testrunner/PageListSetUpTearDownSurrounder.java
+++ b/src/fitnesse/testrunner/PageListSetUpTearDownSurrounder.java
@@ -1,89 +1,63 @@
 package fitnesse.testrunner;
 
-import java.util.Collections;
+import fitnesse.wiki.PageData;
+import fitnesse.wiki.WikiPage;
+
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-
-import fitnesse.wiki.PageData;
-import fitnesse.wiki.PathParser;
-import fitnesse.wiki.WikiPage;
-
-import static java.util.Arrays.asList;
+import java.util.Optional;
+import java.util.Set;
 
 public class PageListSetUpTearDownSurrounder {
-  private final WikiPage root;
-
-  public PageListSetUpTearDownSurrounder(WikiPage root) {
-    this.root = root;
-  }
 
   public List<WikiPage> surroundGroupsOfTestPagesWithRespectiveSetUpAndTearDowns(List<WikiPage> pageList) {
-    Map<String, List<WikiPage>> pageSetUpTearDownGroups = createPageSetUpTearDownGroups(pageList);
-    return reinsertPagesViaSetUpTearDownGroups(pageSetUpTearDownGroups);
-  }
+    List<WikiPage> pagesToRun = new LinkedList<>();
 
-  private Map<String, List<WikiPage>> createPageSetUpTearDownGroups(List<WikiPage> pageList) {
-    Map<String, List<WikiPage>> pageSetUpTearDownGroups = new HashMap<>();
+    Map<WikiPage, WikiPage> tearDownAfter = findPagesBeforeTearDown(pageList);
+    Set<WikiPage> setupsAdded = new HashSet<>();
     for (WikiPage page : pageList) {
-      makeSetUpTearDownPageGroupForPage(page, pageSetUpTearDownGroups);
+      addSetUpIfNeeded(pagesToRun, setupsAdded, page);
+      pagesToRun.add(page);
+      addTearDownIfPossible(pagesToRun, tearDownAfter, page);
     }
-    return pageSetUpTearDownGroups;
+
+    return pagesToRun;
   }
 
-  private void makeSetUpTearDownPageGroupForPage(WikiPage page, Map<String, List<WikiPage>> pageSetUpTearDownGroups) {
-    String group = getSetUpTearDownGroup(page);
-    List<WikiPage> pageGroup;
-    if (pageSetUpTearDownGroups.get(group) != null) {
-      pageGroup = pageSetUpTearDownGroups.get(group);
-      pageGroup.add(page);
-    } else {
-      pageGroup = new LinkedList<>();
-      pageGroup.add(page);
-      pageSetUpTearDownGroups.put(group, pageGroup);
+  private Map<WikiPage, WikiPage> findPagesBeforeTearDown(List<WikiPage> pages) {
+    Map<WikiPage, WikiPage> pageBeforeSuiteTearDown = new HashMap<>();
+    for (WikiPage page : pages) {
+      getClosestInheritedPage(page, PageData.SUITE_TEARDOWN_NAME).ifPresent(suiteTearDown -> {
+          pageBeforeSuiteTearDown.put(suiteTearDown, page);
+        });
+    }
+    Map<WikiPage, WikiPage> tearDownAfter = new IdentityHashMap<>();
+    for (Map.Entry<WikiPage, WikiPage> entry : pageBeforeSuiteTearDown.entrySet()) {
+      tearDownAfter.put(entry.getValue(), entry.getKey());
+    }
+    return tearDownAfter;
+  }
+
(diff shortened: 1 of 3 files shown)


JessefSpecialisterren:
Just tested this PR both with my simplified reproduction suite and the full testsuite for our client that brought the problem to our attention in the first place. Both of them now run in the expected order 🎉!

I can't vouch for more complicated trees with byzantine combinations of setups and teardowns, perhaps it is worth the effort writing an acceptance test for that?

fhoeben:
I feel confident the current unit tests cover the relevant cases. What byzantine cases are you thinking of?

What worries me more is whether people expect/rely on suitesetups or -teardowns to be executed more than once. To me it is counter intuitive and it should not be done according to http://fitnesse.org/FitNesse.UserGuide.WritingAcceptanceTests.SpecialPages, but it WAS what FitNesse did.

woodybrood:
I'm not sure about users depending on multiple executions of the SuiteSetUp and tear down, but it has always been recommended that users do not rely on the order of test execution. It implies a coupling of tests to each other that isn't very clean. Depending on multiple executions of the SuiteSetUp and SuiteTearDown is probably just as problematic.

I don' t do FitNesse any more, but when I did, I know we just wanted the SuiteSetUp and SuiteTearDown to do their job for the tests between them. If they ran more than once it neither helped, or hurt us.

Dan

JessefSpecialisterren:
I've been thinking about this PR a bit more, and one case came to mind that may cause trouble with the new behavior. Suppose you have the following test hierarchy:

* TopSuite
  * SubSuite1
    * [tests]
  * SubSuite2
    * [tests]
    * SuiteSetUp => !includes TopSuite.SuiteSetUp, performs some action after the browser has started (e.g. logs in on a certain website)
  * SuiteSetUp => starts a browser
  * SuiteTearDown => stops the browser

Under FitNesse's current handling of nested SuiteSetUps, SubSuite2's SuiteSetUp would always be run without a browser already open. Even if SubSuite1 runs before it, TopSuite.SuiteTearDown will be run before SubSuite2.SuiteSetUp.

Under the new behavior, SubSuite2's SuiteSetUp would run _with_ a browser already open if TopSuite were run. This would result in an attempt to start a browser while a browser is already open. AFAIK, FitNesse currently has no way to prevent this, so one would have to adapt the fixture to deal with this. Not ideal.

I wonder if there is an elegant way to solve this problem. Getting rid of the include in SubSuite2's SuiteSetUp would result in no browser being started when SubSuite2 or one of its tests is run directly, so that wouldn't work. I know PHP has an `include_once`, something like that might do the trick. Or perhaps some kind of keyword or option that runs SuiteSetUps higher in the tree if they haven't run yet. Thoughts?

fhoeben:
A browser being closed (or not) is at the start of a suite setup indeed the kind of problem I'm thinking of. 
The easiest way to prevent this problem in the new situation seems to be not to nest the two suites inside one another. But that is indeed a breaking change if someone has setup their tests this way in at the moment.

fhoeben:
@woodybrood do you recall, can you explain, that the reasoning is/was behind the choice to run only the _closest_ SuiteSetUp before each test? Quoting from http://fitnesse.org/FitNesse.UserGuide.WritingAcceptanceTests.SpecialPages:

> The SuiteSetUp and SuiteTearDown that are included in a suite run are the pages that are the nearest brothers or uncles of the Suite page.

I would have expected all the ones at the current level or higher to be included (like is done for ScenarioLibraries):
> Unlike the other special pages, all brother and uncle ScenarioLibrary pages are included. The oldest (grandest uncle) is included first. The brother, if it exists, is included last.

This has always puzzled me, but I've taken it as a given. Causing me to include higher level SuiteSetups in the closest one. Which causes the issue with the current PRs content, which makes the higher-level SuiteSetup to be executed before a nested one but does not execute the corresponding SuiteTearDown (until the entire suite, including all nested suites to-be-run,  has completed).

So my question boils down to, given the structure below: why does FitNesse currently (before this PR's changes) not run TopSuite.SuiteSetUp (before TopSuite.SubSuite2.SuiteSetUp) when executing only TopSuite.SubSuite2?

- TopSuite
  - SubSuite1
    - [tests]
  - SubSuite2
    - [tests]
    - SuiteSetUp
  - [tests]
  - SuiteSetUp
  - SuiteTearDown

woodybrood:
@fhoeben I have a one answer for you that you may or may not find satisfactory. I can explain the ScenarioLibrary difference, as that is a feature that I championed to get in. When I was at RedPrairie (now JDA) we were using some rather insane libraries of scenarios that allowed us to run commands against our system under test. We could then !include library pages to tests based on what the test needed to do. 

As we grew this, we also realized that we had common themes and needs in our test suites, as well as common needs to all tests. This started a conversation about having scenario pages that were automatically included in a suite, but we also had scenarios that we wanted in every test. The discussion quickly grew into the idea of multiple levels of scenario libraries and the parent (technically uncle) scenario libraries being included as well as the siblings. We knew this was not the same as SuiteSetUp, but we felt it was necessary and convinced Uncle Bob that it made sense.

I'd tell you that I'm pretty sure I had the same questions about SuiteSetUp, and I'm guessing during this time we discussed it again. And while I can recall the decision to be inconsistent, I cannot recall the reason that only the closest SuiteSetUp is called. 

If I recall correctly, the behavior itself is quite old. I don't think we wanted to change the behavior, as it would really throw a wrench in some users test suites.

I wish I was more help, but that's all I've got for you right now.

fhoeben:
Thanks for explaining @woodybrood. 
I must admit I'm not quite satisfied with the reasoning, but I understand the thinking about what changing SuiteSetUp behaviour would mean to existing users, that's what I facing now as well.

To me a hierarchical test setup would always mean that a nested suite is ALSO part of the parent's suite and therefore all 'uncle' SuiteSetUps are applicable and should be run.

In my experience having multiple different SuiteSetUps in one hierarchy is actually quite rare, and when it occurs I see people importing the higher level SuiteSetUp in the lower ones (to work around the fact they are not executed automatically). For people using that approach changing the SuiteSetUp behaviour would be a breaking change, but one that is easy to incorporate and actually makes the tests more straightforward.

Can anyone provide a good example why NOT executing higher level (uncle) SuiteSetUp is a better idea?

fhoeben:
I will update this PR to be based on #1231 allowing a choice to use either the current/previous test ordering or opting for a new approach.

In the new approach I intend to have all uncle and brother suite set ups be executed (once) before each test, and not just the nearest one. Similar to how scenario libraries are currently added

JessefSpecialisterren:
> Can anyone provide a good example why NOT executing higher level (uncle) SuiteSetUp is a better idea?

I've been thinking about this, but any example I can come up with can be trivially solved by reorganizing your suites to something that makes more sense anyway. If anything, executing higher level SuiteSetUps encourages better suite organization, as far as I can tell. Should I be able to think of a counterexample, I'll post it, though.

> allowing a choice to use either the current/previous test ordering or opting for a new approach

That seems like a great idea. Just like using SLIM is opt-in based right now, so could executing higher level SuiteSetUps be.

fhoeben:
I made an (opt-in) switch enabling an alternate ordering of tests and suite set ups and tear downs. By adding `!define ALL_UNCLE_SUITE_SETUPS {true}` to the top level of your suite you get:
 * all (higher level) SuiteSetUps are added when executing your suite (or page)
 * the tests in a suite/run are executed in the (alphabetical) order they are also listed in the contents of the suite (instead of grouped based on the nearest suite setup and teardown applicable). This also applies to subsuites irrespective of whether they are a SuiteSetUp or SuiteTearDown
 * each SuiteSetUp and SuiteTearDown is executed exactly once (and not once per group they apply to)

To me the new approach makes much more sense when thinking of nested suites, than the current 'nearest only' approach. But I don't feel confident to make it the default, given that it will likely break existing suites where multiple suite setups and teardowns are used.

I would like to hear other people's thoughts and experiences.

fhoeben:
Replaced by #1233

PR #1233, opened 2019-06-26
New approach to address test ordering in suite hierarchies with multiple SuiteSetUps

fhoeben:
I made an (opt-in) switch enabling an alternate ordering of tests and suite set ups and tear downs, to address #1229. By adding !define ALL_UNCLE_SUITE_SETUPS {true} to the top level of your suite you get:

all (higher level) SuiteSetUps are added when executing your suite (or page)
the tests in a suite/run are executed in the (alphabetical) order they are also listed in the contents of the suite (instead of grouped based on the nearest suite setup and teardown applicable). This also applies to subsuites irrespective of whether they are a SuiteSetUp or SuiteTearDown
each SuiteSetUp and SuiteTearDown is executed exactly once (and not once per group they apply to)
To me the new approach makes much more sense when thinking of nested suites, than the current 'nearest only' approach. But I don't feel confident to make it the default, given that it will likely break existing suites where multiple suite setups and teardowns are used.

I would like to hear other people's thoughts and experiences.


jinlxz:
I like this new feature, currently I achieved it by including SuiteSetUp of parent suite into current level SuiteSetUp page, it is not convenient.

PR #1241, opened 2019-09-27
Allow slim symbols in scenario arguments to be used in action name

tcnh:
Scenario arguments can be used to construct action names in a scenario table.
However, if the argument is filled using a slim symbol, this symbol was not expanded before the action name was determined. This PR fixes that omission, so that slim symbols passed to a scenario behave in the same manner as a normal String passed to a scenario.

### DIFF
--- a/src/fitnesse/testsystems/slim/tables/ScenarioTable.java
+++ b/src/fitnesse/testsystems/slim/tables/ScenarioTable.java
@@ -165,9 +165,9 @@ public class ScenarioTable extends SlimTable {
         for (Map.Entry<String, String> scenarioArgument : scenarioArguments.entrySet()) {
           String arg = scenarioArgument.getKey();
           if (getInputs().contains(arg)) {
-            String argument = scenarioArguments.get(arg);
-            content = StringUtils.replace(content, "@" + arg, replaceSymbols(argument));
-            content = StringUtils.replace(content, "@{" + arg + "}", replaceSymbols(argument));
+            String argument = replaceSymbols(scenarioArguments.get(arg));
+            content = StringUtils.replace(content, "@" + arg, argument);
+            content = StringUtils.replace(content, "@{" + arg + "}", argument);
           } else {
             throw new SyntaxError(String.format("The argument %s is not an input to the scenario.", arg));
           }
--- a/src/fitnesse/testsystems/slim/tables/ScenarioTable.java
+++ b/src/fitnesse/testsystems/slim/tables/ScenarioTable.java
@@ -63,7 +63,7 @@ public class ScenarioTable extends SlimTable {
     return isNameParameterized(firstNameCell);
   }
 
-    protected void getScenarioArguments() {
+  protected void getScenarioArguments() {
     if (parameterized) {
       getArgumentsForParameterizedName();
     } else {
@@ -95,7 +95,7 @@ public class ScenarioTable extends SlimTable {
     String[] arguments = argumentString.split(",");
 
     for (String argument : arguments) {
-        splitInputAndOutputArguments(argument);
+      splitInputAndOutputArguments(argument);
     }
   }
 
@@ -166,8 +166,8 @@ public class ScenarioTable extends SlimTable {
           String arg = scenarioArgument.getKey();
           if (getInputs().contains(arg)) {
             String argument = scenarioArguments.get(arg);
-            content = StringUtils.replace(content, "@" + arg, argument);
-            content = StringUtils.replace(content, "@{" + arg + "}", argument);
+            content = StringUtils.replace(content, "@" + arg, replaceSymbols(argument));
+            content = StringUtils.replace(content, "@{" + arg + "}", replaceSymbols(argument));
           } else {
             throw new SyntaxError(String.format("The argument %s is not an input to the scenario.", arg));
           }
@@ -199,7 +199,7 @@ public class ScenarioTable extends SlimTable {
   }
 
   protected ScriptTable createChild(Class<? extends ScriptTable> parentTableClass, Table newTable, SlimTestContext testContext) throws TableCreationException {
-      return SlimTableFactory.createTable(parentTableClass, newTable, id, testContext);
+    return SlimTableFactory.createTable(parentTableClass, newTable, id, testContext);
   }
 
   public List<SlimAssertion> call(String[] args, ScriptTable parentTable, int row) throws TestExecutionException {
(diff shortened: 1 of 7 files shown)


src/fitnesse/testsystems/slim/tables/ScenarioTable.java
```diff
@@ -166,8 +166,8 @@ public String substitute(String content) throws SyntaxError {
           String arg = scenarioArgument.getKey();
           if (getInputs().contains(arg)) {
             String argument = scenarioArguments.get(arg);
-            content = StringUtils.replace(content, "@" + arg, argument);
-            content = StringUtils.replace(content, "@{" + arg + "}", argument);
+            content = StringUtils.replace(content, "@" + arg, replaceSymbols(argument));
```
fhoeben: I believe it would be nicer to have only a single call to `replaceSymbols`, before the first content alteration.

fhoeben:
Is expanding symbols always a good idea? 
If the symbol is not used for an action name, but will be passed to an action as a parameter (i.e. will be an argument to a method invoked) we don't want the (string) value, as known by the wiki, to be passed to the method but we want the symbol reference to be passed (so the actual symbol content/instance can be used instead of just its String representation). 

Passing the string value/representation is not a problem for 'simple' symbol values (e.g. strings or numbers), but will be relevant for symbols that contain/reference complex objects (e.g. a map, list or a point).

tcnh:
As long as the wiki has a converter for these object types, that should not be a problem. But I see the potential issue if there is not. I'll give it some thought and see if we can come up with a solution that works in both situations..

fhoeben:
I believe supporting symbols to be used in action names should probably be done inside ScriptTable, as it seems strange that that would only work when invoking the action via a scenario and not when making a plain script. The scenario arguments (where you made the change now) are of course only applicable when using a scenario, but slim symbols are always available after all.

tcnh:
Implementing this in the scenario table is intentional. I think supporting symbols in action names in a plain script table is not a great idea, as it opens pandora's box of tests containing business logic. The reason I dit this in the scenario table is that it would force this to explicitly be an intentional action. (and the symbol (data) is at least passed as an argument once)

six42:
Hi Tom,
could you add a wiki page which demonstrates the usage of this PR.
I am not sure that I understand what you want to achieve and can't judge if it breaks existing test cases.

tcnh:
Hi @six42, I added a wiki page that demonstrates an abstract use case for this PR.

six42:
Hi Tom,

thanks for adding the test. Makes it more clear what you want to achieve.
I would not say this increases readability of test cases. 
Nevertheless there might be edge cases where this is the right pattern.

 
As suspected already by Fred your proposed PR can break existing code where the parameters are supposed to  be custom classes.
For example this test:

|import|
|fitnesse.slim.test|

!|script| test slim|
|set String | Bob|
|$o=| get Fixture|

!|scenario|check other|value|
|check |getStringFromOther | @value| Bob|


!|script| test slim|
|check |getStringFromOther | $o| Bob|
|check other| $o|

!|check other|
|value|
|$o|


I assume you have a good reason to rule out the usage of variables instead of symbols. This would work without the PR and is more readable.
|Library    |
|eg.Division|

|scenario     |Mydivision _|firstMethod,numerator,secondMethod,denominator,quotient?|
|@firstMethod |@numerator                                                           |
|@secondMethod|@denominator                                                         |
|$quotient=   |quotient                                                             |

!define numeratorSetter {setNumerator}
!define denominatorSetter {setDenominator}

|Mydivision                                                         |
|firstMethod     |numerator|secondMethod      |denominator|quotient?|
|${numeratorSetter}|35       |${denominatorSetter}|5          |~=7      |

six42:
Have you considered adding this feature in the script table

https://github.com/unclebob/fitnesse/blob/master/src/fitnesse/testsystems/slim/tables/ScriptTable.java


Line 320:
- return actionName.toString().trim();
+ return replaceSymbols(actionName.toString()).trim();

Llimits the impact to the action name which currently fails if a symbol is used in it. But leaves other parameters untouched.

tcnh:
That is in fact a very nice suggestion.. :)
The latest commit implements it like that and also expands the symbols when evaluating results in all cases (when a symbol was used as input in the scanearrio, the result page would only show the symbol reference and not the value)

I think this approach didn't break any existing functionality. (At least not that I could find)

six42:
Hi Tom, 
sorry to ask.
I cloned your build and have some junit tests fail and FitNesse.SuiteAcceptanceTests.SuiteSlimTests.TestDetailedDiff is not readable anymore.
Is it working 100% for you.

fitnesse.testsystems.slim.tables.ReturnedValueExpectationTest > matchedSymbolIsReplaced FAILED
    org.junit.ComparisonFailure at ReturnedValueExpectationTest.java:32

fitnesse.testsystems.slim.tables.ReturnedValueExpectationTest > mismatchedSymbolIsReplaced FAILED
    org.junit.ComparisonFailure at ReturnedValueExpectationTest.java:32

fitnesse.testsystems.slim.tables.ScenarioAndScriptTableTest > inputAndOutputWithSymbol FAILED
    org.junit.ComparisonFailure at ScenarioAndScriptTableTest.java:207

fitnesse.testsystems.slim.tables.ScriptTableExtensionTest > symbolReplacement FAILED
    org.junit.ComparisonFailure at ScriptTableExtensionTest.java:96

fitnesse.testsystems.slim.tables.ScriptTableExtensionTest > sameSymbolTwiceReplacement FAILED
    org.junit.ComparisonFailure at ScriptTableExtensionTest.java:96

fitnesse.testsystems.slim.tables.ScriptTableTest > symbolReplacement FAILED
    org.junit.ComparisonFailure at ScriptTableTest.java:116

fitnesse.testsystems.slim.tables.ScriptTableTest > showDoesEscapes FAILED
    java.lang.AssertionError at ScriptTableTest.java:697

fitnesse.testsystems.slim.tables.ScriptTableTest > sameSymbolTwiceReplacement FAILED
    org.junit.ComparisonFailure at ScriptTableTest.java:116

fhoeben:
I also believe some tests may show an impact. See: https://travis-ci.org/fhoeben/fitnesse/jobs/598610503

tcnh:
Yes, there's a bug in intellij 2019.2.1 that results in no tests being executed when using gradle... the build then reports success :(

I'll look into it, but I may need some time to find a moment to do so

PR #1245, opened 2019-11-05
 Add support to execute Java static methods from slim 

six42:
Often static methods are used as factories to create class instances.
These factories can now be called from slim scripts


src/fitnesse/slim/fixtureInteraction/SimpleInteraction.java
```diff
@@ -157,8 +178,10 @@ private boolean hasConstructorArgsTypes(final Class<?>[] constructorArgs, final
     try {
       //from fixture call we get only String values
       return ConverterSupport.convertArgs(args, argumentTypes);
-    } catch (SlimError ex) {
+    } catch (Throwable ex) {
```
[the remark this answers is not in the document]
six42: NumberFormatException for example. See the newly added test page.
convertArgs raises a standard exception if no converter has been found. But if the conversion of the data fails any exception might be raised.

src/fitnesse/slim/fixtureInteraction/SimpleInteraction.java
```diff
@@ -262,7 +306,7 @@ public Object methodInvoke(Method method, Object instance, Object... convertedAr
       }
     } catch (IllegalArgumentException e) {
       throw new RuntimeException("Bad call of: " + method.getDeclaringClass().getName() + "." + method.getName()
-              + ". On instance of: " + instance.getClass().getName(), e);
+              + ". On instance of: " + ((instance == null) ? "null" : instance.getClass().getName()), e);
```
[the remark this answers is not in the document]
six42: thanks good point. changed it.

src/fitnesse/slim/fixtureInteraction/SimpleInteraction.java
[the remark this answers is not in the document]
fhoeben: I understand. But I'm wondering whether it is desirable to completely ignore the exception if a constructor is found, the arguments could be convertible but the value(s) cause an error? Would it not maybe be better to report this problem (i.e. throw the exception)?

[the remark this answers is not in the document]
six42: convertArgs is called twice. 
 1.  Here (line 180), just to find a suitable constructor. The result is never used
 2. on line 64 in newInstance. This result is used to instantiate the class. Any error at this time is raised and visible in Slim.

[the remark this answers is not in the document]
six42: Therefore ignoring the exception here is ok.

fhoeben:
on [travis](https://travis-ci.org/fhoeben/fitnesse/jobs/609727351) I see a failing test case (FitNesse.SuiteAcceptanceTests.SuiteSlimTests.TestSystemSlimSuite.JavaSlimServer.CallStaticJavaMethod). any thoughts on what's causing that?

fhoeben:
The problem in the test is probably that on travis the timezone is different (causing a different result when displaying the data (timezone is not CET but something else)

src/fitnesse/slim/fixtureInteraction/SimpleInteraction.java
```diff
@@ -183,10 +211,31 @@ public MethodExecutionResult findAndInvoke(String methodName, Object instance, O
...
+    Method method;
+    int i = methodName.lastIndexOf('.');
+    if (i >=0) {
+      // Static Method
+      String className = methodName.substring(0, i);
+      String staticMethodName = methodName.substring(i+1);
+      Class<?> clazz = searchPathsForClass(paths, className);
+      if (clazz != null) {
+        method = findMatchingMethod(new String[]{staticMethodName}, clazz.getMethods(), args.length);
+	return this.invokeMethod(null, method, args);
```
fhoeben: Why is this line not indented more?

More importantly: I believe you should check whether method is not null before trying to invoke it...

six42: good point, added a check

src/fitnesse/slim/SlimError.java
```diff
@@ -4,16 +4,58 @@
...
+  public SlimError(String s, Throwable throwable, String tag, boolean prettyPrint) {
+      super(prettyPrint ? makeSlimMessage(s, tag) : s, throwable);
+      this.prettyPrint=prettyPrint;
+    }
+
   public SlimError(Throwable e) {
     this(e.getClass().getName() + " " + e.getMessage());
   }
+
+//  @Override
```
fhoeben: Do you intend to remove this, or should it be uncommented?

src/fitnesse/slim/fixtureInteraction/SimpleInteraction.java
```diff
@@ -261,17 +266,39 @@ private Method findMatchingMethod(String methodName, Method[] methods, int nArgs
...
+      Object[] convertedArgs = null;
+      Object retval = null;
+      String methodName = method.getDeclaringClass().getName() + "." + MethodExecutionResult.methodToString(method)
+                          + "." + ((instance == null) ? "" : " On instance of: " + instance.getClass().getName());
+      Class<?> retType = method.getReturnType();
+      try {
+        convertedArgs = convertArgs(method, args);
+      } catch (SlimError e) {
+	throw e;
+	// TODO: create test cases for this to check that it works and is displayed correctly in the result
```
fhoeben: Do you intend to address this TODO (and the one below)?

src/fitnesse/testsystems/slim/results/SlimExceptionResult.java
```diff
@@ -95,7 +95,11 @@ private String translateExceptionMessage(String exceptionMessage) {
       case COULD_NOT_INVOKE_CONSTRUCTOR:
         return "Could not invoke constructor for " + tokens[1];
       case NO_METHOD_IN_CLASS:
-        return String.format("Method %s not found in %s", tokens[1], tokens[2]);
+	if (tokens.length == 3){ // Legacy from Slim.Version <= 0.5
```
fhoeben: Can you elaborate on why this check is needed?

src/fitnesse/testsystems/slim/results/SlimExceptionResult.java
```diff
@@ -95,7 +95,11 @@ private String translateExceptionMessage(String exceptionMessage) {
       case COULD_NOT_INVOKE_CONSTRUCTOR:
         return "Could not invoke constructor for " + tokens[1];
       case NO_METHOD_IN_CLASS:
-        return String.format("Method %s not found in %s", tokens[1], tokens[2]);
+	if (tokens.length == 3){ // Legacy from Slim.Version <= 0.5 
+          return String.format("Method %s not found in %s", tokens[1], tokens[2]);
+	} else {
+	  return exceptionMessage.substring(exceptionMessage.indexOf(" ") + 1);
```
fhoeben: Why do we only want the part of the string after the first space? A comment in the code seems in order.

src/fitnesse/slim/fixtureInteraction/SimpleInteraction.java
```diff
@@ -261,17 +266,39 @@ private Method findMatchingMethod(String methodName, Method[] methods, int nArgs
...
+	throw e;
+	// TODO: create test cases for this to check that it works and is displayed correctly in the result 
+        //throw new SlimError(e.getMessage()+ "\n\n"+ "You may want to check if the correct method signature has been identified by Slim:\n" + methodName + "\n", e);
+      }
+      try {
+        retval = callMethod(instance, method, convertedArgs);
+      } catch (Exception e) {
+        // We can't modify the exception text as it could be an exception coming from the SUT
+	// TODO print this only if there are at least to methods with the same name and number of parameters  
+        System.out.println("InvokeMethod failed during invocation, you may want to check if the correct method signature has been identified by Slim: "  + " : " + methodName);
```
fhoeben: Instead of writing to System.out it seems more appropriate to create a logger (e.g. `  private static final Logger LOG = Logger.getLogger(SimpleInteraction.class.getName());
`) and use `LOG.warn(...`.

src/fitnesse/testsystems/slim/results/SlimExceptionResult.java
six42: This is old undocumented functionality. I added a new wiki testpage to explain why and improve documentation. See the commit 
"added Unit test of SlimExceptionResult"

src/fitnesse/testsystems/slim/results/SlimExceptionResult.java
six42: This is old undocumented functionality. I added a new wiki testpage to explain why and improve documentation. See the commit
"added Unit test of SlimExceptionResult"

src/fitnesse/slim/fixtureInteraction/SimpleInteraction.java
six42: Hi @fhoeben, could you run travis-ci. I would like to see if the fix to the test case I made worked.
Your other open comments I will rework or remove depending on my time in the next week.

six42:
Hi @fhoeben, could you run travis-ci. I would like to see if the fix to the test case I made worked.
Your other open comments I will rework or remove depending on my time in the next week.

src/fitnesse/slim/fixtureInteraction/SimpleInteraction.java
six42: the SlimService never uses a Logger. I assume to not interfere with the system under test.
Nevertheless you are right I will remove this or make it only appear in verbose mode.

six42:
Hi @fhoeben,
closed your final review comments. Let me know if anything else is required to merge.

src/fitnesse/slim/fixtureInteraction/SimpleInteraction.java
```diff
@@ -183,10 +217,35 @@ public MethodExecutionResult findAndInvoke(String methodName, Object instance, O
...
+    int i = methodName.lastIndexOf('.');
+    if (i >=0) {
+      // Static Method
+      String className = methodName.substring(0, i);
+      String staticMethodName = methodName.substring(i+1);
+      Class<?> clazz = searchPathsForClass(paths, className);
+      if (clazz != null) {
+        method = findMatchingMethod(new String[]{staticMethodName}, clazz.getMethods(), args.length);
+        if (method != null) {
+	  return this.invokeMethod(null, method, args);
```
fhoeben: Why is line not indented?

src/fitnesse/slim/fixtureInteraction/SimpleInteraction.java
```diff
@@ -23,19 +27,44 @@
...
+          IllegalAccessException, InvocationTargetException{
+    pathsCache = paths;
+    Class<?> k = null;
+    try{
+      k = searchPathsForClass(paths, className);
+    }
+    catch (SlimError errorClassNotFound) {
+      try {
+    	MethodExecutionResult mER = invokeStaticMethod(className, paths, args);
+  	    if (mER != null) {
```
fhoeben: Can you check the indentation of this method?

src/fitnesse/slim/SlimError.java
```diff
@@ -4,16 +4,53 @@
...
+    }
+
   public SlimError(Throwable e) {
     this(e.getClass().getName() + " " + e.getMessage());
   }
+
+  public static String makeSlimMessage(String msg, String tag){
+    StringBuilder sb = new StringBuilder();
+    sb.append(SlimVersion.PRETTY_PRINT_TAG_START);
+    if (tag != null && !tag.isEmpty()) {
```
fhoeben: check indentation

src/fitnesse/slim/SlimError.java
```diff
@@ -4,16 +4,53 @@
...
+	  sb.append(" ");
+	}
+    if (msg != null && !msg.isEmpty()) {
+      sb.append(msg);
+    }
+    sb.append(SlimVersion.PRETTY_PRINT_TAG_END);
+    return sb.toString();
+  }
+  public static String extractSlimMessage(String msg){
+    msg = msg.replaceFirst(".*"+SlimVersion.PRETTY_PRINT_TAG_START, "");
```
fhoeben:
```suggestion
    msg = msg.replaceFirst(".*" + SlimVersion.PRETTY_PRINT_TAG_START, "");
```

src/fitnesse/slim/MethodExecutionResult.java
```diff
@@ -51,13 +57,36 @@ public boolean hasResult() {
...
+
+	    public InvalidParameters(String methodName, Exception conversonException) {
+	      super(null, null);
+	      this.methodName = methodName;
+	      this.conversionException = conversonException;
+	    }
+
+	    @Override
+	    public Object returnValue() {
+	      throw new SlimError(SlimError.extractSlimMessage(conversionException.getMessage())+ "\nTryed to invoke: " + methodName, conversionException, "", true);
```
fhoeben:
```suggestion
	      throw new SlimError(SlimError.extractSlimMessage(conversionException.getMessage())+ "\nTried to invoke: " + methodName, conversionException, "", true);	
```
This also impact so unit tests

six42:
I finally found where the intention setting was wrong.
Should all be good now.

fhoeben:
@six42 it seemed to work fine in the branch, but the current (after merging it in) master seems to fail consistently. Can you take a look whether you see why (did I break something with the squash merge)?

fhoeben:
I found the problematic test `ParameterConversionExceptions`: 
`|and cell                           |expected .H..span..span class="diff".ello            |has result      |fail                                     |`
fails

You probably wrote that test before I fixed the way 'diffs' were displayed (to fix the failure navigator). So the expected html pattern is no longer correct.
I'll see whether I can fix this, although I have to say I can't quite understand why/what this row is exactly testing.

fhoeben:
@six42 can you review whether my change in https://github.com/unclebob/fitnesse/commit/736a36113e08fba480ce76c1d0e04dbbec50b0c7 left the intention of the test intact?

six42:
Sure, I will have a look in the next days

six42:
Hi @fhoeben,
your fix looks good. Many Thanks, 

That's what is expected and I need to test the red (failed) part:

Bildschirmfoto von 2020-01-04 21-10-16

In HTML it is
` class="fail">expected [H<span class="diff">ello<
`

I am happy to learn if you know a smarter way to test for it than this:

`|ensure | content matches | class="fail">expected .H.span class="diff".ello<|
`

fhoeben:
Good to head I adjusted the test properly.

I don't quite understand why you need to create a 'dynamic' page during this test and check how the failure is displayed. (I do see why its needed to check the error/exception messages) It seems like a normal wiki page would be sufficient to test the constructor being used. You didn't change anything on how failures are displayed, did you? So why not do this part of the 'TestPage' as a normal page that is run as part of the test suite?
```
Would expect that this uses the String(String) constructor but it rather uses String(Character) and truncates the string

|script|String|Hello|
|check|to string|H|
|check|get class|class java.lang.String|
```

Actually I thought that the ability to supply constructor parameters was already present, and this table checks normal instance methods, so why did you add this check here.
I agree with the text you placed above the table, so it seems better to remove this altogether and create an issue for the fact the 'wrong' constructor is used and add the table (with 'Hello' as expected value) to that issue...

six42:
This is a negative test case page. 
If I would use a normal test page it would fail and with it the build.
So I need to call it as a 'dynamic' page. And check that the right 'failure' happen.

The page was created to test for the exceptions which can happen during method calling.
The one TODO at the end is special. As you correctly said it is documenting an open issue. 
I would call it a "known limitation" as it seems to be minor to me.

It is a question of style how to document know limitations.
I prefer to do it in FitNesse directly than creating an issue in GitHub.
Having it in FitNesse gives the advantage that:

- It is documented next to where the principal functionality is described.
- Its presence is tested in each release. If a newer or different Java or FitNesse version fixes it this is discovered automatically and ensures that the documentation (wiki) gets updated.
- In case somebody wants to fix it he has directly the test case to work with. 

Having an issue created in GitHub which never gets fixed as it is of low priority seems to  me not the right approach. In addition It would require much more manual effort. Every year or so somebody would need to check if the issue is still relevant or if the issue can be closed in GitHub. 

In summary this is why people should use FitNesse. :)

fhoeben:
I agree with your points on an issue being less useful than a test that runs all the time. 

My only real remark here was that the TODO at the end did not need to be a negative test to document this behaviour. The sample table I gave tested/documented exactly the same behaviour in a positive test, that could be added to the test set separately. (The fact that you get an 'H' can be tested explicitly and you could even add a 'check not' with 'Hello'). That would serve as easier to find/understand/maintain documentation of this behaviour that the current negative setup using a dynamic page.

Issue #1148, opened 2018-07-18
java.lang.StringIndexOutOfBoundsException when user tries to create an empty symbolic link


orlyk:
When user tries to create a symbolic link with empty fields "Name" & "Path to page" and clicks button "Create/Replace" there is an Error Occurred (java.lang.StringIndexOutOfBoundsException: String index out of range: 0)

![screen shot 2018-07-18 at 4 26 09 pm](https://user-images.githubusercontent.com/5757189/42884553-50f61474-8aa7-11e8-8e83-bb06ab28c70c.png)

![screen shot 2018-07-18 at 4 26 36 pm](https://user-images.githubusercontent.com/5757189/42884569-5dee3c7e-8aa7-11e8-9f67-8da2f3998758.png)

fhoeben:
fixed in 20200205

PR #1260, opened 2020-02-01
Add usefull message to user.

superzaky:
This is a fix regarding this issue: https://github.com/unclebob/fitnesse/issues/1148


superzaky:
> Pretty good, but I thought it would be nice if the browser would remind the user to fill both fields before submitting the form. What do you think of that?

superzaky:
> Pretty good, but I thought it would be nice if the browser would remind the user to fill both fields before submitting the form. What do you think of that?

Sounds good I already see you've added some code. So this is already sufficient? (the screenshot below is from your code) :
image

PR #1262, opened 2020-02-08
Reduce duplicated code.

superzaky:


src/fitnesse/services/PageService.java
```diff
@@ -0,0 +1,31 @@
...
+
+import fitnesse.FitNesseContext;
+import fitnesse.http.Request;
+import fitnesse.http.Response;
+import fitnesse.responders.NotFoundResponder;
+import fitnesse.wiki.PathParser;
+import fitnesse.wiki.WikiPage;
+import fitnesse.wiki.WikiPagePath;
+
+public class PageService {
```
fhoeben: Instead of creating this new service, maybe it would be better to see whether you can make the responders you modified extend `BasicResponder`. It provides a `makeResponse()` implementation that seems to map nicely on what is needed for these responders including a `pageNotFoundResponse()` method....

src/fitnesse/responders/versions/VersionSelectionResponder.java
```diff
@@ -29,25 +29,23 @@
...
   public Response makeResponse(FitNesseContext context, Request request) throws Exception {
-    SimpleResponse response = new SimpleResponse();
-    String resource = request.getResource();
-    WikiPagePath path = PathParser.parse(resource);
-    WikiPage page = context.getRootPage().getPageCrawler().getPage(path);
-    if (page == null)
-      return new NotFoundResponder().makeResponse(context, request);
+    PageService pageService = new PageService(context, request);
+    WikiPage page = pageService.getPage();
+    pageService.pageIsNull(page);
```
fhoeben: Previously the method returned if page is null, now it just continues. That's not good.

If no test shows that this is broken now, please add such a test.

src/fitnesse/responders/editing/SavePropertiesResponder.java
```diff
@@ -20,20 +20,19 @@
...
   public Response makeResponse(FitNesseContext context, Request request) throws Exception {
-    SimpleResponse response = new SimpleResponse();
-    String resource = request.getResource();
-    WikiPagePath path = PathParser.parse(resource);
-    WikiPage page = context.getRootPage().getPageCrawler().getPage(path);
-    if (page == null)
-      return new NotFoundResponder().makeResponse(context, request);
+    PageService pageService = new PageService(context, request);
+    WikiPage page = pageService.getPage();
+    pageService.pageIsNull(page);
```
fhoeben: Previously the method returned if page is null, now it just continues. That's not good.

If no test shows that this is broken now, please add such a test.

src/fitnesse/responders/versions/VersionSelectionResponder.java
superzaky: I've checked the code and from what I've seen is that the page object can **never** be null. Otherwise the code would stop at `PathParser.parse(resource)` because of resource object being null. A page originates from a resource. For this reason I'm going to remove the null check for a page.

src/fitnesse/responders/editing/SavePropertiesResponder.java
superzaky: I've checked the code and from what I've seen is that the page object can **never** be null. Otherwise the code would stop at `PathParser.parse(resource)` because of resource object being null. A page originates from a resource. For this reason I'm going to remove the null check for a page.

src/fitnesse/responders/DisabledResponder.java
```diff
@@ -17,7 +17,7 @@ protected String contentFrom(FitNesseContext context, Request request, WikiPage
     return prepareResponseDocument(context).html(request);
   }
 
-  private HtmlPage prepareResponseDocument(FitNesseContext context) {
+  protected HtmlPage prepareResponseDocument(FitNesseContext context) {
```
fhoeben: Why do you leave this method here (instead of just having the class use the new method you created in BasicResponder)?

src/fitnesse/responders/versions/VersionSelectionResponder.java
```diff
@@ -12,47 +12,47 @@
...
     html.setMainTemplate("versionSelection");
 
+    SimpleResponse response = new SimpleResponse();
     response.setContent(html.html(request));
 
     return response;
   }
 
+  @Override
+  protected String contentFrom(FitNesseContext context, Request request, WikiPage requestedPage) {
```
fhoeben: Where/when is this method called (as you overridden `BasicResponder. makeResponse()`)

src/fitnesse/responders/editing/SavePropertiesResponder.java
```diff
@@ -6,38 +6,39 @@
...
     }
     context.recentChanges.updateRecentChanges(page);
-    response.redirect(context.contextRoot, resource);
+    response.redirect(context.contextRoot, request.getResource());
 
     return response;
   }
 
+  @Override
+  protected String contentFrom(FitNesseContext context, Request request, WikiPage requestedPage) {
```
fhoeben: Where/when is this method called (as you overridden `BasicResponder. makeResponse()`)

Maybe this implementation can be moved to BasicResponder? I believe it is also present in `DisabledResponder`, `PageDataWikiPageResponder` and `DefaultResponder`

src/fitnesse/responders/DisabledResponder.java
superzaky: Because it has a different implementation. It doesn't uses the implementation of BasicResponder. I've added a `@Override` to make it clear.

src/fitnesse/responders/versions/VersionSelectionResponder.java
superzaky: I'm working now on another computer. Primarily I let Intellij implement the required with method and left the implementation with `null`. Which caused a null pointer exception because some code expects to receive a `String `object but instead receiving `null`. But on this computer I don't receive a null pointer exception. I also don't find any code invoking the `contentFrom()` method. So I'm going to remove it.

src/fitnesse/responders/editing/SavePropertiesResponder.java
superzaky: I'm working now on another computer. Primarily I let Intellij implement the required with method and left the implementation with `null`. Which caused a null pointer exception because some code expects to receive a `String `object but instead receiving `null`. But on this computer I don't receive a null pointer exception. I also don't find any code invoking the `contentFrom()` method. So I'm going to remove it.

Yes I will move the implementation to BasicResponder for other classes who does makes use of it 👍.

src/fitnesse/responders/DisabledResponder.java
```diff
@@ -17,7 +17,8 @@ protected String contentFrom(FitNesseContext context, Request request, WikiPage
     return prepareResponseDocument(context).html(request);
```
fhoeben: This method can be removed. It does that same thing as the method it overrides

superzaky: > This method can be removed. It does that same thing as the method it overrides

No `DisabledResponder` sets a different template which differs from `BasicResponder`.

fhoeben: I meant the method `contentFrom()` it, which you have not touched yet. Sorry I wasn't more clear. The implementation of `prepareResponseDocument()` is indeed different, but `contentFrom` not, I believe.

PR #1265, opened 2020-02-21
Add SonarQube.

superzaky:


fhoeben:
What does this get us, can we now get a report about the code quality? How/where?

fhoeben:
Thanks for adding a description of what to do. It is (of course) specific to people running Windows, but I believe it would be easily adjusted for other OSs. 
The remark about skipping the unit test throws me off a bit. I would expect that one would also run all tests (jUnit and FitNesse.AcceptanceSuite) and see their coverage in Sonar.

Having said that. I don't plan to run Sonar locally to track issues this way and without guidelines exactly what should be done with the output/report in Sonar I don't see other people picking it up and using it as part of a regular regime contributing to the project. So I don't see much added value in merging this PR in at the moment.

This is not to say you cannot run Sonar locally this way, use it to find issues and then submit PRs addressing those. PRs addressing issues are greatly appreciated and if you find them using Sonar and have time to address them that's great. Or even if you just create issue reports based for things you find that would already be great also.

When it turns out we can prevent many problems in future by running this analysis regularly (before each release, on each PR?) incorporating it into the regular routine and guidelines would be a logical step. But for me it is not at the moment.

PR #1268, opened 2020-02-24
Add new feature: suite partitioning

fhoeben:
With this PR FitNesse's suite responder (and the jUnit `FitNesseRunner`) gets a new feature: suite partitioning. 
Partitioning allows the set of tests in a suite to be split into multiple sub-sets (all tests are present in exactly one sub-set). The idea for making such a split is that a test execution on a CI/CD server can be sped up if its parts are done in parallel on multiple workers. The suite partitioning feature allows the suite to be split without requiring specific alterations on the tests themselves (e.g. no need to add tags or group tests in sub-wikis). And the split will remain 'balanced' (i.e. all partitions have approximately the same size) even when tests are added and removed from the suite.

Besides just running a certain partition from a suite there is also a new responder (partition) which provides insight into how tests would be partitioned, and ordered within a suite. This responder can also be useful when not splitting a suite (i.e. number of partitions is 1) to understand how a suite will be run. It lists all tests in a suite all the suite set ups and tear downs that will be added to the suite and the order in which they will be executed (taking into account which test systems they use), without actually running all tests.

This PR introduces a basic partition algorithm based on the number of tests, but it also allows plugins to define more advanced/elaborate mechanisms.

Below is the documentation added to FitNesse's user guide: 

-----

Suites can be partitioned, that is split into multiple parts, for instance to run each partition in parallel. By running the parts in parallel the total execution time can be reduced.
FitNesse is not thread-safe so each partition's run should be executed by a separate Java process. But this can be a very effective way to speed up the time needed for testing when tests do not interfere which each other. In CI/CD servers one can spread test execution across multiple workers and get very quick tests results, even when test sizes grow.

 There are multiple ways to select certain parts of a suite. SubWikiSuites and TagsAndFilters both allow suites to be split, but these are static. And if the goal is only to improve a test suite's execution time, it could be considered improper use of these mechanisms as they are intended to define functional aspects of tests.
A purpose-build mechanism to partition a suite is also available. When running a suite one can provide a `partitionCount` and `partitionIndex` parameter (zero-based) to indicate in how many parts the suite should be split, and which part to run now.
An example usage of this mechanism is a call to `http://<host>:<port>/<suite path>?suite&partitionCount=2&partitionIndex=0`. This would partition the suite in two, executing the first.

 So one can start multiple runs for the same suite, all using the same partition count, but a different partition index.
By starting 'partition count' runs all tests are run, without the need to manually define and maintain exactly which tests should go in which partition, even when tests are added and removed from the suite.

 **Partition Preview**

A specific partition responder is available which will list the partitions (and test order in them) that will be created for a specific suite and partition count. It will list all test pages in HTML (or tab-separated text) format indicating in which partition the test will be placed and what its order in the partition will be.
So one could for instance do `http://<host>:<port>/<suite path>?partition&partitionCount=5` to so see how that suite would be divided in 5 partitions.

Tab separated format can be requested by supplying the `format` parameter with value `tsv`, e.g. `http://<host>:<port>/<suite path>?partition&partitionCount=5&format=tsv`.

 **Partition Index File**

One can also supply a file containing a partitioning in the tab-separated format created by the partition preview responder when running a suite. This file can list the desired partitioning for all known test pages. Any other pages found in the suite will be spread evenly across all partitions.
The filename for such a file must be supplied to the suite responder using the `partitionIndexFile` parameter. The file will be looked up inside the wiki's files section.

 So for example a call to `http://<host>:<port>/<suite path>?suite&partitionCount=2&partitionIndex=1&partitionIndexFile=partitions.tsv`. This would partition the tests in the suite in two groups using the division described in a file called `partitions.tsv` in the files section, executing the second group.
The file's content could look something like this (where whitespace between columns should be tab characters):
```
Page	Partition	Test System	Order
FitNesse.SuiteAcceptanceTests.SuiteAuthenticationTests.AlwaysSecureOperation	0	slim	0
FitNesse.SuiteAcceptanceTests.SuiteAuthenticationTests.SecureReadOperations	1	slim	0
FitNesse.SuiteAcceptanceTests.SuiteAuthenticationTests.SecureTestOperations	1	slim	1
```

 **Partitioning Algorithm**

 The standard partitioning algorithm tries to make partitions with the same number of tests. It will, for instance, for a partition count of 4:
 * split a suite of 100 tests into 4 partitions of 25 elements and
 * a suite of 102 elements would be split into 2 partitions of 26 elements and 2 of 25.

 Plugins can tailor this behaviour by registering a custom 'test run factory'. For instance when historic run times of tests are known a plugin could use this information to create partitions with an equal expected run time.

**Usage via FitNesseRunner**

 When running tests from jUnit using the `FitNesseRunner` one can use this feature by applying the `@Partition`, and possibly the `@PartitionFile`, annotation to the test class.

### DIFF
--- a/src/fitnesse/testrunner/run/PagePositions.java
+++ b/src/fitnesse/testrunner/run/PagePositions.java
@@ -156,7 +156,7 @@ public class PagePositions {
     }
   }
 
-  protected String formatDimension(Object dimension) {
+  public String formatDimension(Object dimension) {
     if (dimension instanceof WikiPageIdentity) {
       return ((WikiPageIdentity) dimension).testSystem();
     } else {
--- a/src/fitnesse/testrunner/run/PagePositions.java
+++ b/src/fitnesse/testrunner/run/PagePositions.java
@@ -9,6 +9,7 @@ import java.io.Reader;
 import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -28,7 +29,7 @@ public class PagePositions {
 
   private final Map<List<Object>, List<Object>> groupCache = new HashMap<>();
   private String lineSep = "\n";
-  private List<String> groupNames = new ArrayList<>();
+  private List<String> groupNames = new ArrayList<>(2);
   private Map<String, List<PagePosition>> positions = new LinkedHashMap<>();
 
   public List<String> getGroupNames() {
@@ -89,16 +90,8 @@ public class PagePositions {
 
   protected void parseHeader(BufferedReader reader, String groupSep) throws IOException {
     splitNextLine(reader, groupSep)
-      .map(parts -> {
-        int count = parts.length;
-        if (count < 3) throw new IllegalArgumentException("Expected at least 2 columns, got: " + count);
-        groupNames.clear();
-        for (int i = 1; i < count - 1; i++) {
-          groupNames.add(parts[i]);
-        }
-        return groupNames;
-      }).
-      orElseThrow(() -> new IllegalArgumentException("No header line"));
+      .map(parts -> Collections.addAll(groupNames, stripHeadAndTail(parts)))
+      .orElseThrow(() -> new IllegalArgumentException("No header line"));
   }
 
   protected void parseRows(BufferedReader reader, String groupSep) throws IOException {
@@ -108,12 +101,16 @@ public class PagePositions {
         break;
       }
       String[] parts = linePresent.get();
-      int count = parts.length;
-      if (count < 3) throw new IllegalArgumentException("Expected at least 2 columns, got: " + count);
-      Integer pos = Integer.valueOf(parts[count - 1]);
-      List<Object> group = asList((Object[])Arrays.copyOfRange(parts, 1, count - 1));
+      List<Object> group = asList((Object[]) stripHeadAndTail(parts));
+      Integer pos = Integer.valueOf(parts[parts.length - 1]);
       addPosition(parts[0], group, pos);
     }
+  }
+
+  protected String[] stripHeadAndTail(String[] parts) {
+    int count = parts.length;
+    if (count < 3) throw new IllegalArgumentException("Expected at least 2 columns, got: " + count);
+    return Arrays.copyOfRange(parts, 1, count - 1);
   }
 
   protected Optional<String[]> splitNextLine(BufferedReader reader, String groupSep) throws IOException {
--- a/src/fitnesse/testrunner/run/PagePositions.java
+++ b/src/fitnesse/testrunner/run/PagePositions.java
@@ -1,0 +1,166 @@
+package fitnesse.testrunner.run;
+
+import fitnesse.testrunner.WikiPageIdentity;
+import org.apache.velocity.util.StringBuilderWriter;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.Reader;
+import java.io.Writer;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static java.util.Arrays.asList;
+
+/**
+ * Describes the order in which pages will be executed in a test run.
+ */
+public class PagePositions {
+  public static final String PAGE_HEADER = "Page";
+  public static final String ORDER_HEADER = "Order";
+
+  private final Map<List<Object>, List<Object>> groupCache = new HashMap<>();
+  private String lineSep = "\n";
+  private List<String> groupNames = new ArrayList<>();
+  private Map<String, List<PagePosition>> positions = new LinkedHashMap<>();
+
+  public List<String> getGroupNames() {
+    return groupNames;
+  }
+
+  public Integer getGroupIndex(String name) {
+    return groupNames.indexOf(name);
+  }
+
+  public List<String> getPages() {
+    return new ArrayList<>(positions.keySet());
+  }
+
+  public List<PagePosition> getPositions(String page) {
+    return positions.computeIfAbsent(page, p -> new LinkedList<>());
+  }
+
+  public boolean hasPositions(String page) {
+    return positions.containsKey(page);
+  }
+
+  public void addPosition(String page, List<Object> group, int positionInGroup) {
+    // re-use group objects if same elements are used to define group
+    List<Object> groupFromCache = groupCache.computeIfAbsent(group, g -> group);
+    getPositions(page).add(new PagePosition(groupFromCache, positionInGroup));
+  }
+
+  public Comparator<String> createByPositionInGroupComparator() {
+    Map<String, Integer> testOrdering = new HashMap<>();
+    for (String page : getPages()) {
+      for (PagePosition pos : getPositions(page)) {
+        testOrdering.put(page, pos.getPositionInGroup());
+      }
+    }
+    return (page1, page2) -> {
+      int pos1 = testOrdering.getOrDefault(page1, -1);
+      int pos2 = testOrdering.getOrDefault(page2, -1);
+      return pos1 == pos2 ? page1.compareTo(page2) : pos1 - pos2;
+    };
+  }
+
+  public void appendTo(Writer writer, String groupSep) throws IOException {
+    appendHeader(writer, groupSep);
+    for (Map.Entry<String, List<PagePosition>> entry : positions.entrySet()) {
+      appendIndices(writer, entry.getKey(), entry.getValue(), groupSep);
+    }
+    writer.flush();
+  }
+
+  public static PagePositions parseFrom(Reader reader, String groupSep) throws IOException {
+    BufferedReader b = new BufferedReader(reader);
+    PagePositions positions = new PagePositions();
+    positions.parseHeader(b, groupSep);
+    positions.parseRows(b, groupSep);
+    return positions;
+  }
+
+  protected void parseHeader(BufferedReader reader, String groupSep) throws IOException {
+    splitNextLine(reader, groupSep)
+      .map(parts -> {
+        int count = parts.length;
+        if (count < 3) throw new IllegalArgumentException("Expected at least 2 columns, got: " + count);
+        groupNames.clear();
+        for (int i = 1; i < count - 1; i++) {
+          groupNames.add(parts[i]);
+        }
+        return groupNames;
+      }).
+      orElseThrow(() -> new IllegalArgumentException("No header line"));
+  }
+
+  protected void parseRows(BufferedReader reader, String groupSep) throws IOException {
+    while(true) {
+      Optional<String[]> linePresent = splitNextLine(reader, groupSep);
+      if (!linePresent.isPresent()) {
+        break;
+      }
+      String[] parts = linePresent.get();
+      int count = parts.length;
+      if (count < 3) throw new IllegalArgumentException("Expected at least 2 columns, got: " + count);
+      Integer pos = Integer.valueOf(parts[count - 1]);
+      List<Object> group = asList((Object[])Arrays.copyOfRange(parts, 1, count - 1));
+      addPosition(parts[0], group, pos);
+    }
+  }
+
+  protected Optional<String[]> splitNextLine(BufferedReader reader, String groupSep) throws IOException {
+    return Optional.ofNullable(reader.readLine()).map(l -> l.split(groupSep));
+  }
+
+  @Override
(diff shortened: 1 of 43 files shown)


tcnh:
Cool feature! Some initial thoughts:
 * I think we need some way to configure behaviour of SuiteSetUp/SuiteTearDown. As these are normally executed once, I'd expect them to also be executed only once (before and after all partitions have been executed - this is not trivial.... I get that) if the suite is divided into multiple partitions? Currently, each partition will execute the SuiteSetUp and SuiteTearDown. This may lead to serious complications when using these special pages for environment setup/cleanup.
* I'm unable to see partitioning in action. I added this branch to a default hsac fitnesse project ant executed both `http://localhost:9090/HsacExamples.SlimTests.HttpTests?suite&partitionCount=2&partionIndex=0` and `http://localhost:9090/HsacExamples.SlimTests.HttpTests?suite&partitionCount=2&partionIndex=1` 
I expected to see partitioning like the output file from `http://localhost:9090/HsacExamples.SlimTests.HttpTests?partition&partitionCount=2`, but both partitions seem to execute everything. Did I miss some configuration?
Also tried executing `http://localhost:9090/HsacExamples.SlimTests.HttpTests?suite&partitionCount=2&partionIndex=1&partitionIndexFile=partitionindex.tsv` where partitionIndex.tsv is the output of `?partition&partitionCount=5` to no avail.
* The ?partition responder now returns a partitioning TSV. Maybe it's nice to make the output configurable (tsv als partitioning output, html table for a quick look)? That ma be particularly nice when testing roll your own partitioning mechanisms?

fhoeben:
@tcnh 

**SuiteSetUp/SuiteTearDown**
SuiteSetUp and -TearDown are executed as normal **for each partition** this means: first the tests in the partition are determined and then, based on their suite structure, all relevant set up and tear down pages are added to this partition. So if no test from a suite is selected for a specific partition its suite set up and tear down are not executed. If one or more tests are included in the partition they are added. Each partition is expected to run completely isolated and there MUST run all initialisation for the fixtures and SUT environment setup and cleanup actions. 

Your statement about environment setup/cleanup to me suggests you have an environment where there is a single SUT instance that will be the subject of tests by all partitions. This is a risky situation to use partitions: your tests have to be written very well not to interfere with each other (e.g. what happens if a test checking you can add an item runs in one partition while another partition is checking that all items can be deleted). My goal for the partitions is that each tests an independent environment (e.g. separate docker container, uses a different tenant in a multitenant system), so there is no single setup but it needs to be repeated for each environment.

If only the environment setup conflicts or maybe a few tests can not be run in parallel, but the great majority can, I would suggest to spit your tests (for instance via suite tags) in a set that is run sequentially first and after that is completed the remaining tests are run divided across n partitions (and the test suites in those partitions do NOT execute the SuiteSetUp which configures the environment). I've used this approach in the past (with partitions configured using suite tags) and that worked quite well.

**partitioning in action**
I did not try this feature on the HSAC-fixtures yet (;-)) only on FitNesse's acceptance tests. There I saw that calling http://localhost:8001/FitNesse.SuiteAcceptanceTests?suite&partitionCount=10&partitionIndex=9 gives me 30 something tests (out of a total of approx 320). Did you make both calls at the same time, or sequentially? 
Calling them at the same time gives unreliable results as FitNesse is not thread safe. To run these tests in parallel you need to start 2 FitNesse instances (each listening on its own port) and have each execute a partition. So for instance start one of port 9090 and another on port 9091 and then do `http://localhost:9090/HsacExamples.SlimTests.HttpTests?suite&partitionCount=2&partionIndex=0` and `http://localhost:9091/HsacExamples.SlimTests.HttpTests?suite&partitionCount=2&partionIndex=1`

** partition responder **
This responder is indeed created mainly as a quick feedback tool when messing with partitioning. I wanted to have something like it while I was building this feature, that's why I built it. I thought about supporting multiple formats, but I found I actually quite liked the tsv format in my browser also (and it copy-pastes nicely to excel also). Having a HTML table does not add that much value for me, unless there would be nice (JavaScript) sorting and filtering. But you are probably right tsv a bit strange default output format. Maybe a parameter should be added to indicate desired format (just like suite responder currently supports), with initial support for (basic) HTML and TSV support. I did not have time to add that so far, could you maybe.....?

fhoeben:
I had a typo in my original PR text (and @tcnh copied precisely) the parameter for the index to run must be `partitionIndex` and **not** `partionIndex`. Oops. That's why it didn't work...

tcnh:
I can';t seem to create a PR to merge my version of this branch into yours, but have a look at commit: https://github.com/unclebob/fitnesse/commit/d5ee58d228569de85ed90e73c1510883bb9272b5

If integrated, you can get a nice HTML preview of partioning and the option to download the partioning tsv file by setting the html key as request parameter i.e. `?partition&partitionCount=2&html`

src/fitnesse/testrunner/run/PagePositions.java
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  protected void parseHeader(BufferedReader reader, String groupSep) throws IOException {
+    splitNextLine(reader, groupSep)
+      .map(parts -> {
+        int count = parts.length;
+        if (count < 3) throw new IllegalArgumentException("Expected at least 2 columns, got: " + count);
+        groupNames.clear();
+        for (int i = 1; i < count - 1; i++) {
+          groupNames.add(parts[i]);
```
superzaky: Replace for loop with addAll method example:

public void makeCopies(String[] source) {
  this.array = Arrays.copyOf(source, source.length);
  Collections.addAll(this.list, source);
}

fhoeben: Wouldn't that also copy first element (which I don't add now)?

src/fitnesse/testrunner/run/PartitioningTestRunFactory.java
```diff
@@ -0,0 +1,103 @@
...
+    int partitionIndex = 0;
+    String partitionIndexStr = page.getVariable(PARTITION_INDEX_ARG);
+    if (StringUtils.isNotEmpty(partitionIndexStr)) {
+      partitionIndex = Integer.parseInt(partitionIndexStr);
+    }
+    return partitionIndex;
+  }
+
+  public Function<List<WikiPage>, ? extends PagesByTestSystem> getFactory() {
+    return factory;
```
superzaky: Remove usage of generic wildcard type and instead use a bounded parameter.

fhoeben: This is a bounded parameter, isn't it: the function may return any subclass of `PagesByTestSystem`. I don't want to force that the function returns `PagesByTestSystem` if any subclass suffices.

If I misunderstand you: can you make an explicit suggestion what to change this to?

src/fitnesse/testrunner/run/PagePositions.java
fhoeben: Did you mean:
```
        groupNames.clear();
        groupNames.addAll(Arrays.asList(Arrays.copyOfRange(parts, 1, count - 1)));
        return groupNames;
```
Maybe it's more efficient, but I don't really think it is as clear as the loop. Is it really better?

fhoeben: I believe I found a good way to prevent adding one by one

fhoeben:
@tcnh I've incorporated the HTML view (and made it the default format) in the partition responder (2nd line gets a bit strange in 'bootstrap-plus theme', but other than that seems fine)

tcnh:
Looks good. The bootstrap-plus issue is resolved in future bootstrap-plus versions as we never encountered tables with a thead before.
I'm not particularly charmed bij de maximum width and scrollbar-overflow when testcase paths become very long (as everywhere else, we just widen the table), but it's a bit of a corner case I guess. And in this case the partition is more relevant and should be on screen.
Maybe we should consider changing the column order to `partition, order, page, test system` in the html view?

fhoeben:
I like the page being first as that is what I recognise, so seeing that key before values (i.e. partition and order) is logical to me. But you are right the max-width and overflow are not great and I do want to see those values.

I introduced the max-width as a backstop for the acceptance tests have some VERY long names and I'm really not that interested in full page names so I don't want to have to scroll thousands of pixels right....

Real analysis does require a program or something like excel so you can sort/group/order/count I believe.

I try to change it so that the right columns have a fixed width, so the page name can take as much as available, instead of setting a max on its width...

fhoeben:
See below how the page renders with different browser widths. Good enough?

Screenshot 2020-03-02 at 14 26 48
Screenshot 2020-03-02 at 14 26 53

src/fitnesse/testrunner/run/PagePositions.java
superzaky: > 
> 
> Wouldn't that also copy first element (which I don't add now)?

That's correct.

superzaky: > 
> 
> Did you mean:
> 
> ```
>         groupNames.clear();
>         groupNames.addAll(Arrays.asList(Arrays.copyOfRange(parts, 1, count - 1)));
>         return groupNames;
> ```
> 
> Maybe it's more efficient, but I don't really think it is as clear as the loop. Is it really better?



> 
> 
> Did you mean:
> 
> ```
>         groupNames.clear();
>         groupNames.addAll(Arrays.asList(Arrays.copyOfRange(parts, 1, count - 1)));
>         return groupNames;
> ```
> 
> Maybe it's more efficient, but I don't really think it is as clear as the loop. Is it really better?

That depends how many elements the `parts[]` array contains. If it contains so much elements which would affect the performance. Then I would make a consideration otherwise the (whole) idea about partition running tests faster could be reduced because of a for loop. But again it depends on how many elements `parts[]` array contains. And I do agree that a for loop is more clear.

I added breakpoints in `parseHeader()` to check how many elements there are.

I used http://localhost:8001/FitNesse.SuiteAcceptanceTests?suite&partitionCount=10&partitionIndex=9

But the `parseHeader()` wasn't invoked. How does this method get invoked?

src/fitnesse/testrunner/run/PartitioningTestRunFactory.java
superzaky: > 
> 
> This is a bounded parameter, isn't it: the function may return any subclass of `PagesByTestSystem`. I don't want to force that the function returns `PagesByTestSystem` if any subclass suffices.
> 
> If I misunderstand you: can you make an explicit suggestion what to change this to?

No it isn't the parameter can be of any subclass, because of the wildcard type used as return type. To clarify what I mean I will share a screenshot:
image


In your case I would change it to: `Function<List<WikiPage>, PagesByTestSystem>`

src/fitnesse/testrunner/run/PagePositions.java
fhoeben: It is only used when loading a previously saved partition 'index file' e.g.
http://localhost:8001/FitNesse.SuiteAcceptanceTests?suite&partitionCount=10&partitionIndex=9&partitionIndexFile=indexes.tsv

But you have to create such a file first, and save it in the files sections.

See FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteResponderTests/PartitionPreviewResponder/PartitionPreviewBasedOnFile.wiki

Did you notice I already changed the code? (The number of headers values to get is normally 2)

src/fitnesse/testrunner/run/PartitioningTestRunFactory.java
fhoeben: Sorry I'm stopping this discussion now. I'm not changing it. A `Function<A,B>` indicates a function that will return an instance of type `B` this is fundamentally different than the type parameter of a List. 

This signature says: you get a function that takes a list WikiPages and returns a PagesByTestSystem or a subclass of PagesByTestSystem. 
(The type parameter of a List also says something about its input parameters, for instance the `add` method will accept only instances of the type supplied, so you don't know what you can add to a `List<? extends A>`. This is not the case for the Function interface.)

The consumer of this function is not any client that now has to deal with a function with a wildcard. The consumer of this function in the partitioning algorithm and it has a wildcard to allow customisation by users. So other functions can be used/supplied.

Johnny2Shoes:
Is there a way to use the Fitnesse wiki to trigger these tests? I was thinking of could my QA guys have a page with five links that they can open in 5 tabs to start the 5 partitions?

tcnh:
You could of course just put the necessary links on a wiki page.. but generally, doing scheduling and/or parallel execution is something I'd configure in my ci-tool 10/10 times

fhoeben:
I would also recommend a CI server, but you can run them using the browser: just add the parameters to the query parameters: `....?suite&partitionCount=5&partionIndex=1` for the first set, `....?suite&partitionCount=5&partionIndex=2` for the second, etc

Johnny2Shoes:
Yep - we have a CI server but our tests are a little flakey at the moment, so we want to get them stable before investing the effort right now.

Thanks - that's what I was missing. I was trying to put suite on the end of the string not the start!

PR #1271, opened 2020-03-03
Simplify Slim's table colouring

fhoeben:
Simplify Slim's table colouring both in code and functionally:
* instead of creating its own Slim table factory use the one from the FitNesse context (which includes table types installed by plugins)
* don't colour comment tables, or tables on static pages (but do colour ScenarioLibrary and SetUp/TearDown/SuiteSetUp/SuiteTearDown)
* don't try to find fixture classes using reflection:
  * Slim classes don't have to be on the wiki's classpath
  * assume tables without explicit table type are decision tables: it is Slim's default (this also ensures decision tables based on scenarios are coloured)
  * this is faster/simpler as Java reflection is slow (a cache is no longer required)
* don't get the content of all cells in all rows since only the first is used, read only that

I'm still not really happy with the way the table colouring decorator is set up. But I believe this is a step in the right direction.

In the future I would like to have a setup without a static singleton and where a first filtering of whether or not to a apply a decorator would be done at page level, instead of for each table on each page (for instance the test system is the same for all tables on a page after all).


fhoeben:
@joprsal @mwarhaftig I believe you have most experience with Slim colouring. Any thoughts?

tcnh:
>In the future I would like to have a setup without a static singleton and where a first filtering of whether or not to a apply a decorator would be done at page level, instead of for each table on each page (for instance the test system is the same for all tables on a page after all).

Does that mean that you'd like to see that tables can no longer be independently decorated? Reason I ask is that I have a backlog item that allows for visual identification of script/scenario/table template/comment/dt table types in bootstrap-plus, because the new decorator approch allows for that. If we're going to drop support for individual table styling, I'll have to drop that plan :)

fhoeben:
No I don't mean to drop support. I just want to set it up differently, getting rid of the singleton approach in the current default colouring class.

Maybe a bit more like a registry where a plugin can register its decorators, instead of having to put that directly on SymbolType itself.
And each decorator could indicate at page level whether it should be applied and then for which symbol types.

I'm quite happy with the current changes as decorators no longer get called for parent pages that are only being parsed to find variables (only the pages that are actually being converted to HTML get decorated), and the default decoration is simpler and faster (as it no longer uses Class.forName()).

joprsal:
Hi Fried,
I had a look and all changes seem a good improvement to the code. In terms of performance as well as readability.
There is no drop in functionality supported by the symbol decorators.
There might be a slight drop in features implemented in SlimTableDefaultColoring, but as you argued, there are good reasons for it (inadequate classpath assumptions). However, I do not fully understand the extent of non backwards compatible scenarios due to this.

joprsal:
With respect to your final remark about symbol decorators setup:
"...setup without a static singleton and where a first filtering of whether or not to a apply a decorator would be done at page level..."
Do I understand correctly that you would like the page-level filter, which is currently done in SlimTableDefaultColoring line 67, be somehow declared by the decorator itself (e.g. method parsedSymbolDecorator.isApplicableToPage(parsingPage/sourcePage))? This would then be used by HtmlTranslator once for the whole page and only applicable decorators would be given chance to handle individual symbols?

fhoeben:
@joprsal you understood my page level filtering correctly.

fhoeben:
@tcnh to do proper colouring of Slim tables I believe one has to take a different approach. You would have to have the page interpreted by Slim, so you also support alias tables, prefixes indicating table type etc. 
I believe one can probably do that, but you'd probably have to use a different approach than the current symbol decorators. As this would not something you would want to do first before a test (-suite) is actually run, in that case it would have to be integrated in Slim's current test execution. Only for rendering individual pages this would be an extra (new) step in the process.

In my plugin I have the 'slim coverage' test system. This already goes over all test pages in a suite to figure out which scenarios are used where. To do this it actually does a 'test' execution but it doesn't execute any of the calls on the fixtures. Something along those line can probably be done quite quickly for a single page, so the tables can be marked properly without duplicating the code to interpret the page (as the current default slim colouring does).

Issue #1280, opened 2020-04-15
Slow parsing in whole suite tree when using specfic variable syntax containing whole tables


sqeezy:
In my team we used the following syntax for defining whole tables in variables:

```
!define VariableName {|script|Adapter Name|CtorArgument|
|Method|Parameter|}
```

Every addition of such a variable slowed down the whole wiki down from the place where this was defined. The growth in parsing time seems exponential. When it got obvious (~30 s) we searched for the problem. Putting the closing curly brace on a new line fixed the problem.

```
!define VariableName {|script|Adapter Name|CtorArgument|
|Method|Parameter|
}
```

This seems like something that should either fail or work faster.

fhoeben:
Good find. No idea why it would get that slow. And how did you ever find the workaround?

Can you give some more details on what you encountered? 
- which FitNesse version
- what JVM
- which OS
- how much memory did the machine have, any special parameters to the JVM?
- what kind of complexity made it obvious?
  - can you describe the situation in which it was ~30s?
  - did it get worse by adding more pages, or only when using the variable more? Or only when you added more variables?
  - did it matter whether the variable was used in scenario libraries or other special pages (SetUp, SuiteSetUp)?
- anything else

sqeezy:
* **which FitNesse version**
v20160618
* **what JVM**
java version "1.8.0_25"
Java(TM) SE Runtime Environment (build 1.8.0_25-b18)
Java HotSpot(TM) 64-Bit Server VM (build 25.25-b02, mixed mode)
* **which OS**
Win Server 2008 R2
_Don't judge ;)_
* **how much memory did the machine have, any special parameters to the JVM?**
12 GB total
About half of that free
* **what kind of complexity made it obvious?**
* **can you describe the situation in which it was ~30s?**
* **did it get worse by adding more pages, or only when using the variable more? Or only when you added more variables?**
* **did it matter whether the variable was used in scenario libraries or other special pages (SetUp, SuiteSetUp)?**
We have a large (~700) suite of integration tests for one of our programs. It all was working well until a few days ago. First the testers thought it was a server issue and didn't think much of it. When it got to the point of it taking 30 seconds we went in and investigated.
Luckily we have version control for our suite so the point in time when it got slow was easily found. After that it was just thinking of what could be the problem and luck. After finding that moving the curly brace in one variable, I tried out moving them all. From that point on the whole app was as fast as new. The whole problem seems only to scale with number of variables with curly braces in the last line.
As for the structure:
[...]

PR #1282, opened 2020-04-18
Issue 1280

fhoeben:
Allow table symbol type to detect its line ends on close brace before a newline, so newline does not directly follow closing `|`. This can occur if table is used inside a `!define` 

#1280

### DIFF
--- a/src/fitnesse/wikitext/parser/Parser.java
+++ b/src/fitnesse/wikitext/parser/Parser.java
@@ -58,7 +58,7 @@ public class Parser {
         return scanner.peek(size, new ParseSpecification().provider(specification));
     }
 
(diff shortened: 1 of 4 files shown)


jediwhale:
I need to spend a bit of time remembering what I did when I originally wrote this! This looks like it fixes the "table in a define" issue. But, my initial concern is what does it do if the table is not in a define. (I haven't tried this yet), e.g.:
|a|b|}|c|d|

fhoeben:
It's not the easiest code to get your head around. So I'm grateful you can have a look at it.

Initially I did break closing braces inside tables. But now I believe I catch those correctly (and covered that in the unit tests on TableType). Let me know what you find.

jediwhale:
fixed with #1285

PR #1307, opened 2020-11-16
Allow tables and symbolProviders to be extended in plugins

tcnh:
This PR allows the Parser's table to be extended from plugins.

To make useful extensions (i.e. create a table that only parses certain symbols whilst supporting custom plugin symbols), it helps if Symbolproviders are a bit more flexible to configure. Therefore, I've added means to copy a provider (alongside the parent/child structure), so we can alter it wthout breaking existing implementation Also added a remove method that allows removing existing tyes from a provider.

Formatting has made big changes to this file (formatting was inconsistent) and replaced a star import. If those are unwanted, I'll revert them

### DIFF
--- a/src/fitnesse/wikitext/parser/SymbolProvider.java
+++ b/src/fitnesse/wikitext/parser/SymbolProvider.java
@@ -1,119 +1,141 @@
 package fitnesse.wikitext.parser;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 public class SymbolProvider {
-    public static final SymbolProvider refactoringProvider = new SymbolProvider( new SymbolType[] {
-            Alias.symbolType, SymbolType.OpenBracket, SymbolType.CloseBracket, Comment.symbolType, Image.symbolType,
-            Literal.symbolType, Preformat.symbolType, Link.symbolType, Path.symbolType, WikiWord.symbolType,
-            SymbolType.Newline, SymbolType.Whitespace
-    });
+  public static final SymbolProvider refactoringProvider = new SymbolProvider(new SymbolType[]{
+    Alias.symbolType, SymbolType.OpenBracket, SymbolType.CloseBracket, Comment.symbolType, Image.symbolType,
+    Literal.symbolType, Preformat.symbolType, Link.symbolType, Path.symbolType, WikiWord.symbolType,
+    SymbolType.Newline, SymbolType.Whitespace
+  });
 
-    public static final SymbolProvider wikiParsingProvider = new SymbolProvider( new SymbolType[] {
-            Link.symbolType, Table.symbolType,
-            new HashTable(),  new HeaderLine(), Literal.symbolType, Nesting.symbolType, new Collapsible(),
-            new AnchorName(), new Contents(), SymbolType.CenterLine, new Define(), new Help(),
-            new Include(), SymbolType.Meta, SymbolType.NoteLine, Path.symbolType, new PlainTextTable(),
-            See.symbolType, SymbolType.Style, new LastModified(), Image.symbolType,
-            new Today(), SymbolType.Delta,
-            new HorizontalRule(), SymbolType.CloseLiteral, SymbolType.Strike,
-            Alias.symbolType, SymbolType.UnorderedList, SymbolType.OrderedList, Comment.symbolType, SymbolType.Whitespace, SymbolType.CloseCollapsible,
-            SymbolType.Newline, SymbolType.Colon, SymbolType.Comma,
-            Evaluator.symbolType, SymbolType.CloseEvaluator, Variable.symbolType, Preformat.symbolType,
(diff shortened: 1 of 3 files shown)


fhoeben:
Can you add a test/example where this new feature is used?

@jediwhale you've been doing quite a bit with the parser recently. What are your thoughts?

tcnh:
My specific use case is a table with its own prefix (like a literal table) that handles urls, wikiwords, e-mailadresses as literals, whilst still parsing symbols like !today and (and -this is why SymbolProvider was altered) any custom symbols defined in plug-ins such as !randomInt (in hsac-fitnesse-plugin), !Fake (in toolchain-plugin), etc.

To achieve that, we need a SymbolProvider like the wikiParsingProvider, but with some symboltypes removed. also, this provider should be created **after** any custom symbols are registered (as these are registered to the wikiparsingprovider from the PluginFeatureFactory. the registration to wikiParsingProvider is hardcoded in the contextConfigurator)

As an example, my plugin's NoLinkParsingTable (extends Table) now defines its own custom provider and prefix:
```
static final SymbolProvider noLinksTableParsingProvider = SymbolProvider.copy(wikiParsingProvider)
            .remove(Link.symbolType)
            .remove(WikiWord.symbolType)
            .remove(SymbolType.EMail)
            .add(SymbolType.EndCell);

public NoLinksTable() {
        super("NoLinksTable");
        wikiMatcher(new Matcher().startLine().string("-^|"));
        wikiMatcher(new Matcher().startLine().string("^|"));
        wikiRule(this);
        htmlTranslation(this);
    }
```
And then applies that from the (now protected) parseCell method:
```
 @Override
    protected Symbol parseCell(Parser parser, String content) {
        Symbol cell = parser.parseToWithSymbols(cellTerminators, noLinksTableParsingProvider, ParseSpecification.tablePriority);
        cell.setType(tableCell);
        return cell;
    }
```
Whilst preventing the need to duplicate existing parse() and/or toTarget methods for tables.

fhoeben:
Would it not be easier (in the code handling the table's content) to unescape any urls, wikiwords and e-mailadresses back to their literal form? Or maybe disable processing of these elements totally, for all tables? (I have to admit I've struggled with this in the past myself and even considered going for an option to disable wikiwords completely.) 

It seems a bit strange to have inconsistent handling of such symbols in a single wiki

tcnh:
That would very much not be easier. Imagine the following example with browser test:

```
| script | browser test             |
| open   | http://www.google.com    |
| enter  | SearchTerm | as | name=q |
```
This will enter `SearchTerm<a title="create page" href="MyFirstSuite.SearchTerm?edit&amp;nonExistent=true">[?]</a>` in the search field. From a user perspective, this is completely irrational behaviour in a script table.

The solution would be to use `!-SearchTerm-!` or a literal tabe (`!| script | ....`). Often, the latter is used because escaping using !- -! cripples readability in edit mode. This however results in !today (etc) symbols no longer being parsed. Which is unwanted in (in my experience) >99% of cases.

Handling any input and stripping unwanted markup in any fixture is cumbersome and error-prone, not to mention inflexible in case different projects have different ways of working.

This PR merely provides the flexibility to adapt this kind of behaviour to your specific context, instead of having to constantly apply workarounds because somehow somebody deemed WikiWords or convenient link markup on email adresses are similar to symbols that actually have value when executing tests.

fhoeben:
I see your point (and in your example recognise the desire not to have WikiWords for which I created an issue). 

But I wonder how the new flexibility addresses the underlying issue. If this PR is merged, what does a test with a 'NoLinksTable' look like, can it be used for anywhere a script/scenario was used or ...?

tcnh:
If this is merged and you are using a plugin that implements its own table type, the table can be any table. It can be used in scripts/scenarios/decisiontables etc, just like you can make table a literal table using !|.

Example: 
image

Normal, literal and custom tables can also be mixed/nested:
image

jediwhale:
I'll take a look tomorrow.  Reformatting to standard layout is good but it would be nice to have two commits - one with the code changes and one with the reformatting. 

BTW what's the best practice with import *'s? (I'm not a real Java guy so I'm not sure)

fhoeben:
@jediwhale `import *` is considered a bad practice in my book, as:
 - your not being very clear on what you use
 - if later extra classes are added to the imported package the semantics of the class code could change (as suddenly another class is imported from this package, that previously did not exist)

See also https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad#answer-147461

fhoeben:
@tcnh thanks for explaining. I never really used literal tables and hadn't grasped the concept of different table types.

I actually would like to have such a NoLinksTable in FitNesse itself. Any thoughts there @jediwhale?

tcnh:
@jediwhale I agree I should have split the commits in change and formatting change.
The easiest way to see the actual change is to compare whilst ignoring the whitespace changes. Github has a nifty trick to do so by appending w=1 as a query param to the url. 
See https://github.com/unclebob/fitnesse/pull/1307/files?w=1 for the actual changes to SymbolProvider without the whitespace diff

jediwhale:
Thanks, the w=1 is a great tip! I didn't know that.

The code looks good. We should have a new test in SymbolProviderTest - copy a provider, remove, assert doesn't match the removed symbol.

tcnh:
Added two tests: one for copy and one for remove

Issue #1308, opened 2020-11-16
Disable automatic link generation from WikiWords


fhoeben:
I often get frustrated in (Slim scripts/scenarios) tests where I need to enter a text that matches the definition of a FitNesse Wiki-word (e.g. `SomeUser`). I always have to remember to escape such texts (e.g. `!-SomeUser-!`), to prevent FitNesse from making it a link to a non existing page.

For me the ability to create links using WikiWords (instead of having to make an explicit link) is far less useful than just being able to use such texts without additional escaping in a script table.

I would like the ability to turn off WikiWords links for the entire wiki, or maybe on a suite hierarchy level. 

I understand this might be different for other users, and can be a backwards compatibility issue, but it would make my life easier. So the change should be configurable.

I believe I'm not alone in not (always) wanting WikiWords to be turned into link, see also https://github.com/unclebob/fitnesse/pull/1307#issuecomment-727958651

tcnh:
Some thoughts about this:

Addinga table prefix like in my example in the reference PR  to the parser's table class is pretty simple. 
However, because any symbolTypes defined in plugins are -by default- only registered to the wikiParsingProvider, these would have to be explicitly registered to a new 'no links' Provider as well, or we'd have to let the ContextConfigurator explicitly call loadSymbolTypes for both providers.

What would be the preferred approach? Or is there another option I've overlooked?

jediwhale:
Some initial thoughts:
In many cases, the automatic generation of links from wikiwords is not useful. It might be helpful to be able to turn this off and use some kind of simple markup to generate links only when wanted, e.g. This is NotAlink but this @IsAlink.
Should this be configurable for the entire wiki or by individual pages?
The NoLink table allows very fine control, i.e., at the table level, but only applies to one type of markup.

jediwhale:
The alias widget already provides a way to specify a link: [[description][link]]. We could make the description part optional so you could specify a link as [[link]] if auto-linking was turned off. There could be a command line argument to turn auto linking off: something like: -parser noautolink. (This could also be used to enable other parsers, if I ever finish one :). )  This would let you turn off auto-linking for your entire wiki and specify links where you needed them.  If you wanted auto-links on some pages, you could leave auto-linking on and use the no-link table and literals to turn them off where required.

lennyg1:
What is the problem you're trying to fix here? I haven't noticed any situation where there's a problem aside from within tables for testdata or input.

[...]

PR #1310, opened 2020-11-20
No links table

tcnh:
Support a Table type (prefix ^) that doesn't parse wikiwords, links and emailadresses.

Extended version of #1307, implementing the 'requirement' from #1308

### DIFF
--- a/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteWidgetTests/TestTables/content.txt
+++ b/FitNesseRoot/FitNesse/SuiteAcceptanceTests/SuiteWidgetTests/TestTables/content.txt
@@ -60,3 +60,29 @@
 |Response Examiner.|
 |type|pattern|matches?|
 |contents|!-''italic''-!|true|
+#
+
+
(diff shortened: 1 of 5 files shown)


fhoeben:
Looks pretty good. We should probably add an acceptance test for the noLinks table.

Maybe we can prevent having to call `pluginsLoader.loadSymbolTypes()` twice (for each provider) by making the copy at a later point in time (i.e. after the plugins are loaded)? But on the other hand this approach allows a plugin to choose whether a symbol should be included in both providers, or only in one...

tcnh:
What would a logical place for such an acceptance test be? I see no acceptance test for i.e. a literal table.

We could indeed copy at a later point to prevent the double administration, but I can imagine situations where one would want to exclude a custom symboltype from this table. I'd alway opt for more extensibility if the cost is not too high.

jediwhale:
See FitNesse.SuiteAcceptanceTests.SuiteWidgetTests.TestTables - could add to this or make a new FitNesse.SuiteAcceptanceTests.SuiteWidgetTests.TestTablesNoLink

tcnh:
Thanks! Added an acceptance test to that page

PR #1311, opened 2020-11-26
Fix broken rendering of maps inside lists

tcnh:
Since a (long) while, putting one or more hashmaps inside a list caused the hashmap to be escaped, resulting in hard-to-read test result pages. This PR fixes that, specifically for lists, so that other (unconventional) html doesn't break the page's rendering.

Effect of these changes below in before (left) and after (right) situations:
image


fhoeben (commented):
The current list formatter is very limited and displaying nested tables in one example of that. 

BUT the list formatter in Slim is configurable, so one can configure lists to be displayed differently. 
As an example of another formatter please look at https://github.com/fhoeben/hsac-fitnesse-fixtures/blob/master/src/main/java/nl/hsac/fitnesse/slim/converter/NumberedListConverter.java (which my fixtures allow to be activate via a call to [ListFixture's `display lists numbered`](https://github.com/fhoeben/hsac-fitnesse-fixtures/blob/42173e7afda99a4e1448069a4b588c64a936c311/src/main/java/nl/hsac/fitnesse/fixture/slim/ListFixture.java#L22)
I enable this converter when I need to work with complex structures where maps and lists are nested. There's also a method to switch back to the default formatter, because I recall some Tables don't respond well to different list formatting and when checking simple lists the square braces format is quite nice.
 
The regex approach to finding the lists does not feel quite 'right'. But I believe the wiki does not have access to the current list formatter (as that lives in the test runner's process).

Does this change break the usage of custom list formatters?
Could you add some tests?

tcnh:
The list formatter is not the issue, that works fine. This change should not impact anything but the rendering of lists in slim tables. The mapconverter change is needed to make sure that maps containing lists that contain maps - that sounds pretty specific :-) -  are not escaped by the map converter.

What breaks the formatting is the fact that slim's HtmlTable has a very strict rule to determine if a cell qualifies as HTML. It has to be 'HTML-only' (or a symbolassignment containing just html). This works for all situations I know, except lists containing HTML (practically: lists containing hashes). 

I experimented with setting the 'qualifiesAsHtml-rule' less strict (as in .*HTMLPATTERN.* instead of ^HTMLPATTERN$), but that breaks wiki formatting in some cases where html-output is rendered in cells.

One could even argue that the not-escaping of HTML in all cases other than explicit conversions such as lists or maps is unwanted, as it mostly results in broken result pages.

I'll add some tests to show that only lists are affected

PR #1316, opened 2021-02-23
Create distinct rerun failed pages, to allow concurrent suites to run

raducu-deaconu-visma:
In a scenario where we have multiple suites running at the same time towards the same fitnesse instance, at the end of the run we get only one page for rerunning failed tests, as each run overwrites them. 
Adding the page name to "RerunLastFailures" prefix allows identifying and rerunning the failed tests for each individual suite we run.

### DIFF
--- a/src/fitnesse/responders/run/SuiteResponder.java
+++ b/src/fitnesse/responders/run/SuiteResponder.java
@@ -292,7 +292,10 @@ public class SuiteResponder extends ChunkingResponder implements SecureResponder
   }
 
   protected String getRerunPageName() {
-    return "RerunLastFailures";
+    PageCrawler pageCrawler = page.getPageCrawler();
+    WikiPagePath fullPath = pageCrawler.getFullPath();
+    String fullPathName = PathParser.render(fullPath);
+    return "RerunLastFailures_"+fullPathName.replace(".","-");
   }
 
   protected String getTitle() {


fhoeben:
Sorry it took me a while to respond.
You are completely right the current approach does not support rerun properly when you run multiple suites from the UI. (I addressed this problems in my own project when running test via jUnit, by allowing the re-run page name to be explicitly configured for each run, which is the only time I run suites concurrently).

I see some problems with your approach unfortunately:
 - The 'Rerun' button on the top of the page is not updated to match the new pagename, see `fitnesse/resources/templates/menu.vm`
 -  Less importantly: when the suite is filtered (using in- or exclude of tags) ideally that would also lead to different re-run page names

If the pages generated would share a single parent/suite then the user could easily rerun all failures of all concurrent suites in a single rerun after all suites have completed (which is sometimes desirable to rerun without side-effects of other runs running at the same time), while still allowing rerun of the failures of a single run.

raducu-deaconu-visma:
Thank you for the feedback!
Are you referring to the "Rerun Failed" button in `errorNavigator.vm`  ? The reference to that should be set correctly in SuiteResponder , as for the name , i was thinking to rename the button to "Rerun Failed in <suitename>", but the suitename can be quite long. Should it be renamed to something like "Rerun Failed in suite" ? I thought keeping the previous name would be consistent with the previous behavior,  is this what you are referring to?
I do not know about handling include/exclude i will try to look into that.
I was thinking also about placing subpages under a parent "Rerun Failed" , but couldn't get my head around some of the issues in principal, like when to create/delete this parent page, adding a button for the page....and i don't know if i could make a proper solution for that

fhoeben:
You are right I pointed to the wrong .vm. And I overlooked that link is dynamic. I think keeping the displayed name is fine.

Keeping track of the parent, and especially when it may be cleaned, does indeed look challenging. I have to admit this (and the in- and excludes) kept me from doing anything more complicated than just tracking the last test run. ;-)
In a CI run life is a bit easier...

Think about it some more. It would be great if we could support all these concerns.

PR #1318, opened 2021-04-08
Feature: Show number of test pages inside a suite.

suratdas:
When the tests grow, it becomes a little difficult to know the number of test pages inside a suite just by looking at the suite level. This pull request will make it possible to see the test page count for a test suite if enabled. This is enabled by "-c" option (c for count) like below
!contents -R1 -g -p -f -h -c
For details, please see the attached video.

https://user-images.githubusercontent.com/9042580/113959782-533a6c80-97d8-11eb-8975-58e75f1d7475.mov

### DIFF
--- a/src/fitnesse/wikitext/shared/ContentsItemBuilder.java
+++ b/src/fitnesse/wikitext/shared/ContentsItemBuilder.java
@@ -77,16 +77,21 @@ public class ContentsItemBuilder {
         }
         return listItem;
     }
+    private boolean isSpecialPageToBeCountedAsTest(SourcePage page){
+      String pageName = page.getName();
+      return pageName.contains("SuiteSetUp") || pageName.contains("SuiteTearDown");
+    }
 
-    private int getTotalTestPagesInASuite(SourcePage page, int counter) {
-      if (page.hasProperty(PageType.TEST.toString())) {
-        return ++counter;
+    private int getTotalTestPagesInASuite(SourcePage page) {
+      if (page.hasProperty(PageType.TEST.toString()) || isSpecialPageToBeCountedAsTest(page)){
+        return 1;
       }
+      int counter = 0;
       if (page.hasProperty(PageType.SUITE.toString())) {
         Iterator<SourcePage> pages = page.getChildren().iterator();
         while (pages.hasNext()) {
           SourcePage sourcePage = pages.next();
-          counter = getTotalTestPagesInASuite(sourcePage, counter);
+          counter += getTotalTestPagesInASuite(sourcePage);
         }
       }
       return counter;
@@ -97,7 +102,7 @@ public class ContentsItemBuilder {
         //Will show count of test pages under this suite
         if (hasOption("-c", Names.TEST_PAGE_COUNT_TOC)) {
           if (page.hasProperty(PageType.SUITE.toString()))
-            itemText += "(Test pages:" + getTotalTestPagesInASuite(page, 0) + ")";
+            itemText += " ( " + getTotalTestPagesInASuite(page) + " )";
         }
 
         if (hasOption("-g", Names.REGRACE_TOC)) {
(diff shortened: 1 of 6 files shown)


src/fitnesse/wikitext/shared/ContentsItemBuilder.java
```diff
@@ -77,8 +78,27 @@ else if (hasOption("-H", Names.HELP_INSTEAD_OF_TITLE_TOC)) {
...
+      }
+      return counter;
+    }
+
     private String buildBody(SourcePage page) {
         String itemText = page.getName();
+        //Will show count of test pages under this suite
+        if (hasOption("-c", Names.TEST_PAGE_COUNT_TOC)) {
+          if (page.hasProperty(PageType.SUITE.toString()))
+            itemText += "(Test pages:" + getTotalTestPagesInASuite(page, 0) + ")";
```
fhoeben, line 100:
```suggestion
            itemText += " (Test pages: " + getTotalTestPagesInASuite(page, 0) + ")";
```

A few spaces will make this easier to read.
I actually wonder: is the `Test pages:` needed, why not only have the count between parentheses?

src/fitnesse/wikitext/shared/ContentsItemBuilder.java
```diff
@@ -77,8 +78,27 @@ else if (hasOption("-H", Names.HELP_INSTEAD_OF_TITLE_TOC)) {
         return listItem;
     }
 
+    private int getTotalTestPagesInASuite(SourcePage page, int counter) {
```
fhoeben, line 81: To me it seems easier just return 1 for test pages and sum inside the loop for a suite, then no `counter` parameter is needed.
Something like:
```
        int counter = 0;
        while (pages.hasNext()) {
           SourcePage sourcePage = pages.next();
           counter = counter + getTotalTestPagesInASuite(sourcePage);
         }
         return counter;
```

test/fitnesse/wikitext/parser/ContentsItemTest.java
```diff
@@ -47,6 +47,12 @@ public void buildsRegraced() throws Exception {
         assertBuildsOption("PlainItem", new String[]{}, "-g", "REGRACE_TOC", "<a href=\"PlainItem\" class=\"static\">Plain Item</a>");
     }
 
+  @Test
+  public void buildsTestPageCount() throws Exception {
+    assertBuildsOption("PlainItem", new String[]{"Suite=true"}, "-c", "TEST_PAGE_COUNT_TOC",
+            "<a href=\"PlainItem\" class=\"suite\">PlainItem(Test pages:0)</a>");
```
fhoeben, line 53:
```suggestion
            "<a href=\"PlainItem\" class=\"suite\">PlainItem (Test pages: 0)</a>");
```

Please also add tests for a test page and a filled suite (including non-test pages like SetUp, ScenarioLibrary, and TearDown which should not be counted).

Thinking of what should be counted: how are SuiteSetUp and SuiteTearDown pages handled? Can you ensure that they are counted the same as in an actual test run (i.e. count them if and only if they are counted as tests when a suite is run)

fhoeben (changes requested):
Looks pretty good. Besides the comments per file: please also add the new option to the user guide (so people know that can use this flag)

src/fitnesse/wikitext/shared/ContentsItemBuilder.java
suratdas: Thanks. I changed it.

src/fitnesse/wikitext/shared/ContentsItemBuilder.java
suratdas: Thanks for the comment. I changed it.

test/fitnesse/wikitext/parser/ContentsItemTest.java
suratdas: Thanks. SuiteSetUp and SuiteTearDown are counted as test when suite is run. So, I added logic to show that as count. Also added the requested tests.

fhoeben (approved):

PR #1322, opened 2021-05-01
Add support to identify output chunk by test pages

daanipuui:
Hello, 

On the project I work, we use FitNesse for end to end validation. We have recently switched from running the tests from a JUnit test on Jenkins to launching the tests on the test machine using the suite responder. The problem is that the suite responder does not create reports for each test page nor does it aggregates them nicely in a summary report file (as JUnit does). So we have created a custom responder to create a report file for each test page and aggregate them in a summary report file.

This has been much difficult than we have previously thought because currently there is no easy way of linking the output chunk to a test page. This pull request tries to fix this problem by adding the test page to the testOutputChunk method.

### DIFF
--- a/src/fitnesse/junit/JUnitRunNotifierResultsListener.java
+++ b/src/fitnesse/junit/JUnitRunNotifierResultsListener.java
@@ -68,7 +68,7 @@ public class JUnitRunNotifierResultsListener
   }
 
   @Override
-  public void testOutputChunk(String output) {
+  public void testOutputChunk(TestPage testPage, String output) {
   }
 
   @Override
--- a/src/fitnesse/junit/JavaFormatter.java
+++ b/src/fitnesse/junit/JavaFormatter.java
@@ -163,7 +163,7 @@ public class JavaFormatter extends BaseFormatter implements Closeable {
   }
 
   @Override
-  public void testOutputChunk(String output) {
+  public void testOutputChunk(TestPage testPage, String output) {
     try {
       resultsRepository.write(output);
     } catch (IOException e) {
--- a/src/fitnesse/junit/PrintTestListener.java
+++ b/src/fitnesse/junit/PrintTestListener.java
@@ -38,7 +38,7 @@ public class PrintTestListener implements TestSystemListener, Closeable {
   }
 
   @Override
-  public void testOutputChunk(String output) {
+  public void testOutputChunk(TestPage testPage, String output) {
   }
 
   @Override
--- a/src/fitnesse/reporting/BaseFormatter.java
+++ b/src/fitnesse/reporting/BaseFormatter.java
@@ -28,7 +28,7 @@ public abstract class BaseFormatter implements Formatter {
   }
 
   @Override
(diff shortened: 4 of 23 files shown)


fhoeben:
@daanipuui Interesting. Can you share some insight into why running on a test machine with the suite responder was desirable for you. What advantages did this bring?

daanipuui:
@fhoeben Yes, I will try to paint a more detailed picture. 

We are working on a trading platform for banks and financial institutions. FitNesse is used to validate some smaller services that offer clients connectivity to third party platforms like Bloomberg, Reuters, etc. Because the application is huge we are testing these services on test machines where the platform is already installed and we deploy the service on top of that.

Like I already said, until recently, the FitNesse tests were launched from Jenkins as an integration test and the fixtures were doing checks on the test machines. So there was already a hybrid use of FitNesse framework (as a side note, I joined the project long after this was implemented and I cannot comment on why it was implemented this way).

The middle management decided that they want something which is closer to how a client would use these services, hence we are now starting a FitNesse server and launch the test suites through the suite responder.

The first downside is that this change has robbed us from having a test summary with links to smaller test reports: we had a very long report containing all the test pages in the suite.
A second downside is that it seems the tests are running a bit slower now, but we are investigating if we can run tests in parallel (so far is seems possible).

For more info on what we do, you can also watch this video:  https://www.youtube.com/watch?v=XIAMUzJq-gc.
If you more questions or suggestions, please don't hesitate.
Thanks

fhoeben:
I believe I get the picture you are sketching.
What might be of interest to you is a small [class I made some time ago to run FitNesse tests using the jUnit runner from the command line](https://github.com/fhoeben/hsac-fitnesse-fixtures/blob/master/src/main/java/nl/hsac/fitnesse/junit/JUnitConsoleRunner.java). (I use this in my [docker images](https://github.com/fhoeben/hsac-fitnesse-docker) to start tests.) 
If your test cases permit running in parallel you can easily do it using this class by starting multiple processes. 

There is a feature in FitNesse to partition a suite so you can spread it across multiple runners. In my [plugin project](https://github.com/fhoeben/hsac-fitnesse-plugin) I created an extension of this feature to use the previous run times to get a balanced partitioning and I created a [class to create a single overview report of multiple test runs](https://github.com/fhoeben/hsac-fitnesse-fixtures/blob/master/src/main/java/nl/hsac/fitnesse/junit/reportmerge/HtmlReportIndexGenerator.java) (i.e. an overview of all partitions that were done in parallel). An example of a test set parallelised this way can be found in [my GitLab pipeline](https://gitlab.com/fhoeben/hsac-fitnesse-fixtures/-/pipelines), and an [example report ](https://fhoeben.gitlab.io/-/hsac-fitnesse-fixtures/-/jobs/1209328065/artifacts/full-results/index.html) also.

I don't believe you need to use my fixtures to use these features and it sounds like exactly the kind of thing you are looking for...

daanipuui:
Thanks @fhoeben  for merging the PR, unfortunately I did got the chance to check your suggestion.

PR #1327, opened 2021-06-21
Report tests with ExecutionResult IGNORED as Ignored to Junit

tcnh:
Tests that FitNesse reports as ignored from a suite run (blue bar in html result, i.e. when a stop test exception was thrown in a SuiteSetUp) are reported as passed tests to JUnit.
This PR calls testIgnored on the notifier in this situation. The JunitXMLresultrecorder is the called to rcord a testresult for a skipped test.

This ensures pass/fail/skip results in JUnit are in sync with what we see when running from the wiki.

### DIFF
--- a/src/fitnesse/junit/JUnitRunNotifierResultsListener.java
+++ b/src/fitnesse/junit/JUnitRunNotifierResultsListener.java
@@ -60,13 +60,14 @@ public class JUnitRunNotifierResultsListener
     increaseCompletedTests();
     Description description = descriptionFor(test);
 
+
     if (firstFailure != null) {
       notifier.fireTestFailure(new Failure(description, firstFailure));
     } else if (testSummary.getExceptions() > 0) {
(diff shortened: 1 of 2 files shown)


tcnh:
I don't agree, for the following two reasons:
1. In the use cases we see, test suites are run (as a whole) and some test cases can be skipped/ignored due to a failing assumption in the SetUp. These tests did not error, they were deliberately skipped for the context they were run in. This is comparable to using Junit's Assume package in a before method. (which will report a test as ignored at run-time and not errored).
2. It's very inconsistent that the wiki reports something else then the junit report.

I too believe the handling of the skip property is not as it should be. It's actually pretty much useless as it is. Fixing that would be nice, but has a lot more impact I'm afraid, as we should then send all tests to the runner and decide if they should be run there (including setup, etc).

fhoeben:
@tcnh good arguments.

Did you test for tests without assertions, are they reported as ignored also?
> What happens for tests without assertions? I believe these are shown as ignored in the wiki also.

PR #1328, opened 2021-06-23
a getter should not modify the data. This screws up a unittest for me

maartenjanvangool:


fhoeben:
The getter doesn't modify any data, does it? It returns a different representation than you want, but it doesn't change any state.

I haven't looked in detail whether your changes has any impact, or why the extra text is added. But can you elaborate why this getter's behaviour is a problem for you? If you want to the raw content you created a dummy page with you could just do `getData().getContent()`, right?

maartenjanvangool:
You're right it does not change the state, I worded that incorrectly. But it alters the returned object before returning. Seemingly without any reason and without any warning. 

I did initially change it to getData().getContent(). Fixed my unittest, broke my code. I haven't checked why; but getData().getContent() does not return the same thing as getHtml() for every type of WikiPage.

fhoeben:
> Fixed my unittest, broke my code. I haven't checked why; but getData().getContent() does not return the same thing as getHtml() for every type of WikiPage.

That was my point, why does your unit test require the dummy page to return exactly the same thing as the content put in? Shouldn't the test verify that getHtml() was used, instead of getData().getContent(). I believe the implementation of the method in this dummy page is different exactly for that reason...

jediwhale:
getHtml is used to translate the markup language from getContent() into html. So it's perhaps poorly named because it looks like a getter.

PR #1333, opened 2021-08-25
hide slim symbol values prefixed with SECRET_ on the slim runner side

tcnh:
This PR replaces slim symbol for display in test results if they are prefixed with `$SECRET_` with `$_SECRET_symbolName->[*****]`

This is useful to prevent unintentional display of secrets (such as passwords, api keys, etc) when using Slim. Ofcourse it's still easily possible to display these values by echo-ing them through a fixture method, but it helps keeping unwanted info out of test results and dashboards.


fhoeben:
Can you add some tests?
We should check that: both secret and normal symbol values are still passed to fixtures, normal symbols are still shown, secret symbols are not shown

PR #1336, opened 2021-10-05
Update of JUnitRunNotifierResultsListenerTest

tcnh:
Fix failing tests;
Added test for ignored situation
Replaced deprecated AssertThat method with hamcrest.MatcherAssert,AssertThat


tcnh:
It was somehow not possible to amend to the earlier PR (maybe because it was already merged?)

PR #1352, opened 2022-02-18
Develop workflow

EdraPrasetio:


fhoeben:
I haven't used GitHub workflows before. So please help me understand.
This looks like a setup for build/test, on the 'master' and a similar on a 'develop' branch. I don't believe we have a develop branch yet.
Why two different setup files? Will this be sufficient to trigger a build on each commit and PR? Or do we have to setup something more in the repo?

When I look at your repo's actions I get the impression the acceptance tests are not enabled yet. Is that correct/intended?

Is there a way to add steps to also generate the 'simple' and 'standalone' jars, so we validate that generating them is also correct?

fhoeben:
Maybe people can also directly download the latest build result, and you can update the README.md so it contains a link to those (addressing #1354)?

jediwhale:
Please provide a description of the purpose of this pull request. If we don't know the purpose, we can't review it.

jediwhale:
A pull request should have a single, clearly-described objective.  The changes should be complete when it is opened, not ongoing. The requester should respond to reviewer questions and comments. This PR is none of these so I am closing it.

Issue #1330, opened 2021-07-30
Missing close anchor on FitNesse.UserGuide.AdministeringFitNesse.ApplicationArchitecture


mbucc:
**Describe the bug**
Starting from "The Test system subsystem is responsible", the rest of the page text behaves as a single href.

**To Reproduce**
1. Go to 'ApplicationArchitecture' page
2. Scroll down to 'Test execution engine'
3. See error

**Expected behavior**
Half of the page text should not be a single link.

**Screenshots**

<img width="854" alt="image" src="https://user-images.githubusercontent.com/85849/127637152-68e8f441-09bd-4667-9633-44cae04fc15e.png">
**Configuration:**
 - OS: osx
 - Browser safari
 - Version v20210606
 - Test System: n/a
 - Plugins in use: the default configuration

djbrown:
I noticed the same bug on http://fitnesse.org/PlugIns
There might be a general problem with `!anchor` in FitNesse ([docs](http://fitnesse.org/FitNesse.UserGuide.FitNesseWiki.MarkupLanguageReference.LinksWithinPages)).

I dug into this issue and looked at the **response** HTML (from the server):
```html
<h2 id="1">Test execution engine <a id="testsystem"/></h2>
The Test system subsystem [...]
<h2 id="2">System under test <a id="systemundertest"/></h2>
The system under test [...]
```

The **rendered** HTML (developer tools) looks like this:
```html
<h2 id="1">Test execution engine <a id="testsystem"></a></h2>
<a id="testsystem">The Test system subsystem [...]</a>
<h2 id="2"><a id="testsystem">System under test </a><a id="systemundertest"></a></h2>
<a id="systemundertest">The system under test [...]</a>
```

I guess modern browsers render the self closing anchor-Tag `<a />` differently from back when this feature was implemented. I'll try to find the code responsible for translating `!anchor xyz` to `<a id="xyz" />` . It should instead translate the template to `<a id="xyz`></a>`.

PR #1366, opened 2022-04-22
make anchor tags non self-closing

djbrown:
fixes unclebob/fitnesse#1330
see also https://www.w3.org/TR/xhtml-media-types/#C_3
and https://developer.mozilla.org/en-US/docs/Glossary/Empty_element

### DIFF
--- a/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
+++ b/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
@@ -1,5 +1,6 @@
 !2 Latest Release ${FITNESSE_VERSION}
 !2 Pending Changes
+ * Bugfix: render [[''!-!anchor-!''][FitNesse.UserGuide.FitNesseWiki.MarkupLanguageReference.LinksWithinPages]] as valid (non empty) element ([[1366][https://github.com/unclebob/fitnesse/pull/1366]])
  * Bugfix: Handle instructions with embedded newlines ([[1363][https://github.com/unclebob/fitnesse/pull/1363]])
  * Add new responder to list the variables available on a page ([[1355][https://github.com/unclebob/fitnesse/pull/1355]])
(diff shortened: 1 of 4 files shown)


djbrown:
@fhoeben I didn't interpret the article like that. It just says we cannot put child nodes (text or elements) inside empty elements (as listed there).
the article is probably not the best reference for thhis issue, since it for the `<a>` element, since it doesn't concern non-empty elements (which the `<a>` element is).

I couln't find a better referenece, the [`<a>` docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) don't say anything about wheter there has to be a child node (text or elements), but I think it's fine to have no children here.

I used the w3 HTML validator to check, and it seems we don't have to put text in `<a>` elements:
https://validator.w3.org/nu/#textarea
```html
<!DOCTYPE html>
<html lang="">
<head>
    <title>Test</title>
</head>
<body>
    <a id="normal">Text</a>
    <a id="no-content"></a>
    <a id="self-closing" />
    <b>some text</b>
</body>
</html>
```

it only complains about the self-closing node, as in https://validator.w3.org/nu/?doc=http%3A%2F%2Ffitnesse.org%2FPlugIns

----

but to develop the discussion even further: actually I think that an `<a>` might not be the best HTML element for representing `!anchor` since it doesnt reference any other element (no `href`). it's not a link to somewhere, but instead a reference for others to link to. it could or should be an empty `<span id="xyz" />`.

fhoeben:
I believe an `a` without a `href` attribute is perfectly valid and was originally the only way to create a link target. Now it indeed adds little value over an (empty) `span` with an `id`.

It all looks fine to be merged. Would you mind adding a small item for this fix to the ReleaseNotes page (that way I don't have to remember to do that :-) )?

djbrown:
@fhoeben done ;-)

Issue #1135, opened 2018-05-04
inline images rendered incorrectly


Kizaemon:
Version: release 20180127 fitnesse-standalone.jar

When copy paste an image from clipboard into Rich Text editor, the image is being copied and rendered correctly in the Rich Text editor.

``
!img data:image/png;base64,iVBORw0KGgoAAAA[...snip...]
``

But the page output is not rendered correctly. The resulting html is

``
<img src="data"/>:image/png;base64,iVBORw0KGgoAAAA[...snip...]
``

The html output should have been

``
<img src="data:image/png;base64,iVBORw0KGgoAAAA[...snip...]"/>
``

Many Thanks!

djbrown:
I can confirm this bug is still reproducable in the master currently

djbrown:
The Documentation only mentions URL based images, no inline/base64-encoded ones: [FitNesse.UserGuide.FitNesseWiki.MarkupLanguageReference.MarkupPicture](http://fitnesse.org/FitNesse.UserGuide.FitNesseWiki.MarkupLanguageReference.MarkupPicture)
I guess the rich text editor somehow automagically handles it correctly (some library) but it isn't a proper feature of fitnesse with tests etc.

I will look into this and try to find a fix to make inline pasted images work.
#featureNotaBug 😁

PR #1367, opened 2022-04-27
allow base64 data url images

djbrown:
fixes unclebob/fitnesse#1135
see https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs
and https://en.wikipedia.org/wiki/Base64

🚧 open tasks:
- [x] test and fix other tokens and symbols which would be valid base64 data
  - [x] handle `+n`: `!img data:image/png;base64,some+5data` falsely renders as `+5data`
  - [x] what about variables in the data part, should they be supported? -> negligible
  - [x] ~or just parse to the next whitespace and ignore possible tokens?~
- [x] add documentation under [FitNesse.UserGuide.FitNesseWiki.MarkupLanguageReference.MarkupPicture](http://fitnesse.org/FitNesse.UserGuide.FitNesseWiki.MarkupLanguageReference.MarkupPicture)
- [x] add acceptance tests under [FitNesse.SuiteAcceptanceTests.SuiteWidgetTests.TestImage](http://fitnesse.org/FitNesse.SuiteAcceptanceTests.SuiteWidgetTests.TestImage) -> negligible
- [x] performance testing -> negligible
- [x] releasenotes entry


fhoeben:
I don't think parsing variables inside the data URL is a priority: we can add that if someone really needs it

I believe the unit tests will suffice, a FitNesse test doesn't seem to add much value, but the documentation and release notes  are of course required.

src/fitnesse/wikitext/parser/Image.java
```diff
@@ -55,14 +56,51 @@ public Maybe<Symbol> parse(Symbol current, Parser parser) {
...
             return makeImageLink(current, link, imageProperty);
         }
         else return Symbol.nothing;
     }
 
+    /**
+     * https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs
+     */
+    private String parseImageUrl(Parser parser) {
+        StringBuilder imageUrl = new StringBuilder();
```
fhoeben, line 72: Maybe move this to just after the check whether the contents starts with `data`, so we don't even allocate it if it is not needed.

src/fitnesse/wikitext/parser/Image.java
```diff
@@ -55,14 +56,51 @@ public Maybe<Symbol> parse(Symbol current, Parser parser) {
...
             addOptions(link, options);
             return makeImageLink(current, link, imageProperty);
         }
         else return Symbol.nothing;
     }
 
+    /**
+     * https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs
+     */
+    private String parseImageUrl(Parser parser) {
```
fhoeben, line 71: Maybe it would be nicer to rename this to `tryParseDataUri` and have it return `null` if no data URI is found? In that case the caller can just keep its current behaviour. I think that is nicer than having this method deal with a `fallback` (which will actually be used most of the time).

src/fitnesse/wikitext/parser/Image.java
djbrown: done :-)

FitNesseRoot/FitNesse/UserGuide/FitNesseWiki/MarkupLanguageReference/MarkupPicture/content.txt
```diff
@@ -17,3 +17,8 @@ You may also set the width of the image if you wish by using '''!-!img -w nnn-!'
...
...I0Vk5r+tiLY+NHINMMmp03uNWYpUJquXlBCVi8o1WlYL0mL8dnie8ooZCLNHBRpaRqghRC5/pIRHiyCAIsJ/fsx1h/Mztt47eIyKDgfskOSXjPaxfNI80SmCBEe6K7Ka2LDOYocTWV/7N2sPvzJ+Apcdc2jqEz9diA4LiWBFNHJyu+TLOBgvAOEhlmSx2TCbJIpFlOsNfdJLAgMzkhvNJpDXBIWtCjjc8QPQHwmNdxjvD/kg4zxLtHa38hsagOiImFVeBBi3UUYAmETIzFiRkwSaMzuyhGJPmN/mBUZNyxnUOI4iJQw0EhLNBpC+jQaszYIV7E+cRwax6RZuxgZr7HPkmLVGqxJs2/HTAmTUcjMFIGQd4hFe1B4oOvlvjfzblUUcmZVufbkzBcqppQc1MTQW1+UTpkH5R4ouGdAfa3ZEIF5cnCyf2ItWxt+GDpYXMQWDcFOq6edwEQGF0lLN15vHHgxYcLI+LOiCQ/xGWaCOaTNhFTZvO1lLJpaa3vhneMRHCJ9/vN4aTTQYWwbI8OW6f8w2owlRiSReI9ReBCTYywaQsaGxe1uCakZU7HaQ1Z41Mq44v2I5zwai0RWAmT7UaSdn1tSDu06QfSW53Pa2Ozkwu1rvGfMea0SzSAxYiGqTx4YNP7v32CBYMbTO6AMI4EkanUN/xwElFeBptJYFycTYgf/N6Cu6TV8AKKAjcZYpZnosdj+xwpNRjoEQUuINcGWkFATrhcZoZEGS0pMooKV/y0k3JQVZCItTKizVmEiSBDmqd4jI0wRiId2bnkoGP/z2t2lN+t/BydHrfjr9H0wXgchgq4vzEGFaX7PIda0rniA0VMYzjkzjr04UqhjmFgjCFPLo772Z/7/GSaCHCGlsdaCRh2sW4IW0dpM6riIQkskemqNFlmINFvtR8P9cSHotprxXDkuEEEzyO9XTiTzLTBwDB+biQUECVogKEhwtTHmVEo0rhqBwg2jp9Bd9gk8b1MS6kyhjmEYvCkhUQ8wMs36xDFKC6Le6YmF+mg7Y7qHpIeNAmPHBUKvk4BFghPquGIdU5R0JoEc+W1JTX59M+AaA8ythKF1q6FgOpFfoN6SKPekkEIKKaRQixCFR627CYDhY6z4t3oJAoQKxpjjAi5csIQLXdBlhu6nI3BP3SRVSyGFFFJIQRr/D8F5MBAkaqwxAAAAAElFTkSuQmCC|
```
fhoeben, line 24: Maybe you can find a bit smaller image? We just want to show it in action, right?

djbrown: I thought about using a minimal image but even an 8 by 8 pixel image would lead to unreadable markup.
So I decided to just use the existing example and added a note addressing the generally bad readability.

Should I use a very minimal image instead (like https://gist.github.com/ondrek/7413434) ?
The rendered image wouldn't be very visible though.

PR #1370, opened 2022-04-30
fix format of preformat examples

djbrown:


djbrown:
Upstream: http://fitnesse.org/FitNesse.UserGuide.FitNesseWiki.MarkupLanguageReference.MarkupPreformatted
Before: before
After: after

PR #1382, opened 2022-08-18
Markup system plugins

jediwhale:
Ability to create plugins for new markup syntax.


fhoeben:
Do you also/already have a new markup plugin? I would love to see what you have in mind....

jediwhale:
It's still in a very experimental stage so nothing to demo yet. Inspired by https://docs.racket-lang.org/scribble/index.html but using JavaScript.

jediwhale:
image

Issue #1384, opened 2022-08-21
Unescaping HTML is done in a wrong order


jediwhale:
It replaces `&amp;` with `&` and then replaces `&lt;` with `<`. So given `&amp;lt;`, the result is `<`, when it should be `&lt;`.

The problem is the replacements should be done in the reverse order.

It's been like this since 2008. A concern is users that have workarounds that rely on this incorrect behavior.

There are 3 FitNesse acceptance tests that are written incorrectly but pass because of this behavior.

PR #1385, opened 2022-08-23
fix html unescaping #1384

jediwhale:


fhoeben:
Please also include this change in the ReleaseNotes page

PR #1391, opened 2022-09-21
Publish unit test result

suratdas:
Based on the comment in Gitter, this publishes unit test results.
image

### DIFF
--- a/.github/workflows/build-fitnesse.yml
+++ b/.github/workflows/build-fitnesse.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Build with Gradle
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: build slimJar sourcesJar javadocJar
+          arguments: build slimJar sourcesJar javadocJar test
           # Only write to the cache for builds on the 'master' branch.
           # Builds on other branches will only read existing entries from the cache.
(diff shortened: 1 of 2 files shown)


suratdas:
The published test result can be found in my fork - https://github.com/suratdas/fitnesse/runs/8462287658

fhoeben:
@suratdas did we need the explicit `test` task in the gradle build? As far as I'm aware the tests were already being executed as part of the `build` task.

Furthermore, seeing the statistics like this, I'm quite surprised to see this many ignored tests. They all seem to come from a single class that is executing a suite of FitNesse tests via jUnit. Could you have a look at why these are reported as 'ignored'? I don't like ignored tests and I think it makes the output look bad this way, I would at least like to understand why they are there. I understand it doesn't really have anything to do with your change here but ...

suratdas:
I removed the @Ignore tag in one of the test classes and now two tests are failing, some are still ignored. Latest result is at 
https://github.com/suratdas/fitnesse/actions/runs/3119216239/jobs/5059044742

fhoeben:
@suratdas I don't have the time to look into the ignored test unfortunately. Can you reintroduce the @ignore tag (so the build passes again) and we will see the test results published?
Thanks

suratdas:
@fhoeben Done.

PR #1395, opened 2022-11-02
Add '-w workers' command line option.

abascom:
Add an optional command line parameter to be able to specify the maximum number of thread pool workers given to the thread pool executor.

### DIFF
--- a/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
+++ b/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
@@ -1,5 +1,6 @@
 !2 Latest Release ${FITNESSE_VERSION}
 !2 Pending Changes
+ * Add a new optional command line parameter to specify the maximum number of workers to use for handling incoming requests. 
 
 !2 20221102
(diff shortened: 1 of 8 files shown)


fhoeben:
You indicate the minimum number of workers must be 5, why is that important? Could you add a check to the code to throw an error if a too small value is provided?

Can you also add the change to the ReleaseNotes page?

abascom:
Thanks for the feedback!

**You indicate the minimum number of workers must be 5, why is that important?**
This is due to the ThreadPoolExecutor created in fitnesse/FitNesse.java having a corePoolSize argument of 5.

**Could you add a check to the code to throw an error if a too small value is provided?**
I elected not to do this as it is consistent with the behavior of the port argument.  The port is just passed along to the ServerSocket which will throw an IllegalArgumentException if the port is out of range.  Similarly the ThreadPoolExecutor will throw an IllegalArgumentException already if the maximumPoolSize is less than the CorePoolSize.

**Can you also add the change to the ReleaseNotes page?**
I will do this and update the PR.

Issue #1402, opened 2022-12-08
When we change a method name in the code, it should be possible to replace the new method in all existing test pages.


suratdas:
After a number of tests already written, if we change a method name in the automaton code, it should be possible to easily modify existing tests with the new methods. It should also retain the existing parameters in the tests.

There is a replace feature already available in FitNesse, we can enhance it to replace method like below.

![image](https://user-images.githubusercontent.com/9042580/206356177-fcb3622e-a642-4f13-b4e8-cb0dbf798837.png)

fhoeben:
This would be a very cool feature, but I believe it will be quite hard to implement. The same method may be present in multiple classes, if the test system is 'slim' the same method could be called from scripts and decision tables and scenarios might be defined with the same name as the original method.

The IntelliJ plugin tries to support something like this, but I believe it is horribly broken and I would not recommend its usage on a project of any complexity/size.

I believe using a text editor on the source files of you wiki pages and doing a regular expression based search replace, followed by careful review of all changes made is the best approach (and actually quite a good one) to deal with this situation. This is one of the reasons I like having my tests in version management and being able to run a full test suite based on a branch of my tests before I share it with the rest of the team. 

Getting this to work right via the wiki, and good enough to be safe enough on a shared wiki server seems very hard to me.

Someone who want to attempt to implement this can maybe take some inspiration from the 'slim coverage' test system I developed some time ago (in https://github.com/fhoeben/hsac-fitnesse-plugin, [slim coverage source](https://github.com/fhoeben/hsac-fitnesse-plugin/tree/master/src/main/java/nl/hsac/fitnesse/slimcoverage)). That does a 'dry run' of a test suite to determine which scenarios are used where. I believe that could be extended to find out which methods are called where, which would be valuable input to determine which places require renaming and which do not, even though that still would not tell you on which classes the methods are called.

suratdas:
Agreed. I raised a pull request with those comments. I think it may still be useful compared to just doing a regex replace as I gave a few examples in the screenshot in the pull request.

PR #1403, opened 2022-12-11
Addresses issue #1402 - Possible to search and replace method/scenario names with new name.

suratdas:
If we want to change method or scenario name in tests, we can use this feature. This will also take care of existing parameters without impacting them. Please refer to below is example.

FitnesseReplaceWindow

FitnesseSerchReplace

### DIFF
--- a/src/fitnesse/resources/templates/refactorForm.vm
+++ b/src/fitnesse/resources/templates/refactorForm.vm
@@ -36,7 +36,7 @@
   <a class="button" href="$viewLocation">Cancel</a>
  </fieldset>
 <p>
-  You may change a method/scenario name and want to reflect this name change across all existing test cases. To accomplish this, you can supply a relevant existing line using that method from any test page. e.g. <strong>|Go to|url|</strong> can be changed to <strong>|Navigate to|url|</strong>. Keep in mind following.
+  Reflect an updated method/scenario name across all existing test cases. To accomplish this, you can supply an existing line using that method from any test page. e.g. <strong>|Go to|url|</strong> can be changed to <strong>|Navigate to|url|</strong>. Keep the following in mind:
   <ul>
    <li>This will keep the original parameters (if any) in the wiki which is good in most cases.</li>
    <li>This will replace any line which evaluates to a method name including a line in script table, a scenario or a decision table having similar name. This will be desired in most cases. If you used same method in multiple classes, all of them will be impacted in the wiki pages.</li>
--- a/src/fitnesse/resources/templates/refactorForm.vm
+++ b/src/fitnesse/resources/templates/refactorForm.vm
@@ -38,7 +38,7 @@
 <p>
   Reflect an updated method/scenario name across all existing test cases. To accomplish this, you can supply an existing line using that method from any test page. e.g. <strong>|Go to|url|</strong> can be changed to <strong>|Navigate to|url|</strong>. Keep the following in mind:
   <ul>
-   <li>This will keep the original parameters (if any) in the wiki which is good in most cases.</li>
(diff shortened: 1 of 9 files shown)


fhoeben:
I don't believe it is presented to the user very clearly like this.

I believe it would be better to have two sections on this page: one for (the existing) regular search/replace and other (below it) to do method search and replace.
This will also allow the caveats regarding the method search/replace to be closer to the actual input boxes.

suratdas:
Thanks for the comment. I changed the UI based on your feedback. The new UI looks like below.
image

src/fitnesse/resources/templates/refactorForm.vm
```diff
@@ -1,22 +1,48 @@
...
+ <fieldset>
+  <label for="replacementString">Replacing Method:</label>
+  <input type="text" size="50" id="replacementString" name="replacementString" value="#if ($request.hasInput("replacementString"))$request.getInput("replacementString")#end"/>
+ </fieldset>
+ <fieldset class="buttons">
+  <input type="submit" name="replace" value="Replace Method!"/>
+  <a class="button" href="$viewLocation">Cancel</a>
+ </fieldset>
+<p>
+  You may change a method/scenario name and want to reflect this name change across all existing test cases. To accomplish this, you can supply a relevant existing line using that method from any test page. e.g. <strong>|Go to|url|</strong> can be changed to <strong>|Navigate to|url|</strong>. Keep in mind following.
```
fhoeben, line 39:
```suggestion
  Reflect an updated method/scenario name across all existing test cases. To accomplish this, you can supply an existing line using that method from any test page. e.g. <strong>|Go to|url|</strong> can be changed to <strong>|Navigate to|url|</strong>. Keep the following in mind:
```

src/fitnesse/resources/templates/refactorForm.vm
```diff
@@ -1,22 +1,48 @@
...
+  <ul>
+   <li>This will keep the original parameters (if any) in the wiki which is good in most cases.</li>
```
fhoeben, line 41:
```suggestion
   <li>This will not change the parameters (if any) in the wiki.</li>
```

src/fitnesse/resources/templates/refactorForm.vm
```diff
@@ -1,22 +1,48 @@
...
+   <li>This will replace any line which evaluates to a method name including a line in script table, a scenario or a decision table having similar name. This will be desired in most cases. If you used same method in multiple classes, all of them will be impacted in the wiki pages.</li>
```
fhoeben, line 42:
```suggestion
   <li>This will replace any line which evaluates to a method name including a line in script table, a scenario or a decision table having a similar name. This will be desired in most cases. But if you have multiple classes with the same method of which you only changed one this will not work as desired, since all calls are replaced.</li>
```

src/fitnesse/resources/templates/refactorForm.vm
```diff
@@ -1,22 +1,48 @@
...
+   <li>It is a good idea to review the replaced lines locally (e.g. using Git) and undo any undesired changes before sharing/committing the tests. Most IDE has version control comparison and selective undo feature.</li>
```
fhoeben, line 43:
```suggestion
   <li>It is a good idea to review the updated lines locally (e.g. using Git) and undo any undesired changes before sharing/committing the tests. Most IDE have a version control comparison and selective undo feature.</li>
```

fhoeben (changes requested):
Some minor text changes.
Let's see whether we get user feedback whether this works good enough for the requestor.

fhoeben (approved):

PR #1409, opened 2023-01-01
Feature to search method/scenario in pages.

suratdas:
Over a period of time, there will be a good number of methods/scenario written and used across teams. When it's time to clean-up/refactor those scenario, it's important to know where they are used to assess impact and plan.

image

### DIFF
--- a/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
+++ b/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
@@ -1,5 +1,7 @@
 !2 Pending Changes
  * Add date and page history link to XML responses. ([[1396][https://github.com/unclebob/fitnesse/issues/1396]])
+ * Search and replace method/scenario names with new name. ([[1403][https://github.com/unclebob/fitnesse/pull/1403]])
+ * Search method or scenario in pages. ([[1409][https://github.com/unclebob/fitnesse/pull/1409]])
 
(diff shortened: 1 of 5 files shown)


fhoeben:
To me this search seems a bit too limited (I prefer the full 'coverage' scan provided by the ['slimCoverage' test system in my plugin](https://github.com/fhoeben/hsac-fitnesse-plugin/blob/25aabda3f32e73a6776a996e70af203c3a8aa152/README.txt#L12), but I assume this search could have its value too).

To me it is not immediately clear what the 'Scope' radio buttons do. I assume they allow a subtree to be searched, if one navigates to the search page from a nested page. But in your screenshot I see 'All' and 'FitNesseRoot' which I assume both will mean all pages will be searched. In that situation I don't think we should show the scope radio, since the user's selection makes no difference.
I don't believe there are any tests to ensure the scope selection affects the search results. Can you add some?

suratdas:
Thank you for your comment. I took the screenshot for "FitNesseRoot" but I tested using sub-suites. "FitNesseRoot" and "All"  will be displayed only if we initiate the search from "FitNesseRoot" page. This is the default behavior even for text and properties search. I added tests for scope selection result.

fhoeben:
Can you add the new feature to the ReleaseNotes page also?

suratdas:
Release notes updated. Thanks.

PR #1426, opened 2023-03-21
JDK11+ nashorn scripting, SecurityManager update, Gradle 7.6

tcnh:
For reference. 

This branch runs scripting (slim symbol expressions) with Java 11 and up. It incorporates OpenJDK Nashorn instead of depending on the previously built in nashorn scriptengine.
Also removed the no-longer existed methods from the SecurityManager implementation (but it should be removed entirely imho) and upgraded Gradle and other dependencies.

Note: this branch can no longer be compiled using JDK8.


buildSrc/src/main/groovy/LessCompiler.groovy
```diff
@@ -1,10 +1,13 @@
...
+import org.gradle.api.tasks.JavaExec
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
 
 class LessCompiler extends JavaExec {
   @InputDirectory
   def inputDir
 
-  @Input
+  @OutputFile
```
fhoeben, line 10: This seems like a strange change. Is this file suddenly no longer an input but an output?

buildSrc/src/main/groovy/WikiFileListBuilderTask.groovy
```diff
@@ -12,7 +12,7 @@ public class WikiFileListBuilderTask extends DefaultTask {
   @OutputFile
   File updateDoNotCopyOverListFile
 
-  @Input
+  @OutputFiles
```
fhoeben, line 15: idem

src/fitnesse/slim/SlimExpressionEvaluator.java
```diff
@@ -14,21 +14,13 @@
...
   public SlimExpressionEvaluator() {
-    this("JavaScript");
+    this.engine = new NashornScriptEngineFactory().getScriptEngine();
   }
 
-  public SlimExpressionEvaluator(String engineName) {
-    this(ENGINE_MANAGER.getEngineByName(engineName));
-  }
-
-  public SlimExpressionEvaluator(ScriptEngine engine) {
```
fhoeben, line 29: Why not keep this constructor and feed it the nashhorn engine?

fhoeben (commented):
Looks pretty good.
If you are happy with it I think it might be mergeable once the release notes page is also updated (we'll also need to update the FitNesse.org site as well of course). 
Do you have any open issues that should be addressed?

fhoeben:
@jediwhale what do you think about switching to Java 11 and dropping Java 8 support. Is that an issue for you?

jediwhale:
OK with me

buildSrc/src/main/groovy/LessCompiler.groovy
tcnh: yes, that was a leftover from some debugging. reverted

src/fitnesse/slim/SlimExpressionEvaluator.java
tcnh: I thought to do it this way, so we had the nashorn import and scriptengine initialization only in this file. I can see the compatibility argument to keep the old constructor. Not really a preference for one or the other.

PR #1428, opened 2023-03-30
Fix incorrect character count in SlimSerializer with surrogate pairs.

epistax:
When serializing Slim streams, count surrogate pairs as a single character in character counts.

This prevents a crash in CSlim that occurs when the number of characters specified exceeds the number of characters present.

### DIFF
--- a/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
+++ b/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
@@ -2,6 +2,7 @@
  * Add date and page history link to XML responses. ([[1396][https://github.com/unclebob/fitnesse/issues/1396]])
  * Search and replace method/scenario names with new name. ([[1403][https://github.com/unclebob/fitnesse/pull/1403]])
  * Search method or scenario in pages. ([[1409][https://github.com/unclebob/fitnesse/pull/1409]])
+ * When serializing Slim streams, count surrogate pairs as a single character in character counts.
 
 !2 20221219
  * Add a new optional command line parameter to specify the maximum number of workers to use for handling incoming requests.
--- a/src/fitnesse/slim/protocol/SlimSerializer.java
+++ b/src/fitnesse/slim/protocol/SlimSerializer.java
@@ -34,7 +34,7 @@ public class SlimSerializer {
 
     for (Object o : list) {
       String s = marshalObjectToString(o);
-      appendLength(s.length());
+      appendLength(s.codePointCount(0, s.length()));
       appendString(s);
     }
     result.append(']');
--- a/test/fitnesse/slim/protocol/SlimSerializerTest.java
+++ b/test/fitnesse/slim/protocol/SlimSerializerTest.java
@@ -31,6 +31,12 @@ public class SlimSerializerTest {
   }
 
   @Test
+  public void oneItemListSerializeWithSurrogatePair() throws Exception {
+    list.add("\ud802\udd00 is only one character.");
+    assertEquals("[000001:000024:\ud802\udd00 is only one character.:]", SlimSerializer.serialize(list));
+  }
+
+  @Test
   public void twoItemListSerialize() throws Exception {
     list.add("hello");
     list.add("world");


jediwhale:
I'm wondering how other Slim server implementations (e.g. Java, .NET) handle this situation. Will this be a breaking change for them?

epistax:
I would be happy to continue to maintain this patch if I could know what direction we want to take this. Right now either the code or the documentation really ought to be changed, and Slim implementation library projects would want to know in either case!

Thanks

fhoeben:
I have to admit I'm not familiar enough with unicode an surrogate pairs to see what the impact might be. I consider it quite likely no one ever tried to use surrogate pair with FitNesse before, and no tests break with your change. So in principle I would consider it mergeable. But I would like to see actual acceptance tests use surrogate pairs besides the unit test you now added.

It should actually be fairly straightforward to have a test in the project to check how the Java implementation of Slim will deal with this I suppose. Just have a test that send such pairs from the wiki to a Slim example fixture and see they are received properly in the Java class  implementing the fixture. And have another method where the fixture emits a such a surrogate pair as part of a method result and ensure it ends up in the wiki correctly.

We already have sample fixtures for Java in the project it doesn't sound hard. But (not sounding hard and not taking a lot of time can be two quite different things I learned and) I personally don't have the time to set this up.

@jediwhale I would assume you have some insight into the impact on the .Net Slim server...

jediwhale:
I'll test with .NET

jediwhale:
This is a breaking change for .NET.  The .NET Slim implementation assumes the length prefix for a string is the number returned by the string Length property, which is the number of UTF-16 characters, not the number of Unicode code points. e.g. the following test passes because the second Unicode code point in the string is stored as a surrogate pair of UTF-16 characters.
`Assert.AreEqual(6, "h🀜llo".Length);`
I suspect the same is true in Java but I haven't tested it yet there.

epistax:
> This is a breaking change for .NET. The .NET Slim implementation assumes the length prefix for a string is the number returned by the string Length property, which is the number of UTF-16 characters, not the number of Unicode code points. e.g. the following test passes because the second character in the string is stored as a surrogate pair of UTF-16 characters. `Assert.AreEqual(6, "h🀜llo".Length);` I suspect the same is true in Java but I haven't tested it yet there. If so, the documentation is incorrect and should be changed from "the number of characters in the string" to "the number of UTF-16 characters in the string"

Wouldn't a 4-byte UTF-16 character have two _code units_ but still only be a single UTF-16 character? Per the javadoc: "Returns the length of this string. The length is equal to the number of Unicode code units in the string.". Indeed this was the root of our problem because on the cslim side we were interpreting this number as the number of characters rather than the number of code units.

Meanwhile I believe a character is something that maps to a grapheme. And it a surrogate pair it takes more than a single UTF-16 code unit to get there.

At any rate, it sounds like I should change cslim to expect the number mean the number of _code units_ in the string!

jediwhale:
Yes, you're right - I should have said code units, not characters!

epistax:
I guess the weird thing about this is that we're saying that the UTF-8 slim protocol needs to be aware of the UTF-16 interpretation of the data. So to be complaint we may be need to go from UTF-8 on either/both sides of the communication to UTF-16 using a library like icu or iconv to see what the UTF-16 representation would be, and report on that for the size of the text fields? Err maybe I'm a little turned around here, but that sounds like what we're saying?

I'm now wondering why this isn't just number of bytes.

jediwhale:
I agree it's not the best protocol. It is very Java-centric, based on how Java handles strings. Other platforms have to try and duplicate what Java is doing.  For .NET, it's not too bad, because Microsoft copied what Java does.  For others, not so simple.

jediwhale:
Updated documentation and added tests #1433

fhoeben:
@epistax @jediwhale  I have to admit I haven't been paying enough attention to this PR. Is this still something to be merged in?

epistax:
I raised a point about the implementation not matching the specification and the spec was changed in response. I think we're now stuck on our fork. We'll either need to overhaul cslim or find an alternative entirely.

Thanks, sorry for keeping it open.

PR #1429, opened 2023-03-31
Refactored how a cell displays when using ignore functionality

eireglas:
Now it will show the original value along with 'Test Not Run'.

See below original issue where 'Test not run' overwrite original value:
image

Now with this code change it will show the original value also:
image

### DIFF
--- a/src/fitnesse/testsystems/slim/HtmlTable.java
+++ b/src/fitnesse/testsystems/slim/HtmlTable.java
@@ -450,6 +450,9 @@ public class HtmlTable implements Table {
         case IGNORE:
           // IGNORE + does not count === Test Not Run
           if (testResult.doesCount()) {
+            if(testResult.getMessage() != null){
+              return String.format("%s <span class=\"ignore\">%s</span>", originalContent, message);
+            }
             return String.format("<span class=\"ignore\">%s</span>", message);
           } else {
             return String.format("%s <span class=\"ignore\">%s</span>", originalContent, message);
--- a/src/fitnesse/testsystems/slim/HtmlTable.java
+++ b/src/fitnesse/testsystems/slim/HtmlTable.java
@@ -450,7 +450,7 @@ public class HtmlTable implements Table {
         case IGNORE:
           // IGNORE + does not count === Test Not Run
           if (testResult.doesCount()) {
-            if(testResult.getMessage() != null){
+            if (testResult.getMessage() != null) {
               return String.format("%s <span class=\"ignore\">%s</span>", originalContent, message);
             }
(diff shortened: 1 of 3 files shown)


eireglas:
@fhoeben If you could review, that would be great!

src/fitnesse/testsystems/slim/HtmlTable.java
```diff
@@ -450,6 +450,9 @@ public String formatTestResult() {
         case IGNORE:
           // IGNORE + does not count === Test Not Run
           if (testResult.doesCount()) {
+            if(testResult.getMessage() != null){
```
fhoeben, line 453:
```suggestion
            if (testResult.getMessage() != null) {
```

fhoeben (changes requested):
Looks good to me

nguyentuanviet77:
I didn't agree with this change. Actually, if it was shown the value "Hello" then it was running the Echo function already but the result check was not run and it was inconsistent. I think shouldn't change anything and keep it current for now

eireglas:
I disagree with the last comment. I don't see why we can't have the original value shown when we are ignoring the test.

fhoeben:
@nguyentuanviet77 I don't believe the `Hello` shown in this case is the return value from the echo method, but the expected value as was present in the table before the test was run. 

But this confusion is likely to occur with this change.

@eireglas can you elaborate why it is good/important to show the expected value even when the test is aborted and the lines are not actually run?

Thinking about it now I believe one could even argue that all rows after the exception should not be shown and only a single one (spanning the whole table) instead showing 'Test aborted' (or a similar text).

nguyentuanviet77:
@fhoeben. 

> @nguyentuanviet77 I don't believe the `Hello` shown in this case is the return value from the echo method, but the expected value as was present in the table before the test was run.
> 
> But this confusion is likely to occur with this change.
> 
Yes, agreed, I didn't read the code change yet, but let's imagine that when the step was ignored but still shown the result in the table content was not make sense

@eireglas as you can see the original value was still shown in the table content as the parameters why did you need to show it again? let's imagine that if this was not the **echo** method it was your code method. What should behave with more than one parameter? what was the benefit to show the original param values here?

eireglas:
@fhoeben @nguyentuanviet77  Apologies, I probably explained it incorrectly when I said "Now it will show the original value along with 'Test Not Run'.". When I say original value here I mean the original cell value before it was populated with 'Test not run' which actually is the expected value. The reason I suggested this change is purely for readability. Nothing else. So if a test is ignored we can still see what the expected value was. If this small change is to confusing I've no problem cancelling this change.

PR #1472, opened 2024-01-12
Add Variable values to ScopeVariablesResponder

AlexNetcare:
This will add the values for the variables to the table. So instead of going to a variables location you can see the value directly in the table.

### DIFF
--- a/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
+++ b/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
@@ -1,5 +1,5 @@
 !2 Pending Changes
- * Added the values to the variables page. ([[1472][https://github.com/unclebob/fitnesse/pull/1472]])
+ * Added the defined values to the overview of all variables in scope. ([[1472][https://github.com/unclebob/fitnesse/pull/1472]])
 
 !2 20231203
 * Updated README and dependencies
--- a/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
+++ b/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
@@ -1,4 +1,5 @@
 !2 Pending Changes
+ * Added the values to the variables page. ([[1472][https://github.com/unclebob/fitnesse/pull/1472]])
 
 !2 20231203
 * Updated README and dependencies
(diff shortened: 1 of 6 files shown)


fhoeben:
Can you add a description of this feature to `FitNesseRoot/FitNesse/ReleaseNotes/content.txt`

Furthermore I just noticed that the responder is not documented at all in the user guide. That is a shame as it is a really useful tool. Could you maybe add a short description somewhere in the user guide pages (I think `FitNesseRoot/FitNesse/UserGuide/WritingAcceptanceTests/VariableScope/content.txt` might be a good location)? 
Doesn't have to be fancy, but it would be good if people could see they can get an overview (and how).

AlexNetcare:
There is a small information in the User Guide for the Responder under `FitNesseRoot/FitNesse/UserGuide/AdministeringFitNesse/RestfulServices/content.txt`. But updating the user guide for the VariableScope still seems good to me.

Additionally on the Variable Scope page, I added the information that only variables with !define are shown on the page. Variables like PAGE_PATH or PAGE_NAME are not shown. I am not sure if that was intended or not but I didn't want to change too much.

FitNesseRoot/FitNesse/UserGuide/WritingAcceptanceTests/VariableScope/content.txt
```diff
@@ -5,3 +5,5 @@ For example, lets say the variable we are interested in is X.
  * If X is defined on this page then it is definately in scope.
  * If X is not defined on this page but is defined in the page .FitNesse (this page's parent), then X is still in scope.
  * If X is defined on .FitNesse.UserGuide.FitNesseWiki then X is not in scope because .FitNesse.UserGuide.FitNesseWiki is not a parent of this page.
+
+All variables that have been declared using !define can be checked by adding ?variables at the end. For example .FitNesse.UserGuide.FitNesseWiki?variables will show a table with the names, their values and their location on the pages path.
```
fhoeben, line 9:
```suggestion
All variables that have been declared using !style_code(!define) can be checked by adding !style_code(?variables) at the end of the page's url.
For example !style_code(!-http://myFitNesseHost/FitNesse.UserGuide.FitNesseWiki?variables-!) will show a table with the names, their values and the pages the values are defined.
```

FitNesseRoot/FitNesse/ReleaseNotes/content.txt
```diff
@@ -1,4 +1,5 @@
 !2 Pending Changes
+ * Added the values to the variables page. ([[1472][https://github.com/unclebob/fitnesse/pull/1472]])
```
fhoeben, line 2:
```suggestion
 * Added the defined values to the overview of all variables in scope. ([[1472][https://github.com/unclebob/fitnesse/pull/1472]])
```

fhoeben (changes requested):
Minor rephrase of some of the documentation, but great otherwise

PR #1485, opened 2024-02-11
HTTP Basic Authentication: Don't fail on colons (":") in passwords

manubahr:
`fitnesse.http.Request#getCredentials()` fails to split the basic authentication header value correctly if the password contains one or more colons (":"). The splitting has to stop after the 1st colon separating user-id and password. Following colons are part of the password.

See:
*  https://www.rfc-editor.org/rfc/rfc7617.html page 4
> Furthermore, a user-id containing a colon character is invalid, as
> the first colon in a user-pass string separates user-id and password
> from one another; text after the first colon is part of the password.

* https://stackoverflow.com/a/28115727

### DIFF
--- a/src/fitnesse/http/Request.java
+++ b/src/fitnesse/http/Request.java
@@ -331,7 +331,8 @@ public class Request {
     if (hasHeader("Authorization")) {
       String authHeader = getHeader("Authorization");
       String userpass = getUserpass(authHeader);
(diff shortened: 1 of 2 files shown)


fhoeben:
@manubahr  do you mind resolving the merge conflicts so we can merge directly?

Thanks for your contributions!

Issue #1488, opened 2024-02-16
Switching off the page and test history


afmg-ekoeppen:
The creation of the test and page history is not relevant for my use case. It would be nice to disable this completely in the configuration so that no zip folders are created in the fitnesse folder structure. 

A possible solution strategy that I could well imagine would be. Define variables in the root page.

```
!define test.history.days {0}
!define page.history.days {0}
````

Or something similar.

**Comment by six42** [comment]
Are you aware of the existing nohistory parameter?
https://fitnesse.org/FitNesse/UserGuide/AdministeringFitNesse/RestfulServices.html

**Comment by afmg-ekoeppen** [comment]
> Are you aware of the existing nohistory parameter? https://fitnesse.org/FitNesse/UserGuide/AdministeringFitNesse/RestfulServices.html

Thank you. Yes, I am aware of that parameter, but it doesn't help me for my application.

**Comment by six42** [comment]
Hi Emil
Can you explain more why doesn't it helps?
What are your requirements?
How do you run your tests?
Do you want to have it switched off for the whole Wiki or just some pages, with or without child pages?

**Comment by afmg-ekoeppen** [comment]
> Hi Emil Can you explain more why doesn't it helps? What are your requirements? How do you run your tests? Do you want to have it switched off for the whole Wiki or just some pages, with or without child pages?

In my case, the history is mapped with a version management system. I work with a checkout on my local PC and host the FitNesse Wiki here. I run the tests on my PC in the browser. I edit the tests within an IDE and work in the FitNesse file system. For the wiki on my local system, I would like to be able to switch off the entire history for all tests in order to clean the file system of the generated zip folders. At the moment, I have to clean my checkout manually.

**Comment by six42** [comment]
Did you tried to set in plugins. properties:

test.history.days=0

This seems to work for me.

**Comment by fhoeben** [comment]
@afmg-ekoeppen I also never want to use these files (with my IDE or otherwise), but they have never really bothered me since I have the following in my project's `.gitignore`:
```
wiki/FitNesseRoot/**/*.zip
wiki/FitNesseRoot/files/testResults
wiki/FitNesseRoot/files/testProgress
wiki/FitNesseRoot/ErrorLogs
wiki/FitNesseRoot/FitNesse
wiki/FitNesseRoot/RecentChanges*
wiki/FitNesseRoot/ReRunLastFailures*.wiki
updateDoNotCopyOverList
updateList
properties
```

[...]

PR #1493, opened 2024-02-23
Feature/add limit to test histories

AlexNetcare:
This PR would add a couple things to the Test Histories.

1. An additional Property on a page to disable Test Histories for that page. This is turned off by default.
2. The behaviour when purging Test Histories would now change to only delete Test Histories of the current page. Additionally, an extra Button, next to the purge Buttons, to still purge globally can be checked.
3. Two new Properties for the plugins.properties:
     * TestHistory.maxCount=10 would define that only 10 Test Histories per page should be saved. This is off by default.
     * TestHistory.purgeTime=0,30,60,90 defines the purge Buttons on the Test History page. Currently will default to 0,30,60,90 when nothing else is configured.

Depending on the configured purge times, the GUI for the Test Histories would look like this:
image

By coincidence this might be taking care of this Request #1488 
The new Property does not work recursive though.


AlexNetcare:
These might be more changes than it should be for a single PR. I will make smaller PRs for each point individually.

PR #1507, opened 2024-04-21
Fix RerunLastFailures problem. Appear the perfix only one time

Eqerem-Hena:


src/fitnesse/responders/run/SuiteResponder.java
```diff
@@ -292,10 +292,32 @@ protected File getRerunPageFile() {
   }
 
   protected String getRerunPageName() {
-    PageCrawler pageCrawler = page.getPageCrawler();
-    WikiPagePath fullPath = pageCrawler.getFullPath();
-    String fullPathName = PathParser.render(fullPath);
-    return "RerunLastFailures_"+fullPathName.replace(".","-");
+	    PageCrawler pageCrawler = page.getPageCrawler();
```
fhoeben, line 295: @Eqerem-Hena you changed the indentation of these lines (and now it no longer matches that of other methods). Can you please fix this?

fhoeben (changes requested):
Please fix the indentation and add a unit test. Also a description of 'RerunLastFailures problem' (either as an issue or in this PR) would be much appreciated.

PR #1510, opened 2024-05-14
Feature/limit test histories

AlexNetcare:
Added a new configuration to limit the total amount of test histories per page.
The idea was to reduce the disk space needed in case the configuration for test.history.days is set to a high value like a year or more. At least in our case we had the problem that we had thousands of histories for a single page and they did not get deleted because the test.history.days was set so high. But reducing the days configuration was not an option.


fhoeben (approved):
Looks good. Can you also update the release notes page to include this?

Issue #1342, opened 2021-11-11
RerunLastFailures file prefix stacks when using the "Rerun Failed" button


NicoleRauch:
**Describe the bug**
When I have a page with failing tests, I get a file called `RerunLastFailures_pagename`. When I then click the "Rerun failed" button, I get a file called `RerunLastFailures_RerunLastFailures_pagename`. When I click the "Rerun failed" button again, I get a file with three prefixes and so on.

**To Reproduce**
1. Go to a page with a failing test
2. Click on 'Rerun Failed'
3. On the resulting page, click on 'Rerun Failed' (repeat this a few times)
3. Go to file system
4. See many files instead of one

**Expected behavior**
I would expect to only get one file `RerunLastFailures_pagename` no matter how often I press the button.

**Screenshots**

<img width="999" alt="image" src="https://user-images.githubusercontent.com/436499/141262800-4e6bd260-33b2-47c4-ad8b-96d3e90f5b0c.png">

**Configuration:**
 - OS: MacOS 11.6
 - Version of FitNesse: v20211030
 - Test System: Slim
 - Plugins in use: none

**Additional wishes and ideas**

Would it also be possible to move those files into a folder called `RerunLastFailures`? This would help keep my `FitNesseRoot` directory from exploding... Also, if you had a dedicated folder for these files, you could remove the prefixing (because the folder name already indicates what files they are) and that would eliminate the bug altogether.

fhoeben:
@raducu-deaconu-visma could you take a look at addressing this bug? I expect it was introduced with your PR to include the suite name into the rerun page name. 
I guess adding that name is not a good idea if the name already starts with 'RerunLastFailures' :-)

raducu-deaconu-visma:
Hi,
The reason why we have different pages created is that we are not actually testing the same thing.
If for example we have a page that fails on 5 tests initially, when we "Rerun failed" we are only targeting those 5 tests. If on this step one more test passes, we will have another rerun page, that only targets 4 tests.
This is the reason behind having more than one "RerunLastFailures" prefix, as we are actually getting a result from running the "RerunLastFailures" on the previous failed tests, not on the whole suite.

NicoleRauch:
Would it be possible to create those files in a subdirectory then? Currently my top-level directory gets flooded with those files, and there is no way to turn them off entirely that I know of...

fhoeben:
[...]

PR #1512, opened 2024-06-06
Fix RerunLastFailures problem. Appear the perfix only one time 

Eqerem-Hena:
Fix the #1342 issue.

### DIFF
--- a/src/fitnesse/responders/run/SuiteResponder.java
+++ b/src/fitnesse/responders/run/SuiteResponder.java
@@ -57,8 +57,6 @@ import java.util.Date;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import static fitnesse.responders.WikiImportingTraverser.ImportError;
 import static fitnesse.wiki.WikiImportProperty.isAutoUpdated;
@@ -299,27 +297,10 @@ public class SuiteResponder extends ChunkingResponder implements SecureResponder
     String fullPathName = PathParser.render(fullPath);
     if (fullPathName.startsWith("RerunLastFailures_")) {
       String newFullPathName = fullPathName.replace(".", "-");
-      if (!(newFullPathName.endsWith("(1)"))) {
-        return newFullPathName.replaceAll("\\(\\d+\\)", "(" + 2 + ")");
-      } else if (newFullPathName.endsWith("SubsetQueryTest")) {
-        return newFullPathName + "(1)";
-      } else {
-        int number = extractNumber(newFullPathName) + 1;
-        return newFullPathName.replaceAll("\\(\\d+\\)", "(" + number + ")");
-
-      }
+      return newFullPathName;
     } else {
       return "RerunLastFailures_" + fullPathName.replace(".", "-");
     }
-  }
-
-  public static int extractNumber(String input) {
-    Pattern pattern = Pattern.compile("\\((\\d+)\\)");
-    Matcher matcher = pattern.matcher(input);
-    if (matcher.find()) {
-      return Integer.parseInt(matcher.group(1));
-    }
-    return -1; // return -1 if no number found
   }
 
   protected String getTitle() {
--- a/src/fitnesse/responders/run/SuiteResponder.java
+++ b/src/fitnesse/responders/run/SuiteResponder.java
@@ -57,6 +57,8 @@ import java.util.Date;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


fhoeben:
@Eqerem-Hena Thanks a lot for your PR!

I believe you are generating unique names for each re-rerun by adding a counter at the end, correct?
To me this seems overkill as it keeps making a lot of pages if the rerun is retried often.

Like I stated in the issue just having one rerun per page/suite seems fine. So this PR could be much simpler: just don't add the `RerunLastFailures_` prefix if the  full page name already starts with that.
Do you think that is too limited?

fhoeben:
It also seems like some of the existing tests failed, can that be related to your change or did we let something else slip through and break that already?

Eqerem-Hena:
> @Eqerem-Hena Thanks a lot for your PR!
> 
> I believe you are generating unique names for each re-rerun by adding a counter at the end, correct? To me this seems overkill as it keeps making a lot of pages if the rerun is retried often.
> 
> Like I stated in the issue just having one rerun per page/suite seems fine. So this PR could be much simpler: just don't add the `RerunLastFailures_` prefix if the full page name already starts with that. Do you think that is too limited?

I think it is a better idea to add the counter at the end, as in this way it will be more understandable to the user which rerun page is.

Eqerem-Hena:
> It also seems like some of the existing tests failed, can that be related to your change or did we let something else slip through and break that already?

I don't think so. It's probably from a previous commit.

fhoeben:
The original submitter of the issue indicated:
> This would help keep my FitNesseRoot directory from exploding...

So they prefer not to have an additional file for each rerun.

Eqerem-Hena:
> The original submitter of the issue indicated:
> 
> > This would help keep my FitNesseRoot directory from exploding...
> 
> So they prefer not to have an additional file for each rerun.

I did a new commit. Please, check it out!

src/fitnesse/responders/run/SuiteResponder.java
```diff
@@ -295,7 +295,12 @@ protected String getRerunPageName() {
     PageCrawler pageCrawler = page.getPageCrawler();
     WikiPagePath fullPath = pageCrawler.getFullPath();
     String fullPathName = PathParser.render(fullPath);
-    return "RerunLastFailures_"+fullPathName.replace(".","-");
+    if (fullPathName.startsWith("RerunLastFailures_")) {
+      String newFullPathName = fullPathName.replace(".", "-");
```
fhoeben, line 299: I think it makes more sense to to the replace on line 297 to remove the duplicated code from both branches

test/fitnesse/responders/run/SuiteResponderTest.java
```diff
@@ -564,7 +570,29 @@ public void loadsCustomFormatters() throws Exception {
...
+  
+  @Test
+  public void testGetRerunPageName_withRerunPrefix() {
+
+    String fullPathName = "RerunLastFailures_SomePage";
+    when(PathParser.render(wikiPagePath)).thenReturn(fullPathName);
+
+    String result = responder.getRerunPageName();
+
+    assertEquals("RerunLastFailures_SomePage".replace(".", "-"), result);
```
fhoeben, line 582:
```suggestion
    assertEquals("RerunLastFailures_SomePage", result);
```

test/fitnesse/responders/run/SuiteResponderTest.java
```diff
@@ -564,7 +570,29 @@ public void loadsCustomFormatters() throws Exception {
...
 
+  @Test
+  public void testGetRerunPageName_withoutRerunPrefix() {
+
+    String fullPathName = "SomeOtherPage";
+    when(PathParser.render(wikiPagePath)).thenReturn(fullPathName);
+
+    String result = responder.getRerunPageName();
+
+    assertEquals("RerunLastFailures_SomeOtherPage".replace(".", "-"), result);
```
fhoeben, line 593:
```suggestion
    assertEquals("RerunLastFailures_SomeOtherPage", result);
```

test/fitnesse/responders/run/SuiteResponderTest.java
```diff
@@ -564,7 +570,29 @@ public void loadsCustomFormatters() throws Exception {
 
     assertTrue(FooFormatter.initialized);
   }
+
```
fhoeben, line 573: It is a good idea to add the replacement of `.` by `-` in that page names, but I would like to see that in separate test cases where the `fullPathName` actually contains a period and then also just do an `assertEquals` where the expect value is a literal with the periods replaced.

In the current setup the `replace` in the assertion doesn't do anything and the test doesn't make explicit what it is expecting

fhoeben (changes requested):
Looks really good. Thank you. Some minor comments and then we can merge

fhoeben:
Could you also update the release notes?

fhoeben:
Cleaned up the tests in #1519

Issue #1514, opened 2024-06-11
Heading anchor covered up by top nav bar in bootstrap theme


causton81:
**Describe the bug**
When I follow a link in a !headings widget the browser seems to scroll too far. I believe the top nav bar uses fixed positioning and is covering the heading.

If I add the below html to my PageHeader page, then I can get the scrolling to look better:
```
!-

<style>
  *[id] {
    scroll-margin-top: 96px;
  }
}
</style>

-!
```

**To Reproduce**
1. Edit a page with a !headings widget and insert newlines until a !1 is offscreen (or resize window so !1 is offscreen).
2. Follow the link in the !headings to the !1 and observe top of content.

**Expected behavior**
Following a link in a !headings should show that heading in the top of the browser when scrolling.

**Screenshots**
<img width="631" alt="image" src="https://github.com/unclebob/fitnesse/assets/4534764/b4b29ea1-4749-405e-b16f-1a4f5ef1dda5">

<img width="646" alt="image" src="https://github.com/unclebob/fitnesse/assets/4534764/3c2dccbd-7f67-4c95-a688-1f4f788f516f">

This is the workaround with style tag on root PageHeader:
<img width="849" alt="image" src="https://github.com/unclebob/fitnesse/assets/4534764/dfef88a7-8682-412d-a83e-83472f8878e6">

**Sample Wiki Structure/content**
If the problem has to do with how a file handled in the wiki or how it is parsed, please add a zip file containing files that demonstrate the problem.

**Stack Trace**
If applicable, add a stack trace of the error.

**Configuration:**
 - OS: [e.g. iOS] OSX 14.5
 - Browser : Chrome 126.0.6478.57
 - Version [e.g. 20200404] of FitNesse: 20240219
 - Test System: Slim
 - Plugins in use: none

**Additional context**

PR #1517, opened 2024-06-20
Fixed header overlapping content

AlexNetcare:
While trying to figure out Bootstrap, I found a Solution for the header overlapping the content when the header size changes. This might be a fix for #1514 unless I misinterpreted the Issue.
I was hoping to find a Bootstrap Solution but I either didn't find it or there is no other more elegant one.

### DIFF
--- a/src/fitnesse/resources/bootstrap/templates/skeleton.vm
+++ b/src/fitnesse/resources/bootstrap/templates/skeleton.vm
@@ -63,8 +63,10 @@
 
   <script>
    function adjustPadding() {
-    var navHeight = document.querySelector('nav').offsetHeight;
-    document.body.style.paddingTop = navHeight + 'px';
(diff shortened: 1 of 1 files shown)


src/fitnesse/resources/bootstrap/templates/skeleton.vm
```diff
@@ -61,5 +61,18 @@
   </footer>
   #end
 
+  <script>
+   function adjustPadding() {
+    var navHeight = document.querySelector('nav').offsetHeight;
```
fhoeben, line 66: Just to be sure I suggest you check whether `document.querySelector('nav')` gives a result before trying to use it.

AlexNetcare: Added a simple check. If a nav Element does not exist then there won't be an exception in the Console now.

PR #1525, opened 2024-10-21
Also build and test on Java 17

fhoeben:
Update GitHub workflow to also build FitNesse on Java 17 and run all tests

@six42 as you can see I had to update one test case to get the build to pass. It appears that `fitnesse.slim.LoggingOutputStream` does not behave the same on Java 17 as it did on earlier versions. An extra newline is added to the execution log for each `println()` call to either stdin or stout when the communication with the Slim server is not via a socket. 

I have had issues in the past with this way of communicating between FitNesse and the Slim server (character encoding trouble on an IBM JVM on windows), so in my tests I always just have Slim choose a random free port (by setting the Slim port to `0`). Changing the communication protocol to use an actual socket also fixed this test. I couldn't put my finger on why the extra newline is added, but maybe it's an obvious fix for you?
I find the name of the class (`LoggingOutputStream`) a bit strange also, it's not logging its decorating the input it receives before forwarding it to the nested stream as I understand it. Maybe it would already be much simpler (the newline handling in there) if instead of just forwarding the lines received as text the records would be Base64 encoded (then each record will just always be 1 line I think)?

### DIFF
--- a/.github/workflows/build-fitnesse.yml
+++ b/.github/workflows/build-fitnesse.yml
@@ -6,6 +6,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        java-version: [ '11', '17' ]
+
     steps:
       - name: Check out Git repository
         uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


fhoeben:
@six42 I don't quite understand why I get the extra newline, but I added some code to deal with the difference. Could probably be done nicer though. So if you have some time....

PR #1538, opened 2024-10-26
Fixed Ignore exception issue whereby it was still executing every sta…

eireglas:
Fixed Ignore exception issue whereby it was still executing every statement in the script table even though it will ignore the result

Example below shows IGNORE_ALL_TESTS exception but what I found was it would still execute the lines below even though it would ignore the result. This fix resolves that issue which also help speed up the overall test
image

### DIFF
--- a/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
+++ b/FitNesseRoot/FitNesse/ReleaseNotes/content.txt
@@ -1,5 +1,6 @@
 !2 Pending Changes
  * Fix SLF4J logging ([[1522][https://github.com/unclebob/fitnesse/pull/1522]])
+ * Fix Ignore exception issue whereby it still executes every statement in the script table even though it will ignore the result
 
 !2 20240707
  * Allow usage of JDK > 18 ([[1513][https://github.com/unclebob/fitnesse/pull/1513]]).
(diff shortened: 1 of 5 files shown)


eireglas:
@fhoeben do you mind reviewing?

fhoeben:
@eireglas what is the advantage of this way of ignoring all compared to just stopping the test?
I believe your change also changed the behaviour of the exception just to ignore assertions in a single table. Is that indeed the intention? That would be functionality to that is not currently available I believe.

I think we also need some tests to document/check the right things are ignored and not executed (not too little or too much)

eireglas:
@fhoeben Thanks for the feedback. I've added some tests which would previously have failed but now pass due to my change. Also, I have updated the release notes to better explain the change.
To answer your question with regards the advantage of ignoring all compared to just stopping....by ignoring all, the end result will  mean a green, success, rather than show yellow or red for failure. I use fitnesse in different environments and by using this IgnoreAllTests or IgnoreScriptTest functionality I can have a test run certain tables and ignore others depending on the environment, which I find very useful.

fhoeben:
@eireglas  looks good, thanks. One last comment: could you update the test so the difference in behaviour between ignore script and all is explicit. I think they currently both show that the script is stopped/ignored, so we don't see that for ignore all subsequent ones are also ignored, while when just the one script is ignored subsequent ones are executed normally.

eireglas:
@fhoeben Thanks, the 2 new tests I added were just to make sure that when either of the ignore exceptions are thrown that the statement results in a null i.e. not executed.
With my last commit I've just added an extra line to 2 existing tests in SlimTestSystemTableProcessingTest to show where it explicitly tests the differences between how the ignore exceptions should be handled.

eireglas:
@fhoeben Do you know when the next release will happen? I am eager to get this fix implemented into our test suites. Thanks!

fhoeben:
Should be in 20241207

PR #1568, opened 2025-06-06
Enhance !contents to support filtering by Suites tag using -tag / -t option

JGProg:
### Summary

This pull request enhances the existing `!contents` directive by adding a new optional argument: `-t`, allowing users to **filter listed pages based on the `Suites` property**.

---

### Motivation
In many projects, the same suite structure is reused across multiple contexts or products (e.g. Product A, Product B). However, some pages are only relevant to a specific product. Until now, there was no way to filter the `!contents` listing dynamically by tag.

This change adds that capability.

---

### New behavior:  `-t` option
You can now use `!contents` like this:

```wiki
!contents -R2 -tProductA

### DIFF
--- a/src/fitnesse/wikitext/parser/Contents.java
+++ b/src/fitnesse/wikitext/parser/Contents.java
@@ -16,43 +16,52 @@ public class Contents extends SymbolType implements Rule, Translation {
 
     @Override
     public Maybe<Symbol> parse(Symbol current, Parser parser) {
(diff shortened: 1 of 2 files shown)


fhoeben (changes requested):
Looks good. Can you also add some tests and some documentation (as a wiki page)?

JGProg:
it's ok @fhoeben  :) 
I creat two unit test and add documentation :)

PR #1569, opened 2025-06-17
VSCode and Dev Container support added

DominikNetcare:
It's time to take the next step.
This pull request adds support for Docker Dev Containers and Visual Studio Code (VSCode).
Everything needed to continue developing FitNesse and run all Gradle tasks is included.
The advantage is that nothing more than a Docker and VSCode installation is required.

In addition to two debug run configurations for VSCode, template watcher start and stop tasks have been added to eliminate the need for manual reloading after changes to .vm files.

Screenshot 2025-06-17 164337
Screenshot 2025-06-17 164204

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,17 +89,17 @@ There are a few things to keep in mind when working from an IDE:
 1. Clone the FitNesse Git repository from https://github.com/unclebob/fitnesse.
 2. Import FitNesse via _File_ -> _Import..._ -> _Gradle Project_.
 3. Select the just cloned project folder. Follow the wizard.
-4. Ensure the project properties have a Java 8 compiler or newer set.
+4. Ensure the project properties have a Java 11 compiler or newer set.
 
-#### Import FitNesse in IntelliJ IDEA (16)
+#### Import FitNesse in IntelliJ IDEA (2025.1)
 
-1. Clone the FitNesse Git repository from https://github.com/unclebob/fitnesse.
-2. From the welcome screen (the one you get when all projects are closed), click _Import Project_.
-3. Select the file `build.gradle` in the fitnesse folder.
-4. Follow the wizard. Deselect the option `Create separate module per source set`. You can use the
-   `Use gradle wrapper task configuration`. Use Java 8 or newer. It should find source and test folders and
-    show you two modules: `fitnesse` and `:buildSrc`; import both.
-5. Open the Gradle Build tool, select the task `copyRuntimeLibs` and (right-click) mark it as _Execute After Make_.
+1. Create a new project from https://github.com/unclebob/fitnesse (i.e. from the menu select _File_ -> _New_ -> _Project from Version Control..._,
+   ensure the 'Version control' selectbox contains 'Git', enter the project's URL and press 'Clone').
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,37 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "java",
+            "name": "Debug FitNesse + Template Watcher",
```
fhoeben, line 6: Shouldn't this just be called 'FitNesse + Template Watcher'? I'm not familiar exactly how this is used in VSCode as I don't use it myself. But this launch configuration just seems to describe how to start FitNesse, whether that is a regular run or a debug session seems to depend on how this config is used.
That is at least how it works in IntelliJ: one defines launch configurations and then you decide whether to 'Run', 'Debug' or 'Profile' those...

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,27 @@
...
+        "vscjava.vscode-gradle",
+        "SonarSource.sonarlint-vscode"
+      ]
+    }
+  },
+  "forwardPorts": [
+    8081
+  ],
+  "remoteUser": "vscode"
+}
```
fhoeben, line 27: Normally we end all files with a newline. Is that also supported for these files? Or are they auto generated by a process that doesn't do newlines at the end and that would remove newlines if they were present?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "FitNesse",
+  "image": "mcr.microsoft.com/vscode/devcontainers/base:focal",
+  "features": {
+    "ghcr.io/devcontainers/features/java:1": {
+      "version": "11",
+      "installMaven": "false",
+      "installGradle": "false"
```
fhoeben, line 8: Given that this container just seems to get the source of the project I would expect it to get Gradle also, so it can compile the source.

Is that not needed for VSCode?

fhoeben (approved):
I support the approach of defining a standard dev container to build and run the project. But I'm not so sure about strongly coupling it to VSCode.

As a generic dev container I would expect it to have Gradle installed as well, so the container has all tools needed to compile the source. (Even though this may not be needed for VSCode?)

As you can tell by the README maintainers in the past did not have the actual definition files in here, but just a description. Having the full definition is of course much more convenient. 

I don't like the fact that we'll have content in here that is not actively maintained, but that also goes for the descriptions in the README (I just noticed the Eclipse one still mentions Java 8!).

So I guess the vscode files don't harm anyone and can go in.

.vscode/launch.json
DominikNetcare: Renamed the launch configs to remove "Debug". In VSCode it is also possible to use them as Run configs without having a debug session, even though for running fitnesse, the "./gradlew run" command should be preferred as stated in the README file.

.devcontainer/devcontainer.json
DominikNetcare: Must have missed that one. Added last newlines now.

